PR #19, opened 2019-12-06
Dispute list

fabriziovigevani:
First iteration of the Dispute lists view with mock data.

Includes: 

- Disputes lists view
- Dispute detail view
- Dispute timeline


src/components/Disputes/DisputeCard.js
```diff
@@ -0,0 +1,112 @@
...
+          }
+        `}
+      >
+        <h3
+          css={`
+            ${textStyle('body1')}
+          `}
+        >
+          Dispute
+          <strong> #{dispute.id}</strong>
```
rperez89: should we define those attributes as const {id, description, rewardAmount, stakedAmount} = dispute?

PR #22, opened 2019-12-10
Dashboard tasks

rperez89:
This PR includes the task list for the dashboard and changing the date-fns for dayjs


src/components/Lib/web3-utils.js
```diff
@@ -0,0 +1,13 @@
+import { toChecksumAddress } from 'web3-utils'
```
bpierre: @rperez89 I recently used [this approach](https://github.com/aragon/aragon-ui/pull/659) to avoid importing web3-utils in aragonUI: do you think it would work here? Or are we going to need more things from web3-utils?

src/components/Tasks/TasksTable.js
```diff
@@ -56,7 +50,7 @@ const TaskTable = ({ tasks }) => {
     () =>
       filteredTasks.sort(({ dueDate: dateLeft }, { dueDate: dateRight }) =>
         // Sort by date ascending
-        compareAsc(dateLeft, dateRight)
+        dayjs(dateLeft).isAfter(dayjs(dateRight)) ? 1 : -1
```
bpierre: Should we also return `0` if the two dates are identical? Or is it too highly improbable?

src/components/Dashboard/TaskTable.js
```diff
@@ -0,0 +1,127 @@
...
+            )}
+          </div>
+        </>
+      }
+      fields={[
+        { label: 'Task', priority: 1 },
+        { label: 'Dispute', priority: 2 },
+        { label: 'Priority', priority: 3 },
+        { label: 'Assigned to juror', priority: 4 },
+        { label: 'Due date', priority: 5 },
```
bpierre: We can skip the `priority` field if we want to keep the natural order, and we can even drop the object syntax entirely if we only deal with labels! :smile: 

```jsx
<DataView
  fields={[
    'Task',
    'Dispute',
    'Priority',
    'Assigned to juror',
    'Due date',
  ]}
/>
```

src/components/Dashboard/TaskTable.js
```diff
@@ -0,0 +1,127 @@
+import React, { useMemo, useState } from 'react'
+import {
+  ContextMenu,
+  DataView,
+  DateRangePicker,
+  GU,
+  Link,
+  Text,
```
bpierre: We should avoid using `Text` now that it is deprecated in aragonUI (replaced by `textStyle()`).

src/components/Lib/web3-utils.js
rperez89: mmm not sure if we are going to be using more utils, what do you think?

bpierre: Not sure, let’s see if we start using other things, and we’ll move back to `web3-utils` at this point!

PR #23, opened 2019-12-12
Layout + Court Theme

bpierre:
~Requires https://github.com/aragon/aragon-ui/pull/685~ updated aragonUI to 1.2.0 to support the new theme API.

- Override the `Layout` component to make it work with the main view.
- Add a custom theme and replace static colors by dynamic ones.
- Also removes `MainButton` now that `Button` knows about the current theme.


src/components/Dashboard/LatestActivity.js
```diff
@@ -12,30 +12,31 @@ function LatestActivity() {
...
+                  padding: ${1 * GU}px;
                 `}
               >
                 <span
                   css={`
-                    background: rgb(255, 151, 128);
-                    border-radius: 50%;
                     display: block;
-                    padding: 5px;
+                    padding: ${0.5 * GU}px;
```
rperez89: Just a question, how do you guys do when for example on a figma file says 5px of padding ? we just say ok let's use the nearest value using the grid unit? in this case 0.5 = 4px?

src/components/Tasks/TasksBox.js
```diff
@@ -2,7 +2,7 @@ import React from 'react'
 import { Box, GU, useViewport } from '@aragon/ui'
 import TaskAmounts from './TasksAmounts'
 
-const TasksBox = () => {
+function TasksBox() {
```
rperez89: ohh sorry about that, you try to keep using the function prefix as i can see

src/components/Dashboard/LatestActivity.js
bpierre: It depends haha :grimacing: 

We generally try to understand if it could work using the nearest 2 / 4 / 8 or if the value is really intended. For normal spacing like in this case, it’s generally safe to assume that it’s a mistake in the Figma file and we can correct it ourselves.

src/components/Tasks/TasksBox.js
bpierre: Ah no worries it’s just a habit from the other Aragon codebases (where we tend to use normal functions for top level components), but I don’t mind either way. Let me know what you prefer and I’ll follow :-)

rperez89: I think that makes more sense to see there the "function" for the top level component, i am good with that

bpierre: Cool, let’s do that then :smile:

bpierre:
@rperez89 All ready now!

rperez89:
> @rperez89 All ready now!

Woohooo , should i merge it then?

bpierre:
@rperez89 If it’s all good for you, yes! :smile:

PR #31, opened 2020-01-08
Web3 provider

fabriziovigevani:
This includes a basic web3 Provider. Still have to finish the contracts interaction that will be done in another PR. 

It also includes the CourtConfig provider and the graph endpoints based on the current chain.

Also fixed the header and sidebar.


src/queries/courtConfig.js
```diff
@@ -0,0 +1,40 @@
+import gql from 'graphql-tag'
```
rperez89: I would rename this file to just court, as we are doing with the others queries and inside have the courtConfig query

src/providers/CourtConfig.js
```diff
@@ -0,0 +1,41 @@
...
+
+const CHAIN_ID = environment('CHAIN_ID')
+const CourtConfigContext = React.createContext()
+
+function CourtConfigProvider({ children }) {
+  const courtAddress = networks[getNetworkName(CHAIN_ID)].court
+
+  console.log(courtAddress)
+
+  const [result] = useSubscription({
```
rperez89: i am using one hook per query so in the component we just use `data = useCourtSubscription`

src/queries/courtConfig.js
```diff
@@ -0,0 +1,40 @@
...
+
+export const CourtConfig = gql`
```
rperez89: one thing that facu mentioned in one of my pr's is to just query what we need not everything

src/providers/CourtConfig.js
```diff
@@ -0,0 +1,41 @@
...
+    query: CourtConfigSubscription,
+    variables: { id: courtAddress.toLocaleLowerCase() },
+  })
+
+  // TODO: handle possible errors
+  const courtConfig = result.data && result.data.courtConfig
+
+  return (
+    <CourtConfigContext.Provider value={courtConfig}>
```
rperez89: should we pass all the config or one prop by prop?

src/queries/courtConfig.js
fabriziovigevani: I think that eventually we will use all this data but will remove the ones we don't currently use and if needed let them be added later 👍

src/providers/CourtConfig.js
fabriziovigevani: Unless we need to do some conversion i think that is ok to pass directly the configuration. Is there a reason that you think would be good to pass it like you mention?

package.json
```diff
@@ -4,19 +4,32 @@
...
+    "subscriptions-transport-ws": "^0.9.16",
     "urql": "^1.6.3"
   },
   "scripts": {
-    "start": "react-app-rewired start",
-    "build": "react-app-rewired build",
+    "start": "npm run start:rpc",
+    "start:rpc": "REACT_APP_CHAIN_ID=1337 react-app-rewired start",
+    "start:mainnet": "REACT_APP_CHAIN_ID=1 react-app-rewired start",
+    "start:rinkeby": "REACT_APP_CHAIN_ID=3 react-app-rewired start",
```
bpierre:
```suggestion
    "start:rinkeby": "REACT_APP_CHAIN_ID=4 react-app-rewired start",
```

package.json
```diff
@@ -4,19 +4,32 @@
...
+    "start:ropsten": "REACT_APP_CHAIN_ID=4 react-app-rewired start",
```
bpierre:
```suggestion
    "start:ropsten": "REACT_APP_CHAIN_ID=3 react-app-rewired start",
```

src/Header.js
```diff
@@ -1,17 +1,58 @@
...
+        <div
+          css={`
+            margin-right: 12px;
+          `}
+        >
+          {account ? (
+            <span
+              css={`
+                background-color: #54c497;
+                color: white;
```
bpierre: We should probably use the theme for these two.

src/web3-utils.js
```diff
@@ -0,0 +1,15 @@
...
+  if (chainId === '1') return 'mainnet'
+  if (chainId === '3') return 'ropsten'
+  if (chainId === '4') return 'rinkeby'
+
+  return DEFAULT_LOCAL_CHAIN
+}
+
+export function isLocalNetwork(chainId) {
+  return getNetworkName(chainId) === DEFAULT_LOCAL_CHAIN
+}
```
bpierre: What would you think about renaming this function `isLocalOrUnknownNetwork()`?

PR #46, opened 2020-01-20
Timeline icons: optimize using svgo

bpierre:


bpierre:
@rperez89 I have `svgo` installed globally (`npm i -g svgo`), and:

```console
svgo src/assets/Icon*.svg
```

PR #48, opened 2020-01-20
Balance module next

fabriziovigevani:
Includes:

- All the different ui reactions to the different juror states.
- Clock provider which tells you the current term, current term start time, current term end time. Clock component so it's easier for validation.
- subscription for non settled appeals

For the rewards module and appeal colateral box will be working from a fork of this branch.


src/hooks/useCourt.js
```diff
@@ -160,14 +178,22 @@ export function useDisputeActions() {
...
+      jurorRegistryContract.totalActiveBalanceAt(termId).then(balance => {
+        if (!cancelled) setTotalActiveBalance(balance)
+      })
+    },
+    [jurorRegistryContract, termId]
+  )
 
-  // TODO: handle possible errors
-  const courtConfig = result.data && result.data.courtConfig
+  useInterval(fetchTotalActiveBalance, POLL_EVERY, true)
```
bpierre: I am wondering if we shouldn’t use a `setTimeout()` instead? With `setInterval()`, every request taking more than one second will get cancelled, but we might want to wait for a request to complete before starting the next one. What do you think?

src/providers/Clock.js
```diff
@@ -0,0 +1,43 @@
...
+
+ClockProvider.propTypes = {
+  children: PropTypes.node,
+}
+
+function useClock() {
+  return useContext(ClockContext)
+}
+
+export { ClockProvider, useClock }
```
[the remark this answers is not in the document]
fabriziovigevani: Yes, totally makes sense 👍

src/hooks/useCourt.js
fabriziovigevani: Yes very good point

fabriziovigevani:
Let me know if you thing the poll with `setTimeout` is okay and i'll merge it :)

PR #51, opened 2020-01-22
Add a script to optimize assets (SVG only for now)

bpierre:
So that we can optimize the SVG files easily:

```console
yarn optimize
```


bpierre:
> Should we use yarn from now on instead of npm?

We recently decided to use `yarn.lock` files, but other than that, keep using `npm run` in scripts.

Then we can run one or the other locally, depending on the personal preferences, but I would recommend yarn to take the `yarn.lock` into account (and also because it’s much faster :smile: ).

PR #56, opened 2020-01-28
Dispute interactions

fabriziovigevani:
Includes all dispute contract interactions and the different states of the dispute detail.

- As discussed a few days ago, moved every subscription hook to a single `subscription-hooks` file (same for queries).


src/components/Disputes/DisputeTimeline.js
```diff
@@ -164,33 +182,116 @@ function ItemStep({ item, index, roundStepContainer }) {
...
+          >
+            {outcomeText}
+          </span>
+        </div>
+      )}
+    </div>
+  )
+}
+
+export function getOutcomeIcon(outcome, theme) {
```
bpierre: Should we make this a hook rather than passing `theme`, and call it something like `useOutcomeStyle()` since it also returns a color?

src/components/Disputes/DisputeTimeline.js
```diff
@@ -164,33 +182,116 @@ function ItemStep({ item, index, roundStepContainer }) {
...
+  if (phase === DisputePhase.RevealVote) {
+    outcomeText = juryOutcomeToString(outcome)
+  } else {
+    const confirm = phase === DisputePhase.ConfirmAppeal
+    outcomeText = appealRulingToString(outcome, confirm)
+  }
+
+  return (
+    <div>
+      {Icon && (
```
bpierre: I don’t think `Icon` can be falsy?

src/components/Disputes/DisputeTimeline.js
```diff
@@ -164,33 +182,116 @@ function ItemStep({ item, index, roundStepContainer }) {
   )
 }
 
+function Outcome({ outcome, phase }) {
+  const theme = useTheme()
+  const title =
+    phase && phase === DisputePhase.RevealVote ? 'JURY OUTCOME' : 'OUTCOME'
```
bpierre: We could use `text-transform: uppercase` instead of uppercase text:

```suggestion
    phase && phase === DisputePhase.RevealVote ? 'Jury outcome' : 'Outcome'
```

src/components/Disputes/DisputesStateProvider.js
```diff
@@ -0,0 +1,20 @@
+import React, { useContext } from 'react'
+import { useDisputesSubscription } from '../../hooks/subscription-hooks'
+
+const DisputesContext = React.createContext()
+
+function DisputesStateProvider({ children }) {
+  // const connectedAccount = useConnectedAccount()
+
+  const Provider = DisputesContext.Provider
```
bpierre: I think we can remove these two :slightly_smiling_face: (and use `<DisputesContext.Provider>` directly).

Also do we really need a provider here? Since we are only consuming `useDisputesSubscription()`.

src/components/Disputes/panels/RevealPanel.js
```diff
@@ -0,0 +1,66 @@
...
+    }
+  }
+
+  return (
+    <form onSubmit={handleReveal}>
+      <Field label="One time code">
+        <TextInput
+          value={password}
+          css={`
+            height: 88px;
```
bpierre:
```suggestion
            height: ${11 * GU}px;
```

src/hooks/useCourtContracts.js
```diff
@@ -174,11 +197,110 @@ export function useDisputeActions() {
...
+export function useAppealDeposits(disputeId, roundId) {
+  const [appealDeposits, setAppealDeposits] = useState([bigNum(0), bigNum(0)])
+
+  const disputeManagerContract = useCourtContract(
+    CourtModuleType.DisputeManager,
+    disputeManagerAbi
+  )
+
+  useEffect(() => {
+    const getNextRoundDetails = async () => {
```
bpierre: Should we call this `nextRoundDetails()` since it doesn’t return anything?

src/hooks/useCourtContracts.js
```diff
@@ -174,11 +197,110 @@ export function useDisputeActions() {
...
+      if (!disputeManagerContract) return
+      const nextRound = await disputeManagerContract.getNextRoundDetails(
+        disputeId,
+        roundId
```
bpierre: Probably something that would impact many places, but I’m wondering if we should have some sort of a retry mechanism for these calls, in case they fail for any reason.

src/hooks/useKeyCodeActions.js
```diff
@@ -0,0 +1,22 @@
+import { useCallback, useMemo, useEffect, useState } from 'react'
+import { saveAs } from 'file-saver'
+import { generateRandomNumber } from '../lib/math-utils'
+import { dateFormat } from '../utils/date-utils'
+
+export default function useKeyCodeActions() {
```
bpierre: As I was suggesting in another comment, I think we should rename this to be `useOneTimeCodeActions()` (or the name we decide).

src/hooks/useSidePanel.js
```diff
@@ -9,12 +9,22 @@ export function useSidePanel() {
     setOpened(false)
   }, [])
 
-  const endTransition = setOpened
+  const endTransition = useCallback(() => {
+    return setOpened
+  }, [setOpened])
```
bpierre: I don’t think this works as expected 😬

src/utils/juror-draft-utils.js
```diff
@@ -0,0 +1,24 @@
...
+
+export function getJurorDraft(round, jurorId) {
+  if (!round) return null
+
+  return round.jurors.find(jurorDraft =>
+    addressesEqual(jurorDraft.juror.id, jurorId)
+  )
+}
+
+export function jurorVoted(jurorDraft) {
```
bpierre: What would you think of using a verb here?

```suggestion
export function hasJurorVoted(jurorDraft) {
```

src/utils/dispute-utils.js
```diff
@@ -222,74 +290,118 @@ function getRoundPhasesAndTime(courtConfig, round, currentPhase) {
...
-      active: DisputesTypes.Phase.AppealRuling === currentPhase.phase,
-      roundId: roundId,
+      endTime: appealEndTime,
+      active:
+        isCurrentRound &&
+        DisputesTypes.Phase.AppealRuling === currentPhase.phase,
+      roundId,
+      outcome: roundAppealed ? appeal.appealedRuling : null,
+      showOutcome:
+        dayjs(new Date()).isAfter(appealEndTime) ||
```
bpierre: Should we store `dayjs(new Date())` above?

src/components/Disputes/DisputeActions.js
```diff
@@ -1,103 +1,337 @@
...
+        jurorDraft={jurorDraft}
+        hasJurorVoted={hasJurorVoted}
+        lastRound={lastRound}
+      />
+      {(() => {
+        if (phase === DisputePhase.VotingPeriod) return null
+
+        if (isJurorDrafted && hasJurorVoted) {
+          if (phase === DisputePhase.RevealVote) {
+            return (
```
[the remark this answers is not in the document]
fabriziovigevani: Yes, much cleaner indeed 👍

src/components/Disputes/DisputeActions.js
```diff
@@ -1,103 +1,337 @@
...
+      </span>{' '}
+      will be discounted
+    </span>
+  )
+}
+
+const ANJRewardsMessage = ({ isOpen }) => {
+  const theme = useTheme()
+
+  const Rewards = (
```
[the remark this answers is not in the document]
fabriziovigevani: Yes, totally true

src/components/Disputes/actions/DisputeVoting.js
```diff
@@ -0,0 +1,63 @@
...
+          disabled={!isJurorDrafted}
+          onClick={() => onRequestCommit(VOTE_OPTION_REFUSE)}
+        >
+          Refuse to vote
+        </VotingButton>
+      </div>
+      <Info mode={isJurorDrafted ? 'description' : 'warning'}>
+        {isJurorDrafted
+          ? ' You will be asked a password before you can commit your vote.'
+          : 'You cannot vote on this dispute with the current enabled address.'}
```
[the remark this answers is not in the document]
fabriziovigevani: Ah yess!! Good catch

src/components/Disputes/panels/CommitPanel.js
```diff
@@ -0,0 +1,272 @@
...
+  }
+
+  const handleDownloadKeyCode = useCallback(() => {
+    downloadKeyCode()
+    setCodeSaved(true)
+  }, [downloadKeyCode])
+
+  const handleCopyKeyCode = useCallback(() => {
+    setCodeCopied(true)
+    toast('One Time Password copied to clipboard')
```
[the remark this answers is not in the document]
rperez89: ohh haha that keyCode in the code was introduced by me, what name do you think that would be better?

src/components/Disputes/panels/AppealPanel.js
```diff
@@ -0,0 +1,206 @@
...
+
+    return false
+  }
+
+  // For submission
+  const handleAppeal = async event => {
+    try {
+      event.preventDefault()
+
+      const error = validateForm(selectedOutcome.value)
```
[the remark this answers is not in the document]
fabriziovigevani: Definitely

src/components/Disputes/DisputeDetail.js
```diff
@@ -81,8 +107,89 @@ const DisputeDetail = React.memo(function DisputeDetail({ match }) {
...
+    title = `Reveal your vote on dispute #${disputeId}`
+  else if (mode === REQUEST_MODE.APPEAL) {
+    if (data.confirm) {
+      title = `Confirm an appeal on dispute #${disputeId}`
+    } else {
+      title = `Appeal ruling on dispute #${disputeId}`
+    }
+  }
+
+  return <span>{title}</span>
```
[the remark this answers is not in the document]
fabriziovigevani: Nice suggestion

src/components/Disputes/DisputeTimeline.js
fabriziovigevani: You are right

src/components/Disputes/DisputeTimeline.js
fabriziovigevani: Yes, good idea

src/components/Disputes/DisputesStateProvider.js
fabriziovigevani: Initially thought that it would be useful but since we only subscribe to disputes it's not really needed in this case (Actually we weren't using it so i'm removing it)

src/hooks/useCourtContracts.js
fabriziovigevani: What about `fetchNextRoundDetails` for consistency with other functions that have the same functionality e.g `fetchTotalActiveBalance`

src/utils/dispute-utils.js
fabriziovigevani: Yup, good idea

src/hooks/useCourtContracts.js
fabriziovigevani: Yes, totally makes sense, created a `retryMax` function with a default number of attempts of 5. Let me know if you see any issues with the implementation

src/components/Disputes/panels/CommitPanel.js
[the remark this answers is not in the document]
dizzypaty: Good catch @bpierre ! I think we should go with One-time-use code, at least for the UI copy.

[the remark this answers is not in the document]
bpierre: Sounds good! For the code, we might want to go for a shorter alternative like `oneTimeCode`, but I would avoid `keyCode` as it’s exactly what is used for DOM events :D

src/hooks/useCourtContracts.js
bpierre: LGTM!

PR #60, opened 2020-01-29
First demo

fabriziovigevani:
**DO NOT MERGE**

Making this PR to have an early preview of the dashboard


src/queries/disputes.js
```diff
@@ -71,14 +62,11 @@ export const SingleDispute = gql`
       state
       metadata
       createdAt
-      subject {
+      evidences {
```
fabriziovigevani: From what i'm seeing in the subgraph schema, this data is not directly obtained from the dispute but rather from the subject (how it was before). 

Testing it the disputes doesn't load so i think this query could be wrong

PR #65, opened 2020-01-30
Balance module validation forms

fabriziovigevani:
Includes:

- All error validations for the ANJ balance interactions
- Balances formatted to 2 decimals
- Input to set the max amount.
- Fixed error that set the max decimals for the amount to 3


src/components/Dashboard/panels/ANJForm.js
```diff
@@ -1,38 +1,49 @@
...
-        digits: Math.min(anjToken.decimals, MAX_INPUT_DECIMAL_BASE),
-      }),
-    [anjToken.decimals]
-  )
-
   // Change amount handler
   const handleAmountChange = useCallback(event => {
     const newAmount = event.target.value
     setAmount(amount => ({ ...amount, value: newAmount }))
   }, [])
```
bpierre: I think we should store two values here: `event.target.value` to update the field in the next render, and a filtered version of it as a BN.js number.

If the value can’t be converted into a BN.js number, we could have a value representing its impossible state. For example, if the user types “abc”, we could use `bigNum(-1)` or `null`.

Another option would be to do this on submit, but we might want to know the parsed value before that (e.g. to inform that the value is higher than the max).

src/components/Dashboard/panels/ANJForm.js
```diff
@@ -56,21 +69,53 @@ const ANJForm = React.memo(function ANJForm({
...
           name="amount"
           wide
           onChange={handleAmountChange}
           value={amount.value}
-          step={minStep}
-          min="1"
           ref={inputRef}
           required
+          adornment={
+            <span
```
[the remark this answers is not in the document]
fabriziovigevani: Ah yes good call

src/lib/math-utils.js
```diff
@@ -38,7 +38,7 @@ export function parseUnits(value, digits) {
  */
 export function formatUnits(
   value,
-  { digits = 18, commas = true, replaceZeroBy = '0' } = {}
+  { digits = 18, commas = true, replaceZeroBy = '0', precision = 2 } = {}
```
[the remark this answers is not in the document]
fabriziovigevani: Totally missed it 😅

src/components/Dashboard/panels/ANJForm.js
fabriziovigevani: Ah yes, good idea. I also ended up using the common form errors on the `ANJForm` and removing them from the parent panel.

src/components/Dashboard/panels/ANJForm.js
```diff
@@ -41,14 +59,28 @@ const ANJForm = React.memo(function ANJForm({
         commas: false,
         precision: anjToken.decimals,
       }),
+      valueBN: maxAmount,
     }))
   }, [anjToken.decimals, maxAmount])
 
+  // Form valdiation
```
bpierre:
```suggestion
  // Form validation
```

PR #70, opened 2020-02-01
Disputes loading states

fabriziovigevani:
Includes all loading states for the dispute and the different messages:

- Dipsutes Loading message
- No disputes message
- Error loading disputes message (with additional error info, mostly from graphql)


src/components/Disputes/DisputeList.js
```diff
@@ -1,9 +1,17 @@
...
+import { Bar, CardLayout, GU, useTheme } from '@aragon/ui'
+
 import * as DisputesTypes from '../../types/dispute-status-types'
 import DisputeCard from './DisputeCard'
 import DisputeFilters from './DisputeFilters'
+import MessageCard from '../MessageCard'
+import DisputesLoading from './Loading'
+
+import noDataSvg from '../../assets/noData.svg'
+import noDraftSvg from '../../assets/noDraft.svg'
```
bpierre: We should move these after the JS imports

src/components/ErrorLoading.js
```diff
@@ -0,0 +1,41 @@
...
+    <div>
+      <span>
+        Something went wrong! You can restart the app, or you can{' '}
+        <span
+          css={`
+            color: ${theme.help};
+          `}
+        >
+          {' '}
+          tell us what went wrong
```
bpierre: Perhaps we could add a `{/* TODO: add link */}` comment here?

src/components/MessageCard.js
```diff
@@ -0,0 +1,81 @@
+import React from 'react'
+import { Box, GU, LoadingRing, textStyle, useTheme } from '@aragon/ui'
+
+export default function MessageCard({
+  icon,
+  title,
+  paragraph,
+  loading,
+  noBorder,
```
bpierre: What would you think of using `border` with a default value of `true`? I generally try to avoid negative names in params / props.

src/components/MessageCard.js
```diff
@@ -0,0 +1,81 @@
...
+      <div
+        css={`
+          margin: ${9 * GU}px auto;
+          display: flex;
+          flex-direction: column;
+          align-items: center;
+        `}
+      >
+        <img
+          src={icon}
```
bpierre:
```suggestion
          src={icon}
          alt=""
```

src/components/MessageCard.js
```diff
@@ -0,0 +1,81 @@
...
+          css={`
+            display: block;
+            width: 100%;
+            max-width: 237px;
```
bpierre: Would 30gu (240px) work here?

src/components/MessageCard.js
```diff
@@ -0,0 +1,81 @@
...
+                width: 30px;
+              `}
+            />
+            <span
+              css={`
+                margin-left: ${1 * GU}px;
+                ${textStyle('title2')}
+              `}
+            >
+              Loading...
```
bpierre:
```suggestion
              Loading…
```

src/components/MessageCard.js
```diff
@@ -0,0 +1,81 @@
...
+                ${textStyle('title2')}
+              `}
+            >
+              {title}
+            </span>
+            <div
+              css={`
+                ${textStyle('body2')}
+                color: ${theme.contentSecondary};
+                margin-top: ${GU * 1.5}px;
```
bpierre:
```suggestion
                margin-top: ${1.5 * GU}px;
```

src/components/MessageCard.js
```diff
@@ -0,0 +1,81 @@
...
+                width: 450px;
```
bpierre: Should we use `GU` here as well?

src/components/MessageCard.js
fabriziovigevani: Makes sense!

src/components/MessageCard.js
fabriziovigevani: It does!

src/hooks/useDisputes.js
```diff
@@ -11,30 +11,43 @@ import { convertToString } from '../types/dispute-status-types'
...
+    return {
+      disputes: disputes.map((dispute, i) => ({
         ...dispute,
         ...disputesPhases[i],
-      }))
-    }, [disputesPhases, disputes, disputesPhasesKey]), // eslint-disable-line react-hooks/exhaustive-deps
-  ]
+      })),
+    }
+  }, [disputesPhases, disputes, disputesPhasesKey, error]) // eslint-disable-line react-hooks/exhaustive-deps
```
sohkai: In cases where we do this, it might be nice to leave a comment explaining why we're not interested in a dependency.

src/components/Disputes/DisputeList.js
```diff
@@ -94,19 +114,62 @@ function DisputeList({ disputes, onSelectDispute }) {
...
+    <span>
+      The more{' '}
+      <span
+        css={`
+          color: ${theme.help};
+        `}
+      >
+        ANJ you activate
+      </span>
+      , more chances you have to be drafted to arbitrate a dispute.
```
sohkai: Copy suggestion:

```
, the more chances you ...
```

PR #71, opened 2020-02-02
Balance module loading states

fabriziovigevani:
Includes loading states for the balance module.

- Used `react-spring` that does a cool fade in effect when the balances are loaded.


src/components/Dashboard/Balance.js
```diff
@@ -46,106 +50,127 @@ const Balance = React.memo(function Balance({
...
+              <div
+                css={`
+                  ${textStyle('title3')}
+                  line-height: 1.2;
+                  display: flex;
+                  align-items: center;
+                `}
+              >
+                {splitAmount(formatUnits(amount, { digits: decimals }))}
+                <img height="20px" width="18px" src={ANJIcon} />
```
bpierre:
```suggestion
                <img height="20" width="18" src={ANJIcon} alt="ANJ" />
```

src/components/Dashboard/Loading.js
```diff
@@ -0,0 +1,29 @@
...
+      `}
+    >
+      <LoadingRing />
+      <span
+        css={`
+        ${textStyle('body1')}
+          margin-left: ${1 * GU}px;
+        `}
+      >
+        Loading...
```
bpierre:
```suggestion
        Loading…
```

src/hooks/useANJ.js
```diff
@@ -21,81 +21,102 @@ import { useConnectedAccount } from '../providers/Web3'
...
-      })),
+      }))
+    },
     [effectiveStatesKey, movements] //eslint-disable-line
   )
 }
 
 // Calculates the latest movement for each balance
-// In case the balance is the active, we must also calculate all non effective movements to get the effective active balance at current term
+// In case the balance is active or ianctive, we must also calculate all non effective movements to get the effective balance at current term
```
bpierre:
```suggestion
// In case the balance is active or inactive, we must also calculate all non effective movements to get the effective balance at current term
```

src/hooks/useANJ.js
```diff
@@ -21,81 +21,102 @@ import { useConnectedAccount } from '../providers/Web3'
...
 }
 
 // Asummes movements in descending order of creation
 function useConvertedMovements(movements) {
   const { currentTermId } = useCourtClock()
 
-  const effectiveStates = movements.map(mov =>
-    isMovementEffective(mov, currentTermId)
-  )
+  const effectiveStates = movements
```
bpierre: Can `movements` be falsy? It comes from here if I’m not mistaking: https://github.com/aragon/court-dashboard/pull/71/files#diff-30a133659d1207cf561a6e864720abecR87-R91

src/hooks/useANJ.js
```diff
@@ -21,81 +21,102 @@ import { useConnectedAccount } from '../providers/Web3'
...
+    : []
   const effectiveStatesKey = effectiveStates.join('')
 
   return useMemo(
-    () =>
-      movements.map((mov, i) => ({
+    () => {
+      if (!movements) {
+        return null
+      }
```
bpierre: Same as the previous comment, I don’t think `movements` can be falsy?

src/hooks/useANJ.js
```diff
@@ -143,15 +164,14 @@ function useBalanceWithMovements(balance, movements, balanceType) {
...
-        isMovementOf(acceptedMovements, anjMovementTypes[movement.type])
-      ),
-    [acceptedMovements, movements]
-  )
-
-  return filteredMovements
+  return useMemo(() => {
+    if (!movements) {
+      return null
+    }
```
bpierre: And same here, I think `movements` is always an array.

src/components/Dashboard/Balance.js
```diff
@@ -46,106 +50,127 @@ const Balance = React.memo(function Balance({
...
-        <div
-          css={`
-            display: flex;
-            align-items: flex-start;
-            padding-bottom: ${2 * GU}px;
-          `}
-        >
-          <div
+        {loading ? (
+          <Loading />
```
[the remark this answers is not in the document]
fabriziovigevani: Ha yes!! Huge improvement

src/hooks/useANJ.js
fabriziovigevani: It can be, see here https://github.com/aragon/court-dashboard/blob/ddfac8c289c4c27ce84b3936beb90fb2805cef16/src/hooks/subscription-hooks.js#L61

This is the case when we are still fetching balances

We only set the movements as empty array if we are done fetching and the user doesn't have any 24hs movements

bpierre: Oh right I missed that, thanks for the explanation!

PR #81, opened 2020-02-06
Inactive movements fix

rperez89:
Fix for [notion item 3](https://www.notion.so/Inactive-24h-message-is-the-same-as-the-active-wallet-message-it-shouldn-t-show-anything-if-the-ba-3f194d7022964b5d979f504d5f9f0359)


src/hooks/subscription-hooks.js
```diff
@@ -76,6 +77,33 @@ export function useJurorBalancesSubscription(jurorId) {
...
+    const movementsFiltered = movements => {
+      // If the juror is activating from the wallent we need to just show an incoming movement in the Active balance
+      return movements.reduce((previousList, movement, index, originalList) => {
+        const convertedMovement = {
+          ...movement,
+          effectiveTermId: parseInt(movement.effectiveTermId, 10),
+          amount: bigNum(movement.amount),
+        }
+        if (index === 0) {
+          previousList[0] = convertedMovement
```
fabriziovigevani: Here we could just do push()

```suggestion
          previousList.push(convertedMovement)
```

src/hooks/subscription-hooks.js
```diff
@@ -76,6 +77,33 @@ export function useJurorBalancesSubscription(jurorId) {
       movements = [],
     } = jurorData.juror || {}
 
+    const movementsFiltered = movements => {
+      // If the juror is activating from the wallent we need to just show an incoming movement in the Active balance
+      return movements.reduce((previousList, movement, index, originalList) => {
```
fabriziovigevani: Should we rename `previousList` to something with more meaning? 
Was thinking something like `accumulatorList`

src/hooks/subscription-hooks.js
```diff
@@ -76,6 +77,33 @@ export function useJurorBalancesSubscription(jurorId) {
...
+          effectiveTermId: parseInt(movement.effectiveTermId, 10),
+          amount: bigNum(movement.amount),
+        }
+        if (index === 0) {
+          previousList[0] = convertedMovement
+        } else if (
+          ANJMovement[originalList[index - 1].type] ===
+            ANJMovement.Activation &&
+          ANJMovement[movement.type] === ANJMovement.Stake &&
+          originalList[index - 1].createdAt === movement.createdAt
```
fabriziovigevani: What do you think of storing a variable for `originalList[index - 1]` like `previousMovement` and use it here so it's more clear?

src/hooks/subscription-hooks.js
```diff
@@ -76,6 +77,33 @@ export function useJurorBalancesSubscription(jurorId) {
...
+        ) {
+          previousList[previousList.length - 1] = {
+            ...previousList[previousList.length - 1],
+            type: 'StakeActivation',
+          }
+        } else {
+          previousList[previousList.length] = convertedMovement
```
fabriziovigevani: Same here we can do `.push()`
```suggestion
          previousList.push(convertedMovement)
```

src/hooks/subscription-hooks.js
```diff
@@ -76,6 +77,33 @@ export function useJurorBalancesSubscription(jurorId) {
...
+          previousList[0] = convertedMovement
+        } else if (
+          ANJMovement[originalList[index - 1].type] ===
+            ANJMovement.Activation &&
+          ANJMovement[movement.type] === ANJMovement.Stake &&
+          originalList[index - 1].createdAt === movement.createdAt
+        ) {
+          previousList[previousList.length - 1] = {
+            ...previousList[previousList.length - 1],
+            type: 'StakeActivation',
```
fabriziovigevani: If we move this function to the file we could also store `StakeActivation` in a const variable

src/hooks/subscription-hooks.js
bpierre: I think we could use only one condition (also taking your suggestions @fabriziovigevani):

```jsx
  return movements.reduce((movements, movement, index, prevMovements) => {
    if (
      index > 0 &&
      ANJMovement[prevMovements[index - 1].type] === ANJMovement.Activation &&
      ANJMovement[movement.type] === ANJMovement.Stake &&
      prevMovements[index - 1].createdAt === movement.createdAt
    ) {
      movements[movements.length - 1].type = 'StakeActivation'
      return movements
    }

    movements.push({
      ...movement,
      effectiveTermId: parseInt(movement.effectiveTermId, 10),
      amount: bigNum(movement.amount),
    })

    return movements
  })

```

PR #82, opened 2020-02-06
any time jury drafting

rperez89:


src/components/Disputes/DisputePhase.js
```diff
@@ -63,7 +63,11 @@ function DisputePhase({ phase, nextTransition }) {
 function DisplayTime({ phase, nextTransition }) {
   const theme = useTheme()
 
-  if (phase === Phase.ExecuteRuling || phase === Phase.ClaimRewards) {
+  if (
+    phase === Phase.ExecuteRuling ||
+    phase === Phase.ClaimRewards ||
+    phase === Phase.JuryDrafting
+  ) {
```
fabriziovigevani: Since now we don't use the timer for the jury-drafting phase, does it make sense to remove the `juryDraftingTerms` variable at `dispute-utils`? Also, don't calculate the `nextTranstiion` attribute in case we are returning this phase ?

PR #83, opened 2020-02-06
Balance: update probability banner for small percentages

rperez89:
This is the fix for [notion item](https://www.notion.so/Adjust-the-wording-it-was-saying-High-probability-of-being-drafted-when-at-24-Need-to-review-th-9993188d45f048298196d9d34806aec5)

need to be checked by patty


src/components/Dashboard/AccountBanner.js
```diff
@@ -186,14 +184,15 @@ const BannerWithProbability = ({ activeBalance }) => {
...
     >
       <span css="margin-right: 8px">
+        You’ll be drafted, on average
         <span
           css={`
             color: ${theme.accent};
           `}
         >
-          {probabilityText} probability{' '}
+          {` 1 on every ${chances} times `}
```
sohkai:
```suggestion
          {` 1 on every ${chances} disputes `}
```

src/components/Dashboard/AccountBanner.js
```diff
@@ -176,7 +174,7 @@ const BannerWithProbability = ({ activeBalance }) => {
   // Calculate probability (since the total active balance is asynconous
   // it can happen that it has not been updated yet when the juror active balance has)
   const draftingProbability = Math.min(1, totalPercentage / 100)
-  const probabilityText = getProbabilityText(draftingProbability)
+  const chances = totalPercentage > 0 ? Math.floor(100 / totalPercentage) : 0
```
sohkai: We most likely want a better number than falling back to `0`. This is going to be wildly confusing :)

Screen Shot 2020-02-13 at 12 32 18 AM

src/components/Dashboard/AccountBanner.js
```diff
@@ -186,14 +184,15 @@ const BannerWithProbability = ({ activeBalance }) => {
       `}
     >
       <span css="margin-right: 8px">
+        You’ll be drafted, on average
```
sohkai:
```suggestion
        On average, you'll be drafted
```

src/components/Dashboard/AccountBanner.js
```diff
@@ -186,14 +184,15 @@ const BannerWithProbability = ({ activeBalance }) => {
...
         <span
           css={`
             color: ${theme.accent};
           `}
         >
-          {probabilityText} probability{' '}
+          {` 1 on every ${chances} times `}
         </span>
-        to be drafted
+        during this term
```
sohkai: I wonder if "during this term" is really needed; it makes it sound like there are actually going to be X disputes created this term.

src/components/Dashboard/AccountBanner.js
```diff
@@ -85,8 +83,6 @@ function AccountBanner({ status, loading, minActiveBalance, activeBalance }) {
     theme
   )
 
-  if (loading) return <Loading height={48} />
```
fabriziovigevani: I think we should keep a loading ring here because otherwise it's going to show an incorrect state to the juror every time he/she loads the dashboard

src/components/Dashboard/AccountBanner.js
```diff
@@ -185,28 +184,61 @@ const BannerWithProbability = ({ activeBalance }) => {
...
     </div>
   )
 
   const paragraph =
     'The more ANJ you activate, more chances you have to be drafted to arbitrate a dispute'
-  return (
+
+  // TODO - change this for the loading indicator once ready
+  return fetchingTotalBalance ? (
+    <LoadingRing />
```
fabriziovigevani: Could we wrap this in a div of `{6 * GU}px` height ? Otherwise the balances misalign when this loading is rendered

PR #85, opened 2020-02-06
Sentry integration

bpierre:


now.json
```diff
@@ -0,0 +1,7 @@
+{
+  "build": {
+    "env": {
+      "REACT_APP_SENTRY_DSN": "@court-dashboard-sentry-dsn"
```
bpierre: @sohkai I suspect this will get included on every build, including the preview ones (PRs). Do you know how we could restrict it to a specific deployment?

sohkai: Hmm, looks like [we might need to use a Github secret](https://spectrum.chat/zeit/now/env-on-master-only~263e2488-3531-434a-a0af-07a2ea4bce7a?m=MTU0NzIxNDk2MDc5NQ==)?

bpierre: Ah yes that’s what I wanted to do at first, but I don’t like to be so tied to Now, e.g. if we want to deploy on IPFS.

Instead I added a new env var, `ENABLE_SENTRY`, based on the local branch. I think that’s more future proof, what do you think?

sohkai: Yes, good point.

What do you think about turning that into a script we call instead of directly adding it to the package.json? It's getting _just_ a bit unwieldy ;).

bpierre: Yes! Done https://github.com/aragon/court-dashboard/pull/85/commits/f7c22da0ef7e062a018a939ba3c6e8bb639e828d

PR #86, opened 2020-02-07
Rewards module

fabriziovigevani:
Includes rewards module and appeal collateral. 

For now to claim the rewards, we are firing every required tx sequentially. Haven't incorporated the signer panel yet. 

Thought that it would be a good idea to have the signer panel once we have all actions incorporated, which with this PR I think we will :)


src/components/Dashboard/Dashboard.js
```diff
@@ -70,7 +78,27 @@ function Dashboard() {
         <Welcome />
       )}
 
-      <Tasks onlyTable />
+      <Split
+        primary={<Tasks onlyTable />}
+        secondary={
+          connectedAccount && (
```
bpierre: Should we provide an empty state, or remove the `Split` when the user is not connected? So that it doesn’t create an empty space on the left side.

src/components/Dashboard/AppealColateralModule.js
```diff
@@ -0,0 +1,80 @@
...
+            `}
+          >
+            <div
+              css={`
+                margin-bottom: ${1 * GU}px;
+              `}
+            >
+              <img
+                src={iconLock}
+                height="14"
```
bpierre:
```suggestion
                height="14"
                alt=""
```

src/components/Dashboard/NoRewards.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react'
+import noDraftSVG from '../../assets/noDraft.svg'
```
bpierre: We should move this SVG import below the JS imports.

src/components/Dashboard/RewardsModule.js
```diff
@@ -0,0 +1,343 @@
...
+      // Claim all arbitrable fee rewards
+      for (const arbitrableFee of rewards.arbitrableFees) {
+        const { disputeId, rounds } = arbitrableFee
+        for (const roundId of rounds) {
+          const rewardTx = await onSettleReward(
+            disputeId,
+            roundId,
+            connectedAccount
+          )
+          await rewardTx.wait()
```
bpierre: I think we should use `Promise.all()` here, to do this in parallel.

src/components/Dashboard/RewardsModule.js
```diff
@@ -0,0 +1,343 @@
...
+        }
+      }
+
+      // Claim all appeal fee rewards
+      for (const appealFee of rewards.appealFees) {
+        const { disputeId, rounds } = appealFee
+        for (const roundId of rounds) {
+          const rewardTx = await onSettleAppealDeposit(disputeId, roundId)
+          await rewardTx.wait()
```
bpierre: Same here, we should probably use `Promise.all()`.

src/components/Dashboard/RewardsModule.js
```diff
@@ -0,0 +1,343 @@
...
+
+  return (
+    <Box
+      heading={showHeading && 'Rewards'}
+      padding={hasRewardsToClaim ? 0 : 3 * GU}
+    >
+      {(() => {
+        if (loading) return <Loading height={150} />
+
+        if (!hasRewardsToClaim) return <NoRewards />
```
bpierre: Just to be explicit :)

```suggestion
        if (!hasRewardsToClaim) return <NoRewards />
        return null
```

src/components/Disputes/DisputeInfo.js
```diff
@@ -166,7 +166,7 @@ const DisputeInfo = React.memo(function({
             onDraft={onDraft}
             onRequestCommit={onRequestCommit}
             onRequestReveal={onRequestReveal}
-            onLeavk={onLeak}
+            onLeak={onLeak}
```
bpierre: :grimacing:

src/hooks/useJurorRewards.js
```diff
@@ -0,0 +1,166 @@
...
+
+  return disputeFees
+}
+
+/**
+ * Converts Map into Array
+ * @param {Map} feeMap Map to convert to an array
+ * @returns {Array} New array containing map data
+ */
+function mapToArray(feeMap) {
```
bpierre: Should we call this `feeMapToArray`?

fabriziovigevani: Yes, good idea.

src/components/Dashboard/Dashboard.js
fabriziovigevani: Ah yes, good call! Until we don't have the dispute total status on the dashboard i think it makes sense

src/components/Dashboard/RewardsModule.js
```diff
@@ -0,0 +1,346 @@
...
+const RewardsModule = React.memo(function RewardsModule({
+  rewards,
+  loading,
+  onSettleReward,
+  onSettleAppealDeposit,
+}) {
+  const connectedAccount = useConnectedAccount()
+
+  // Form submission
+  const handleFormSubmit = async event => {
```
sohkai: This looks like it should be in a `useCallback()`

src/components/Dashboard/RewardsModule.js
```diff
@@ -0,0 +1,346 @@
...
+      // Claim all appeal fee rewards
+      const appealTxs = []
+      for (const appealFee of rewards.appealFees) {
+        const { disputeId, rounds } = appealFee
+        for (const roundId of rounds) {
+          appealTxs.push(await onSettleAppealDeposit(disputeId, roundId))
+        }
+      }
+
+      await Promise.all(appealTxs.map(tx => tx.wait()))
```
sohkai: Wondering if it would make sense to push both the `arbitrableTxs` and `appealTxs` into a single array and then do `Promise.all()`.

Right now if you had both, you would have to wait for the first set to get mined before the next set would appear, right?

fabriziovigevani: Yes totally makes sense, will add this change along the one above in the next PR which includes the additional tx from claiming the fees from the treasury

PR #87, opened 2020-02-07
Evidence fetch

rperez89:
There is still some work to do given that the data attribute from the Evidence type of the subgraph is Bytes, and i couldn't find a way to transform it to String so when the data inside the evidence data field is the ipfs hash when i try to generate the endpoint as a string  the result is not a string, however if you do a  `type of` the data attribute that comes from the graph is a String, so not sure what is going on.
So what i thought was that we should add a stringData to the Evidence type in the graph like this :
```
type Evidence @entity {
  id: ID!
  arbitrable: Arbitrable!
  dispute: Dispute!
  data: Bytes!
  stringData: String!
  submitter: Bytes!
  createdAt: BigInt!
}
```

and inside the Arbitrable.ts file do 

```
export function handleEvidenceSubmitted(event: EvidenceSubmitted): void {
  let id = event.transaction.hash.toHex() + event.logIndex.toHex()
  let evidence = new Evidence(id)
  evidence.arbitrable = event.address.toHex()
  evidence.dispute = event.params.disputeId.toString()
  evidence.data = event.params.evidence
  evidence.stringData = event.params.evidence.toString()
  evidence.submitter = event.params.submitter
  evidence.createdAt = event.block.timestamp
  evidence.save()
}
```

so when i tried that everything worked perfect i was getting a string in the frontend to generate the endpoint, but if the data was indeed a string like "HELLO THIS IS A STRING" for some reason doing the `evidence.stringData = event.params.evidence.toString()` ended up generating 5 evidences with data : HELLO  the other one THIS the other one IS etc, etc.

From the graph docs 

```
The Bytes class extends AssemblyScript's Uint8Array and this supports all the Uint8Array functionality, plus the following new methods:

b.toHex() – returns a hexadecimal string representing the bytes in the array
b.toString() – converts the bytes in the array to a string of unicode characters
b.toBase58() – turns an Ethereum Bytes value to base58 encoding (used for IPFS hashes)
```


fabriziovigevani:
I think we could just change the `data` attribute of the schema to be a `string`. Unless i'm not seeing where else we would need it as bytes.

> but if the data was indeed a string like "HELLO THIS IS A STRING" for some reason doing the evidence.stringData = event.params.evidence.toString() ended up generating 5 evidences with data : HELLO the other one THIS the other one IS etc, etc.

Could this be due to how the cli works that is actually creating 5 evidences?

```
`npm run rpc:dispute -- -s ${arbitrable} -m 'Testing dispute #${i}' -e "HELLO THIS IS A STRING" 
````

**Comment by bingen** [comment]
I'm not sure if understand the problem. If what comes in the evidence bytes field is a regular URL, then the conversion works fine, but if it's an IPFS hash then it doesn't. Is that the problem? Can you point me to the code that does that conversion? Thanks!

**Comment by bingen** [comment]
> Could this be due to how the cli works that is actually creating 5 evidences?
> 
> ```
> `npm run rpc:dispute -- -s ${arbitrable} -m 'Testing dispute #${i}' -e "HELLO THIS IS A STRING" 
> ```

That would be the case if you don't enclose it with "", like:
```
`npm run rpc:dispute -- -s ${arbitrable} -m 'Testing dispute #${i}' -e HELLO THIS IS A STRING
```
But I just tested and with quotes it works fine.

File: src/components/Disputes/DisputeDetail.js [hunk 1]
```diff
@@ -35,6 +35,8 @@ const DisputeDetail = React.memo(function DisputeDetail({ match }) {
     requests,
   } = useDisputeLogic(disputeId)
 
+  const creatorAddress = dispute && dispute.subject && dispute.subject.id
```
bpierre: I think [we can use optional chaining](https://github.com/babel/babel/issues/10809) now! \o/

```suggestion
  const creatorAddress = dispute?.subject?.id
```

src/components/Disputes/DisputeEvidences.js
```diff
@@ -1,113 +1,141 @@
...
-                        margin-bottom: ${2 * GU}px;
-                      `}
-                    >
-                      Date
-                    </h2>
+                    <img src={folderIcon} width="17" height="20" alt="" />
                     <span
                       css={`
-                        ${textStyle('body2')};
+                        margin-left: ${GU * 1.5}px;
```
bpierre:
```suggestion
                        margin-left: ${1.5 * GU}px;
```

src/hooks/subscription-hooks.js
```diff
@@ -119,6 +119,8 @@ export function useSingleDisputeSubscription(id) {
     variables: { id },
   })
 
+  console.log('SINGLE DISPUTE ', data)
```
bpierre: Should we remove this?

src/hooks/useEvidences.js
```diff
@@ -0,0 +1,64 @@
...
+  useEffect(() => {
+    async function getEvidenceData() {
+      const evidenceFetchedData = await Promise.all(
+        evidences.map(async evidence => {
+          const { data: evidenceData, submitter, createdAt } = evidence
+
+          // check if the metadata inside the evidence is a cid
+          if (isIPFS.multihash(evidenceData) || isIPFS.cid(evidenceData)) {
+            const { data, error } = await ipfsGet(evidenceData)
+            if (!error) {
```
bpierre: We could do the opposite to remove one indentation:

```jsx
if (error) {
  // If there was an error fetching the ipfs content return error
  return { error: ERROR_TYPES.ERROR_FETCHING_IPFS }
}
if (typeof data === 'object') {
  // …
}
```

src/hooks/useEvidences.js
```diff
@@ -0,0 +1,64 @@
...
+              if (typeof data === 'object') {
+                // we are just interested in those keys
+                return {
+                  metadata: data.metadata || '',
```
sohkai: It'd be great to add these (for now) to a `docs/metadata.md` file to keep track of what keys we're assuming.

docs/metadata.md
```diff
@@ -0,0 +1,73 @@
...
+## Dispute
+
+Structure from the graph
+
+```javascript
+type Dispute @entity {
+  id: ID!
+  ...
+  subject: Arbitrable!
+  metadata: String!
```
sohkai:
```suggestion
  metadata: Bytes!
```

docs/metadata.md
```diff
@@ -0,0 +1,73 @@
...
+  id: ID!
+  disputes: [Dispute!] @derivedFrom(field: "subject")
+}
+```
+
+In the frontend we try to parse the `metadata` as a json object.
+
+### metadata - Json object
+
+We are going to try to find the following keys:
```
sohkai: We should update this: if the `metadata` is a json, we now only expect:

```js
1. `description`
2. `metadata`
```

Where `metadata` contains:

```
1. `agreementText` (required)
2. `description`
3. `defendant`
4. `plaintiff`
```

docs/metadata.md
```diff
@@ -0,0 +1,73 @@
...
+
+1. `description`
+2. `agreementText`
+3. `defendant`
+4. `plaintiff`
+
+#### metadata - String
+
+In case that the metadata is just a plain string we are going to use it as the dispute description and the dispute creator `subject.id` will be considered as the plaintiff
```
sohkai:
```suggestion
In case `metadata` is a plain string, we assume it to be the dispute's description and set the dispute creator (`subject.id`) as the plaintiff. No agreement link will be available.
```

docs/metadata.md
```diff
@@ -0,0 +1,73 @@
...
+type Evidence @entity {
+  id: ID!
+  ...
+  data: String!
+  submitter: Bytes!
+  ...
+}
+```
+
+The dashboard check if the `data` field is a valid IPFS hash/cid
```
sohkai:
```suggestion
The dashboard checks if the `data` field is a valid IPFS hash/cid, JSON, or string.
```

docs/metadata.md
```diff
@@ -0,0 +1,73 @@
...
+
+1. `metadata`
+
+If we can find it we are going to show that as the evidence text
+
+#### content - String
+
+Show the content as the evidence text
+
+### data - Plain String
```
sohkai: Is this the case where `toUtf8()` errors?

In that case we should just say the data is un-parsable.

docs/metadata.md
```diff
@@ -0,0 +1,73 @@
...
+  submitter: Bytes!
+  ...
+}
+```
+
+The dashboard check if the `data` field is a valid IPFS hash/cid
+
+### data - IPFS hash
+
+Try to fetch the content from IPFS
```
sohkai:
```suggestion
Try to fetch the content from IPFS and continue with one of the below options.
```

docs/metadata.md
```diff
@@ -0,0 +1,73 @@
...
+
+#### content - Json
```
sohkai: Let's throw this case away; having a JSON object may be useful but isn't right now.

docs/metadata.md
```diff
@@ -0,0 +1,73 @@
...
+In case that the metadata is just a plain string we are going to use it as the dispute description and the dispute creator `subject.id` will be considered as the plaintiff
+
+
+## Evidence
+
+```javascript
+type Evidence @entity {
+  id: ID!
+  ...
+  data: String!
```
sohkai:
```suggestion
  data: Bytes!
```

docs/metadata.md
```diff
@@ -0,0 +1,73 @@
...
+  submitter: Bytes!
```
sohkai:
```suggestion
  submitter: ID!
```

src/types/dispute-status-types.js
```diff
@@ -1,3 +1,7 @@
+export const ERROR_TYPES = {
+  ERROR_FETCHING_IPFS: 'Error fetching ipfs content',
```
sohkai: Should this also follow the symbol / string mapping convention used for other errors?

src/types/evidences-status-types.js
```diff
@@ -0,0 +1,4 @@
+export const ERROR_TYPES = {
```
sohkai: Similar to above; should we follow the same formula?

src/lib/ipfs-utils.js
```diff
@@ -0,0 +1,16 @@
+import { IPFS_ENDPOINT } from '../endpoints'
+
+const REQUEST_TIMEOUT = 60000
+
+export const ipfsGet = async hash => {
```
sohkai:
```suggestion
export const ipfsGet = async cid => {
```

Is a bit more semantically correct :).

src/lib/ipfs-utils.js
```diff
@@ -0,0 +1,16 @@
...
+  const endpoint = `${IPFS_ENDPOINT}/${hash}`
+  try {
+    const result = await fetch(endpoint, { timeout: REQUEST_TIMEOUT })
+    const data = await result.text()
+
+    return { data, error: !result.ok }
+  } catch (err) {
+    console.error('Error getting data from IPFS', err)
```
sohkai:
```suggestion
    console.error(`Error requesting data from IPFS for ${endpoint}`, err)
```

src/hooks/useDisputes.js
```diff
@@ -38,37 +40,99 @@ export default function useDisputes() {
...
       dispute: {
-        ...dispute,
-        ...processDisputeData(dispute),
+        ...disputeProcessed,
         ...disputePhase,
       },
       fetching,
     }
-  }, [dispute, disputePhaseKey]) // eslint-disable-line react-hooks/exhaustive-deps
+  }, [disputeProcessed, dispute, disputePhaseKey]) // eslint-disable-line react-hooks/exhaustive-deps
```
sohkai: Still wondering on this; which dependency are we not interested in having here? And why?

docs/metadata.md
rperez89: yeah that's the case where the evidence metadata is not a valid `cid` or does not follow the `ipfs:XXXXX`

src/hooks/useDisputes.js
```diff
@@ -38,37 +40,99 @@ export default function useDisputes() {
...
+          description: disputeDescription,
+          ...disputesPhases[i],
+        }
+      }),
     }
   }, [disputesPhases, disputes, disputesPhasesKey, error]) // eslint-disable-line react-hooks/exhaustive-deps
 }
 
 export function useDispute(disputeId) {
+  const [disputeProcessed, setDisputeProcessed] = useState()
```
sohkai: Should we set an empty object or an initial object here?

src/hooks/useDisputes.js
```diff
@@ -38,37 +40,99 @@ export default function useDisputes() {
...
   const courtConfig = useCourtConfig()
   const now = useNow() // TODO: use court clock
   const { dispute, fetching } = useSingleDisputeSubscription(disputeId)
+  useEffect(() => {
+    const fetchDataFromIpfs = async () => {
+      if (dispute) {
```
sohkai: Perhaps we can do `if (!dispute) { return }` to avoid the further indentation later.

src/hooks/useDisputes.js
```diff
@@ -38,37 +40,99 @@ export default function useDisputes() {
...
+        const ipfsPath = disputeMetadata.replace(/^ipfs:/, '')
+
+        if (isIPFS.cidPath(ipfsPath)) {
+          const { data, error } = await ipfsGet(ipfsPath)
+          if (error) {
+            return setDisputeProcessed({ ...dispute, error: true })
+          }
+          try {
+            const parsedDisputeData = JSON.parse(data)
+            const agreementText = parsedDisputeData.agreementText.replace(
```
sohkai: Ahh, we don't have to do this manually; we should be able to use:

```
const agreementUrl = resolvePathname(
  agreementText,
  `${IPFS_ENDPOINT}/${ipfsPath}/`
)
```

**Note** the end slash at the end :).

------

But we should bail on the agreement text if it's not available (right now this would just throw)

src/hooks/useEvidences.js
```diff
@@ -0,0 +1,74 @@
...
+      if (cancelled) {
+        return
+      }
+
+      setEvidenceProcessed(prev =>
+        [
+          ...prev,
+          evidence,
+        ].sort(({ createdAt: dateLeft }, { createdAt: dateRight }) =>
+          dayjs(dateLeft).isAfter(dayjs(dateRight))
```
sohkai: We should be able to just do `dateLeft - dateRight` if they're both dates.

src/hooks/useEvidences.js
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    evidences.forEach(async evidence => {
+      const { data: evidenceData, submitter, createdAt } = evidence
+
+      const ipfsHash = evidenceData.replace(/^ipfs:/, '')
+
+      // check if the metadata inside the evidence is a cid
+      // if the ipfs tag is not present return the metadata as plain string
+      if (isIPFS.multihash(ipfsHash) || isIPFS.cid(ipfsHash)) {
```
sohkai: Let's move the `evidenceData.replace(/^ipfs:/, '')` and `isIPFS.cidPath(ipfsHash)` into a util that returns the hash or an empty string, so we can abstract this out for metadata strings.

I think `isIPFS.cidPath()` should cover all our cases, so no need to do the separate checks for multihash or cid.

src/hooks/useEvidences.js
```diff
@@ -0,0 +1,74 @@
...
+        const { data, error } = await ipfsGet(ipfsHash)
+        if (error) {
```
sohkai: As we handle generic errors later, it may read better if we move this to be after the `if (data)`

src/hooks/useDisputes.js
rperez89: Could you explain what do you mean by " dont have to do this manually?" 
which would be the difference with ```const agreementUrl = resolvePathname(
            agreementText,
            `${IPFS_ENDPOINT}/${ipfsPath}`
          )```
besides the `/` at the end?

sohkai: Having the `/` at the end makes `resolvePathname()` behave differently.

```
resolvePathname('a', 'b') // = 'a'
resolvePathname('a', 'b/') // = 'b/a'
resolvePathname('./a', 'b/') // = 'b/a'
resolvePathname('../a', 'b/c') // 'a'
resolvePathname('../a', 'b/c/') // 'b/a'
```

So we don't need do the manual `.replace()` here, because `resolvePathname()` will take care of it for us.

docs/metadata.md
sohkai: In that case, can we rename the section to `data - Unknown String`?

PR #89, opened 2020-02-07
Dispute detail: Final outcome and round number

fabriziovigevani:
Adds dispute final outcome and round number to info section.

Also changed `Created by` label to `Plantiff`

Since we now use the outcome text in the timeline and in this info section as well, i made it a separate component. Same goes for the `RoundPill`


src/components/Disputes/DisputeInfo.js
```diff
@@ -45,121 +55,44 @@ const DisputeInfo = React.memo(function({
...
+                  value={<DisputeRoundPill roundId={dispute.lastRoundId} />}
+                  isNode
+                />
+              </Row>
             )}
-          </div>
+            <Row>
+              <Field label="Description" value={description} />
+              {creatorAddress && (
+                <Field label="Plantiff" value={creatorAddress} />
```
bpierre:
```suggestion
                <Field label="Plaintiff" value={creatorAddress} />
```

src/components/Disputes/DisputeInfo.js
```diff
@@ -176,6 +109,122 @@ const DisputeInfo = React.memo(function({
...
+            )}
+          </h1>
+          {transaction && <TransactionBadge transaction={transaction} />}
+        </div>
+      </div>
+    </div>
+  )
+}
+
+function Field({ label, value, isNode }) {
```
bpierre: I notice that `isNode` is used only once: https://github.com/aragon/court-dashboard/pull/89/files#diff-e447e120bfda3c5da3d624e3174f5292R77-R81

But `DisputeRoundPill` is defining its own text styles: https://github.com/aragon/court-dashboard/pull/89/files#diff-44c96f491e62b75470988ff5bf1a3fb9R18-R19

So I think we could drop the `isNode` prop altogether, and always return `value` wrapped in the `textStyle('body2')` element? We could also make it a `div` rather than a span, to accept block elements as well.

src/components/Disputes/DisputeInfo.js
```diff
@@ -176,6 +109,122 @@ const DisputeInfo = React.memo(function({
...
+      >
+        <div
+          css={`
+            background: linear-gradient(
+              233deg,
+              ${theme.accentEnd} -50%,
+              ${theme.accentStart} 91%
+            );
+            border-radius: 50%;
+            padding: 12px;
```
bpierre:
```suggestion
            padding: ${1.5 * GU}px;
```

src/components/Disputes/DisputeRoundPill.js
```diff
@@ -0,0 +1,29 @@
...
+        padding: 1px 16px;
+        border-radius: 100px;
+        background: linear-gradient(
+          13.81deg,
+          rgba(255, 179, 109, 0.3) -0.55%,
+          rgba(255, 136, 136, 0.3) 88.44%
+        );
+        text-transform: uppercase;
+        font-size: 12px;
+        color: #e9756c;
```
bpierre: Shouldn’t this be coming from the theme?

src/components/Disputes/DisputeRoundPill.js
```diff
@@ -0,0 +1,29 @@
...
+
+  const label = `Round ${numberToWord(roundId)}`
+  return (
+    <span
+      css={`
+        padding: 1px 16px;
+        border-radius: 100px;
+        background: linear-gradient(
+          13.81deg,
+          rgba(255, 179, 109, 0.3) -0.55%,
```
bpierre:
```suggestion
          rgba(255, 179, 109, 0.3) 0,
```

src/components/Disputes/DisputeRoundPill.js
```diff
@@ -0,0 +1,29 @@
...
+          rgba(255, 136, 136, 0.3) 88.44%
```
bpierre:
```suggestion
          rgba(255, 136, 136, 0.3) 88%
```

src/components/Disputes/DisputeRoundPill.js
```diff
@@ -0,0 +1,29 @@
...
+import { numberToWord } from '../../lib/math-utils'
+
+function DisputeRoundPill({ roundId }) {
+  if (roundId === undefined) return null
+
+  const label = `Round ${numberToWord(roundId)}`
+  return (
+    <span
+      css={`
+        padding: 1px 16px;
```
bpierre:
```suggestion
        padding: 1px ${2 * GU}px;
```

src/components/Disputes/DisputeRoundPill.js
```diff
@@ -0,0 +1,29 @@
...
+        border-radius: 100px;
+        background: linear-gradient(
+          13.81deg,
```
bpierre:
```suggestion
          14deg,
```

src/components/Disputes/DisputeRoundPill.js
```diff
@@ -0,0 +1,29 @@
...
+          rgba(255, 179, 109, 0.3) -0.55%,
+          rgba(255, 136, 136, 0.3) 88.44%
+        );
+        text-transform: uppercase;
+        font-size: 12px;
```
bpierre: Should we use [`label2`](https://github.com/aragon/aragon-ui/blob/master/src/style/text-styles.js#L50-L54) here?

src/components/Disputes/DisputeInfo.js
fabriziovigevani: Ah yes good call 👍

src/components/Disputes/DisputeRoundPill.js
fabriziovigevani: Yes, definitely

PR #91, opened 2020-02-07
Pre-launch copy changes

izqui:
I will be adding more things as PRs continue landing

### DIFF
--- a/src/components/Dashboard/AccountBanner.js
+++ b/src/components/Dashboard/AccountBanner.js
@@ -58,7 +58,7 @@ const getBannerAttributes = (
     title: 'Activate ANJ to be an active juror',
     paragraph: `You must activate at least ${formatUnits(minActiveBalance, {
       digits: decimals,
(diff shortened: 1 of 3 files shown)


src/components/Dashboard/AccountBanner.js
```diff
@@ -58,7 +58,7 @@ const getBannerAttributes = (
     title: 'Activate ANJ to be an active juror',
     paragraph: `You must activate at least ${formatUnits(minActiveBalance, {
       digits: decimals,
-    })}  ANJ to be drafted as a juror`,
+    })}  ANJ to be elegible for drafting`,
```
sohkai: @dizzypaty and I were discussing this; perhaps we could use "You must activate at least 10,000 ANJ to participate as a juror"?

dizzypaty: Yeah, that sounds good to me!

PR #93, opened 2020-02-09
Dispute info changes

rperez89:


src/components/Disputes/DisputeInfo.js
```diff
@@ -25,6 +25,7 @@ import iconCourt from '../../assets/courtIcon.svg'
 import { addressesEqual } from '../../lib/web3-utils'
 import DisputeRoundPill from './DisputeRoundPill'
 import DisputeOutcomeText from './DisputeOutcomeText'
+import { Link } from 'react-router-dom'
```
fabriziovigevani: I think this should be imported from `@aragon/ui` ? 😄

rperez89: hahah yes , it was an auto import good catch

src/components/Disputes/DisputeInfo.js
```diff
@@ -62,32 +63,62 @@ const DisputeInfo = React.memo(function({
...
+                          <DisputeRoundPill roundId={dispute.lastRoundId} />
+                        }
+                      />
+                    </>
+                  )
+                }
+                return (
+                  <>
+                    <Field label="Description" value={description} />
+                    <div />
```
sohkai: Is this empty div to force a flex layout or for spacing? It seems a bit odd just hanging out here.

src/dispute-logic.js
```diff
@@ -64,10 +64,29 @@ export function useDisputeLogic(disputeId) {
...
+export function processDisputeData(dispute) {
+  if (!dispute?.metadata) {
+    return { ...dispute }
+  }
+  try {
+    const parsedDisputeData = JSON.parse(dispute.metadata)
+    // If the description tag is not present in metadata we assume that the metadata itself is the description
+    return {
+      ...dispute,
+      description: parsedDisputeData.description || dispute.metadata,
```
sohkai: I think this can just be `|| ''`; you can't use `JSON.parse()` on just a string.

src/components/Disputes/DisputeInfo.js
rperez89: yeah that is because [here](https://github.com/aragon/court-dashboard/pull/93/files/49ff56e7fe098d5d51cb624566a9e7e1e155b4e3#diff-e447e120bfda3c5da3d624e3174f5292R248) we are setting always the grid template colums as `1fr 1fr 1fr` and if we only need to show the  first element and the last one i did that trick, i wanted to finish it instead of waste time trying to do it with css, but @bpierre was notified by me about that and will check how we can do it

sohkai: Cool! Might just be worth some comments in the future :).

PR #100, opened 2020-02-10
Add Activity Panel

bpierre:


src/components/Activity/TimeTag.js
```diff
@@ -0,0 +1,52 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { GU, textStyle, useTheme } from '@aragon/ui'
+// import { formatDistance, formatDistanceStrict } from 'date-fns'
```
rperez89: are we planning to import those functions from `dayjs` now?

src/lib/web3-utils.js
```diff
@@ -110,7 +113,7 @@ export function getNetworkName(chainId) {
 }
 
 export function isLocalOrUnknownNetwork(chainId) {
-  return getNetworkName(chainId) === DEFAULT_LOCAL_CHAIN
+  return getNetworkType(chainId) === DEFAULT_LOCAL_CHAIN
```
rperez89: let's also do the change [here](https://github.com/aragon/court-dashboard/blob/fa739200349b326cf4182daddc9bfb0ef10ec01e/src/components/Onboarding/OnboardingModal.js#L10)

src/components/Activity/TransactionProgress.js
```diff
@@ -0,0 +1,91 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { ProgressBar, GU } from '@aragon/ui'
+import { Transition, animated } from 'react-spring/renderprops'
+import { ActivityStatusType } from './prop-types'
+import { norm } from '../../lib/math-utils'
+import useNow from '../../hooks/useNow'
+import { ACTIVITY_STATUS_CONFIRMED, ACTIVITY_STATUS_PENDING } from './symbols'
```
fabriziovigevani: Should we delete `./symbols` and  import these from the `Activity` provider instead (or import symbols from the provider) ? Otherwise i think the progress bar would never show 😄

bpierre: Yes! I removed them from the `Activity` provider instead :-)

src/lib/web3-utils.js
```diff
@@ -99,7 +102,7 @@ export function shortenAddress(address, charsLength = 4) {
   )
 }
 
-export function getNetworkName(chainId) {
+export function getNetworkType(chainId = env('CHAIN_ID')) {
```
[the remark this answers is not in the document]
bpierre: Yes, we generally use “network type” for identifiers like “mainnet”, “rinkeby” etc., and “network name” for the human-readable ones.

src/hooks/useCourtContracts.js
```diff
@@ -70,38 +69,48 @@ export function useANJActions() {
   // activate ANJ directly from available balance
   const activateANJ = useCallback(
     amount => {
-      return jurorRegistryContract.activate(amount, { gasLimit: GAS_LIMIT })
+      return addTransactionActivity(
+        jurorRegistryContract.activate(amount, { gasLimit: GAS_LIMIT }),
+        `Activate ${String(amount)} ANJ`
```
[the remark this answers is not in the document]
bpierre: Yes! I changed a bit the format so we only store the dynamic parameters, but this is fixed now!

src/lib/web3-utils.js
bpierre: Done! https://github.com/aragon/court-dashboard/blob/activity/src/components/Onboarding/OnboardingModal.js#L10

src/components/Activity/TimeTag.js
bpierre: Yes! https://github.com/aragon/court-dashboard/pull/100/files#diff-3f247f7e3ba0dcef7287f99eda17ace6R11-R12

src/lib/web3-utils.js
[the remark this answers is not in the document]
fabriziovigevani: Makes sense!

src/components/Activity/ActivityItem.js
```diff
@@ -0,0 +1,242 @@
...
+  useTheme,
+  IdentityBadge,
+} from '@aragon/ui'
+import { transformAddresses, getNetworkType } from '../../lib/web3-utils'
+import {
+  ACTIVITY_STATUS_PENDING,
+  ACTIVITY_STATUS_CONFIRMED,
+  ACTIVITY_STATUS_FAILED,
+  ACTIVITY_STATUS_TIMED_OUT,
+} from './symbols'
```
rperez89: Should we rename this `symbols` file to something more like `status` or `activityStatus`?

src/components/Activity/TimeTag.js
```diff
@@ -0,0 +1,44 @@
...
+    .from(now)
+    .replace(/minutes?/, 'min')
+    .replace(/seconds?/, 'sec')
+    .trim()
+}
+
+function TimeTag({ date, label, ...props }) {
+  const theme = useTheme()
+  const now = useNow()
+  const targetDate = new Date(date)
```
rperez89: should we use dayjs here? asking because sometimes we use javascript Date and some others Dayjs

src/components/Activity/ActivityItem.js
bpierre: Yes good idea! Done in https://github.com/aragon/court-dashboard/pull/100/commits/22e5f514d75b0acf46523d3f2d03ee2bf12e3cff

src/components/Activity/TimeTag.js
bpierre: Yes! I also consolidated all the date related things into date-utils: https://github.com/aragon/court-dashboard/pull/100/commits/a54e3f523631e768c866da787b8ccc42230b3143

src/components/Activity/ActivityButton.js
```diff
@@ -0,0 +1,132 @@
...
+        }
+        return !opened
+      }),
+    [markActivitiesRead]
+  )
+  const handleClose = useCallback(() => {
+    markActivitiesRead()
+    setOpened(false)
+  }, [markActivitiesRead])
+
```
fabriziovigevani: What do you think if implementing this same logic ? 
 https://github.com/aragon/court-dashboard/blob/d8e5e46e8a8983eb58ad056455d53a195a3926b4/src/components/Account/Account.js#L138-L144

But with `activities` as dependency, so when clearing activities we don't lose focus.

src/components/Activity/ActivityItem.js
```diff
@@ -0,0 +1,242 @@
...
+  const activityData = getActivityData(
+    activity.activityType,
+    activity.activityParams
+  )
+
+  const handleOpen = useCallback(() => {
+    if (activity.transactionHash) {
+      window.open(
+        blockExplorerUrl('transaction', activity.transactionHash, {
+          networkType: getNetworkType(),
```
fabriziovigevani: In this case we should use `useNetwork` hook that does the `mainnet` to `main` convertion.

src/components/Activity/ActivityButton.js
bpierre: Yes good idea! Done in https://github.com/aragon/court-dashboard/pull/100/commits/a32c00c905034dfb7661d67f23bf8f2207ea70ec

src/components/Activity/ActivityItem.js
bpierre: This made me notice that we were a bit inconsistent with the network ID:

- `getNetworkId()` was returning `mainnet` rather than `main` which is more standard.
- `useNetwork()` was only doing this conversion, and doesn’t really need to be a hook.
- In some cases, we were using `getNetworkName()` to get an idea rather than `getNetworkId()` or `useNetwork()`.

So I removed `useNetwork()` and made everything else use `main`, see https://github.com/aragon/court-dashboard/pull/100/commits/2e46f755e6b2550825b7c7cab1c1ae51dbf89098

Let me know what you think!

fabriziovigevani: LGTM!

PR #104, opened 2020-02-10
Add mainnet deployment configuration

sohkai:
This adds a Github Action to "manually" deploy a mainnet build via Now on merges to `development`.

We will continue using the Now Github integration to deploy Rinkeby builds from PRs and merges to `development`.

DNS configuration being set up now.

### DIFF
--- a/.github/workflows/mainnet.yml
+++ b/.github/workflows/mainnet.yml
@@ -13,7 +13,7 @@ jobs:
     - name: Install node
       uses: actions/setup-node@v1
       with:
-        node-version: 10
+        node-version: 12
     - name: npm install
       run: npm install
     - name: npm lint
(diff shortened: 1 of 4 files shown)


.github/workflows/mainnet.yml
```diff
@@ -0,0 +1,24 @@
...
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: Install node
+      uses: actions/setup-node@v1
+      with:
+        node-version: 10
```
bpierre: Should we use 12? In case `now` deprecates 10 soon

sohkai: Sounds good, let's try 12

PR #109, opened 2020-02-10
Add logo to main header

fabriziovigevani:


src/components/Header.js
```diff
@@ -25,6 +27,13 @@ const Header = React.memo(function Header() {
           align-items: center;
         `}
       >
+        <img
+          src={headerLogoSvg}
```
bpierre:
```suggestion
          alt=""
          src={headerLogoSvg}
```

src/components/Header.js
```diff
@@ -25,6 +27,13 @@ const Header = React.memo(function Header() {
...
+          height={6 * GU}
```
bpierre: Should we set the `width` instead? That way, the title next to it wouldn’t jump after the image loads.

PR #115, opened 2020-02-10
Remove hardcoded gas limits

sohkai:
Web3 providers should be able to handle gas estimations on their own. Were we running into any problems with not setting a hardcoded amount?


fabriziovigevani:
Yes, if we didn't set a `gasLimit`, the provider never prompts you to sign and  we would get 
```
data:
message: "execution error: revert"
code: -32000
```

in the console without specific reason.

sohkai:
This is definitely worrying; will look into it. Maybe it's something ethers related but I can't imagine it.

For example, we don't use hardcoded gas limits in the client or CLI except in exceptional cases.

PR #116, opened 2020-02-10
Rinkeby onboarding

rperez89:


src/components/Onboarding/content.js
```diff
@@ -39,9 +42,9 @@ highlights.rinkeby = [
       small: null,
       large: (
         <span>
-          This is a preview release of the official Court Dashboard that will
-          launch on Feb 10th. We'd love for you to explore it and tell us what
-          you think!
+          This is a testnet release of the official Court Dashboard at
+          <b /> <Link href={MAIN_COURT_URL}>court.aragon.org</Link>. We'd love
```
sohkai: Should this be a `<br />`? Do we use `<br />`s in other places for the content?

rperez89: Just changed that `<br />` for a `{' '}`, not we are not using it in another place :)

PR #123, opened 2020-02-10
Decrease gas limit for ANJ actions

fabriziovigevani:
Decreasing gas limit for now, we should figure out how to estimate it or let the provider do it automatically

### DIFF
--- a/src/hooks/useCourtContracts.js
+++ b/src/hooks/useCourtContracts.js
@@ -22,7 +22,7 @@ import { bigNum } from '../lib/math-utils'
 import { retryMax } from '../utils/retry-max'
 
 const ACTIVATE_SELECTOR = getFunctionSignature('activate(uint256)')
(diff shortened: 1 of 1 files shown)


src/hooks/useCourtContracts.js
```diff
@@ -22,7 +22,7 @@ import { bigNum } from '../lib/math-utils'
 import { retryMax } from '../utils/retry-max'
 
 const ACTIVATE_SELECTOR = getFunctionSignature('activate(uint256)')
-const GAS_LIMIT = 1200000 // Should be relative to every tx ?
+const GAS_LIMIT = 325000 // Should be relative to every tx ?
```
sohkai: Let's just apply this to the activate and deactivate functions for now... the other ones let's continue testing / seeing if we can make #115 work.

fabriziovigevani: Yes good point

PR #124, opened 2020-02-10
ANJ Form: Replace input when 0

### LINKED ISSUE #120: Activating / deactivating ANJ input starts at 0 and is not replaced
Mildly annoying; the form starts at `0` but does not get replaced if you start typing.

fabriziovigevani:
closes #120 

Also made the amount formatted with commas when input not focused just as it's done in the microsite.

### DIFF
--- a/src/components/Dashboard/panels/ANJForm.js
+++ b/src/components/Dashboard/panels/ANJForm.js
@@ -25,9 +25,23 @@ const ANJForm = React.memo(function ANJForm({
     valueBN: bigNum(0),
     error: null,
   })
-  const { anjToken } = useCourtConfig()
   const theme = useTheme()
+  const { anjToken } = useCourtConfig()
   const inputRef = useSidePanelFocusOnReady()
+
+  const handleEditMode = useCallback(
+    editMode => {
+      setAmount(amount => ({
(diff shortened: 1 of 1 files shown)


src/components/Dashboard/panels/ANJForm.js
```diff
@@ -25,10 +25,24 @@ const ANJForm = React.memo(function ANJForm({
...
-  const { anjToken } = useCourtConfig()
   const theme = useTheme()
+  const { anjToken } = useCourtConfig()
   const inputRef = useSidePanelFocusOnReady()
 
+  const handleEditMode = useCallback(
+    editMode => {
+      setAmount(amount => ({
+        ...amount,
+        value: formatUnits(amount.valueBN, {
```
sohkai: @fabriziovigevani Something that is not nice about this is that decimals get truncated after the blur.

For example, if I wanted to _exactly_ withdraw `1.23456789`, `formatUnits()` changes this to `1.23` when I blur the input.

fabriziovigevani: Good catch! Fixing it in https://github.com/aragon/court-dashboard/pull/131

PR #126, opened 2020-02-11
Account module fixes

bpierre:
- Explicit “wrong network” error.
- Add a link to “What is an Ethereum provider”.

### DIFF
--- a/src/components/Account/ScreenError.js
+++ b/src/components/Account/ScreenError.js
@@ -28,7 +28,6 @@ function AccountModuleErrorScreen({ error, onBack }) {
   return (
     <section
       ref={elementRef}
-      tabIndex="0"
       css={`
         display: flex;
         flex-direction: column;
--- a/src/components/Account/ScreenError.js
+++ b/src/components/Account/ScreenError.js
@@ -14,7 +14,7 @@ function AccountModuleErrorScreen({ error, onBack }) {
     if (error instanceof UnsupportedChainError) {
       return [
         'Wrong network',
-        `Please select to the ${getNetworkName(
+        `Please select the ${getNetworkName(
           env('CHAIN_ID')
         )} network in your wallet and try again.`,
       ]
--- a/src/components/Account/ScreenError.js
+++ b/src/components/Account/ScreenError.js
@@ -1,16 +1,29 @@
-import React, { useRef } from 'react'
+import React, { useMemo, useRef } from 'react'
 import PropTypes from 'prop-types'
 import { GU, Link, textStyle, useTheme } from '@aragon/ui'
-
+import { UnsupportedChainError } from 'use-wallet'
+import env from '../../environment'
+import { getNetworkName } from '../../lib/web3-utils'
 import connectionError from './assets/connection-error.png'
 
-function AccountModuleErrorScreen({
-  onBack,
-  secondary = 'You can try another Ethereum provider.',
-  title = 'Failed to enable your account',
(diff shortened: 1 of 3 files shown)


src/components/Account/ScreenError.js
```diff
@@ -1,21 +1,33 @@
...
-}) {
+function AccountModuleErrorScreen({ error, onBack }) {
   const theme = useTheme()
   const elementRef = useRef()
 
+  const [title, secondary] = useMemo(() => {
+    if (error instanceof UnsupportedChainError) {
+      return [
+        'Wrong network',
+        `Please select to the ${getNetworkName(
```
sohkai:
```suggestion
        `Please select the ${getNetworkName(
```

src/components/Account/ScreenError.js
```diff
@@ -1,21 +1,33 @@
...
+          env('CHAIN_ID')
+        )} network in your wallet and try again.`,
+      ]
+    }
+    return [
+      'Failed to enable your account',
+      'You can try another Ethereum wallet.',
```
sohkai: Should we be using `provider`, especially since the link is "What is an Ethereum provider?"

bpierre: Yes probably for now… I was noticing that ethereum.org dropped the term “provider” for end-user apps, to only talk about “wallet” instead, which I think is what people use anyway.

For example I couldn’t find an official page for “what is an Ethereum provider”, and I linked to https://ethereum.org/wallets/ instead… which doesn’t contain the word “provider”. I’m starting to think this is all confusing for users, and we might want to consider using the terms “wallet” and “connect your wallet” rather than “provider” and “enable your provider” in the UI. And we could keep the term provider to refer to the EIP 1193 provider.

@dizzypaty @sohkai thoughts?

sohkai: In the context of the Court especially, I think "wallet" makes more sense than "provider".

Perhaps we can keep the "account" nomenclature for the actually account, but write messages in a way that makes it more clear your account is derived from your Ethereum wallet?

E.g. "connect account" is still what I would use for both the Court and the client, but the modal could say something like "Use account from" rather than "Ethereum provider", and change the bottom "What is an Ethereum provider" to "Don't have an Ethereum account?".

bpierre: > Perhaps we can keep the "account" nomenclature for the actual account, but write messages in a way that makes it more clear your account is derived from your Ethereum wallet?

Yes that sounds good!

Changes we could do:

- Replace “Enable account” by “Connect account”.
- Replace “Ethereum providers” by “Use account from”.
- Replace “What is an Ethereum provider” by “Don’t have an Ethereum account?”

What remains to be defined is the title of the popover in the error and connecting screen of the popover. I think we could keep “Use account from” for the connecting state, since we want to focus the user attention on the content which appears briefly already. And we could use “Connection error” or “Error” for the error screen?

Also, should we replace “Active wallet” by “Active account” on the connected state, for consistency?

dizzypaty: Totally agree! Provider feels more technical and niche than account and wallet. I agree with all the proposed changes!

sohkai: > Also, should we replace “Active wallet” by “Active account” on the connected state, for consistency?

I think so; I'm trying to push "account" as it's more similar to a traditional user id than a "wallet" (which holds many "accounts").

PR #132, opened 2020-02-11
Change court address for Rinkeby

fabriziovigevani:


src/networks.js
```diff
@@ -1,6 +1,6 @@
 export const networks = {
   rpc: { court: '0xC89Ce4735882C9F0f0FE26686c53074E09B0D550' },
   ropsten: { court: '0x3b26bc496aebaed5b3E0E81cDE6B582CDe71396e' },
-  rinkeby: { court: '0x014e190d7ae3ed129baca6b72568c4b8555758c8' },
+  rinkeby: { court: '0xb5ffbe75fa785725eea5f931b64fc04e516c9c5d' },
```
sohkai: For the usability testing phase, we should use [`0x014E190D7AE3ED129bACa6b72568C4B8555758c8`](https://github.com/aragon/aragon-network-deploy/pull/44/files#diff-44b335bca7ce47e9b6d3cd45b5e77ebfR3). It should also use [`wss://api.thegraph.com/subgraphs/name/aragon/aragon-court-usability`](https://thegraph.com/explorer/subgraph/aragon/aragon-court-usability) as the endpoint.

Not super sure on how easy it is for us to add an environment like this, but it should be promoted to be our default Rinkeby environment for now.

rperez89: why are we using a different endpoint for the usability?

fabriziovigevani: Just added a new env variable that if present will override the subgraph name. Tell me what you think  https://github.com/aragon/court-dashboard/pull/133

I think that once we are done with it, we can change the `Rinkeby` address and remove the env on now.

sohkai: #133 looks good!

sohkai:
Merging, this is the correct address we have in [the deployments](https://github.com/aragon/aragon-network-deploy/blob/master/data/output/court.rinkeby.json).

PR #137, opened 2020-02-12
Task generation fix

rperez89:
This PR is to fix an issue found about the task, so if the user commits / reveal / appeal or confirm appeal the task should not being displayed  anymore as open since that could be confusing for the user. Given that we don't have the concept of "Past" tasks we are not displaying it anymore because it was completed

The way that was working until now was that the task remained as open until the phase was ended


src/hooks/useFilteredTasks.js
```diff
@@ -56,7 +56,8 @@ function useFilteredTasks(jurorTasksSelected, connectedAccount) {
   const filteredTasks = useMemo(
     () =>
       tasksToFilter.filter(
-        ({ phaseType, dueDate }) =>
+        ({ phaseType, dueDate, open }) =>
+          open === true &&
```
fabriziovigevani: This could be just `open && ...`

src/hooks/useTasks.js
```diff
@@ -89,3 +92,33 @@ function getTaskName(phase) {
     return 'Confirm appeal'
   }
 }
+
+function getRoundStatus(round, currentPhase) {
+  if (currentPhase === DisputesTypes.Phase.AppealRuling) {
+    if (round.appeal) {
+      return false
+    }
+    return true
```
fabriziovigevani: Here we could just `return !round.appeal`

src/hooks/useTasks.js
```diff
@@ -89,3 +92,33 @@ function getTaskName(phase) {
...
+  }
+  if (currentPhase === DisputesTypes.Phase.ConfirmAppeal) {
+    if (round?.appeal?.opposedRuling !== 0) {
+      return false
+    }
+    return true
```
fabriziovigevani: Same here  `return !round.appeal.opposedRuling`

src/hooks/useTasks.js
```diff
@@ -89,3 +92,33 @@ function getTaskName(phase) {
...
+  }
+}
+
+function getDraftStatus(draft, currentPhase) {
+  if (currentPhase === DisputesTypes.Phase.VotingPeriod) {
+    if (draft.commitment) {
+      return false
+    }
+    return true
```
fabriziovigevani: Same logic here

src/hooks/useTasks.js
```diff
@@ -89,3 +92,33 @@ function getTaskName(phase) {
...
+  }
+  if (currentPhase === DisputesTypes.Phase.RevealVote) {
+    if (draft.outcome || !draft.commitment) {
+      return false
+    }
+    return true
```
fabriziovigevani: And here `return draft.commitment && !draft.outcome`

src/hooks/useTasks.js
```diff
@@ -89,3 +92,33 @@ function getTaskName(phase) {
     return 'Confirm appeal'
   }
 }
+
+function getRoundStatus(round, currentPhase) {
```
fabriziovigevani: What do you think of renaming this to something like isAppealTaskOpen ? Since we are not actually returning the status of the round

src/hooks/useTasks.js
```diff
@@ -89,3 +92,33 @@ function getTaskName(phase) {
...
+  }
+  if (currentPhase === DisputesTypes.Phase.ConfirmAppeal) {
+    if (round?.appeal?.opposedRuling !== 0) {
+      return false
+    }
+    return true
+  }
+}
+
+function getDraftStatus(draft, currentPhase) {
```
fabriziovigevani: And also here, it could be renamed to something like `isVotingTaskOpen`

PR #144, opened 2020-02-13
Add final ruling on dispute card

fabriziovigevani:


src/utils/crvoting-utils.js
```diff
@@ -42,28 +44,40 @@ const outcomeStringMapping = {
...
+const finalRulingStringMapping = {
+  [OUTCOMES.Leaked]: 'Invalid ruling',
+  [OUTCOMES.Refused]: 'Refused',
+  [OUTCOMES.Against]: 'Ruling against',
+  [OUTCOMES.InFavor]: 'Ruling in favor',
+}
+
+export function finalRulingToString(outcome) {
+  if (!outcome) {
+    return finalRulingStringMapping[OUTCOMES.refused]
```
sohkai: Interesting; can a final ruling be refused?

fabriziovigevani: Yes, in the case that the majority refuses to vote, or nobody actually votes.

PR #149, opened 2020-02-13
Rollback dispute info outcome

fabriziovigevani:
Changes on `Dipsute Info` from this PR https://github.com/aragon/court-dashboard/pull/144 where overriden by another PR. 
Rolling back the changes


sohkai:
@fabriziovigevani I think you might've included another file here? (`src/asests/Pull Request...`)

fabriziovigevani:
Yes, just removed it. Not sure how those files where generated in the first place

PR #156, opened 2020-02-14
Evidence: duplicates fix

### LINKED ISSUE #185: Duplicate evidence showing up sometimes
There is a bug that shows duplicate evidence for some users some of the time:

Screenshot_2020-02-18_at_19 34 58

rperez89:
This PR includes:

- Fixes from comments on [this PR](https://github.com/aragon/court-dashboard/pull/87)
- Avoid duplicate evidences
- Wrong `In favor` , `Against` evidence type

closes #185.


src/components/Disputes/DisputeEvidences.js
```diff
@@ -133,7 +133,7 @@ const EvidenceContent = React.memo(function EvidenceContent({
               ${textStyle('body2')};
             `}
           >
-            {dayjs(createdAt).format('DD/MM/YY')}
+            {dayjs(toDate(createdAt)).format('DD/MM/YY')}
```
sohkai: Not particularly relevant, but I was looking at the `toDate()` implementation and was quite surprised to see it doing string operations on the returned ISO string.

Maybe `toDate()` can just return `new Date(unixSeconds * 1000)`?

There is also [`dayjs.unix()`](https://day.js.org/docs/en/parse/unix-timestamp)

src/components/Disputes/DisputeEvidences.js
```diff
@@ -153,10 +153,10 @@ const EvidenceContent = React.memo(function EvidenceContent({
           >
             {(() => {
               if (addressesEqual(plaintiff, submitter)) {
-                return 'In favor'
+                return 'Against'
               }
               if (addressesEqual(defendant, submitter)) {
-                return 'Against'
+                return 'In favor'
```
sohkai: Let's check with @dizzypaty on this text. We should use the same text as we do on the voting buttons.

src/endpoints.js
```diff
@@ -13,7 +13,7 @@ const GRAPH_API_BASE_HTTP = 'https://api.thegraph.com'
 const GRAPH_API_BASE_WS = 'wss://api.thegraph.com'
 
 export const IPFS_ENDPOINT = isLocalOrUnknownNetwork(CHAIN_ID)
-  ? 'http://127.0.0.1:8080/ipfs'
```
sohkai: I might still keep this, since when you're on a local chain, our IPFS node isn't going to have your content.

src/hooks/useEvidences.js
```diff
@@ -14,34 +12,31 @@ export default function useEvidences(evidences) {
...
-          ...prev,
-          evidence,
-        ].sort(({ createdAt: dateLeft }, { createdAt: dateRight }) =>
-          dayjs(dateLeft).isAfter(dayjs(dateRight))
-            ? 1
-            : dayjs(dateLeft).isSame(dayjs(dateRight))
-            ? 0
-            : -1
+      setEvidenceProcessed(prev => {
+        if (prev.some(element => element.id === evidence.id)) {
```
sohkai: I'm wondering if we shouldn't add a mapping that will act as a global cache for files fetched from IPFS into this hook instead of doing this deduping here (e.g. an example cache that only "saves" on success: [`AsyncRequestCache`](https://github.com/aragon/aragon.js/blob/master/packages/aragon-wrapper/src/utils/AsyncRequestCache.js)).

Doing this here is expensive, because we've already had to re-fetch the file, only to realize we should throw it away.

src/endpoints.js
rperez89: yeah, this file should not be uploaded i change it for testing

src/components/Disputes/DisputeEvidences.js
rperez89: @fabriziovigevani could we change that function? or was written like this for some reason?

fabriziovigevani: I don't recall implementing that function. Maybe you added it for the dispute detail PR ?

src/hooks/useEvidences.js
```diff
@@ -30,11 +27,19 @@ export default function useEvidences(evidences) {
...
+          return
+        }
+
         const { data, error } = await ipfsGet(ipfsCid)
+
         if (error) {
           return appendEvidence({ error: ERROR_TYPES.ERROR_FETCHING_IPFS })
         }
         if (data) {
+          setEvidenceCache(evidenceCache.set(evidenceCache, true))
```
sohkai: I don't think you need to use the `setEvidenceCache()` here; the `Map` doesn't need to be re-set.

src/hooks/useEvidences.js
```diff
@@ -30,11 +27,19 @@ export default function useEvidences(evidences) {
 
       // If is a valid CID try to fetch the content
       if (ipfsCid) {
+        // If the cid was already processed exit
+        if (evidenceCache.get(ipfsCid)) {
```
sohkai: So asynchronous operations are always a bit tricky, but moving the logic to be here is less safe than before.

What can happen now:

1. In one render, start fetching `cidA`
2. Quickly afterwards, we re-render, and we again start fetching `cidA`
  - Note here that if we're still waiting to hear back from `await ipfsGet('cidA')`, `evidenceCache` won't have been set yet

This can't happen right now, but is really easy to become a possibility if any action after `evidenceCache.set()` becomes async:

1. In one render, start fetching `cidA`
2. Continue after `await ipfsGet('cidA')`
3. Cancel the render, because we re-render (this requires an async action to happen after `evidenceCache.set()`)
4. Now on the next renders, because we've already set the evidence cache, we skip this value and it will never appear.

sohkai: For the sake of sanity and being safe, let's:

- Remove the caching idea; it adds too much complexity to do correctly, and is not valuable enough here
- Make sure `id` is in each object of `evidenceProcessed` (I believe we need to add this in just the error cases)
- Before we do `evidences.forEach()`, let's filter the `evidences` list by what we already have in `evidenceProcessed`, and update `evidenceProcessed` to filter out any items that aren't in `evidences`
  - In set terminology:  `Set(evidences) - Set(evidenceProcessed)`, and `Set(evidences) U Set(evidenceProcessed)`
  - Then do the `forEach()` over only the new evidences (`Set(evidences) - Set(evidenceProcessed)`)

In this scenario, if we've protected ourselves correctly with `cancelled` guarding any state setters (i.e. `setEvidenceProcessed()`), we should never have a case where a re-render would cause inconsistent state. We only "commit" state when we're in the same render, and we only fetch new evidences when they're not in the "committed" state

src/components/Disputes/DisputeEvidences.js
dizzypaty: @rperez89 , @sohkai  – I'd say let's remove this field `Argument` from the evidence altogether. At least for now until we have a more clear solution. It's creating more confusion than helping actually.

src/hooks/useEvidences.js
bpierre: What about:

- Having a single evidence fetching function, that only caches valid evidences (non errored).
- Move the fetching + transformation + caching + error state in this function.
- Only fetch all the evidences + update the result in the state in `useEffect()`.
- Update all the evidences at the same time rather than appending.

/!\ Untested code /!\

```jsx
function useEvidences(rawEvidences) {
  // Contains valid evidences + errored evidences
  const [evidences, setEvidences] = useState([])

  // Contains valid evidences only
  const evidencesCache = useRef(new Map())

  // Fetch an evidence data from IPFS if needed, prepares the evidence object,
  // and cache it if valid. If invalid, returns an errored evidence object.
  const fetchEvidence = useCallback(async rawEvidence => {
    const { id, data: uriOrData, submitter, createdAt } = rawEvidence

    if (ipfsCache.current.has(id)) {
      return ipfsCache.current.get(id)
    }

    const baseEvidence = {
      id,
      metadata: null,
      defendant: '',
      agreementText: '',
      submitter,
      createdAt,
      error: false,
    }

    // We might want to rename this function `getIpfsCidFromUri()` or `parseIpfsUri()`
    const cid = getIpfsCidFromString(uriOrData)

    // Not an IPFS URI
    if (!cid) {
      ipfsCache.current.set(id, { ...baseEvidence, metadata: uriOrData })
      return ipfsCache.current.get(id)
    }

    const { data, error } = await ipfsGet(cid)

    if (error) {
      return { ...baseEvidence, error: ERROR_TYPES.ERROR_FETCHING_IPFS }
    }

    // we might re-set it here, but it’s not really a problem
    ipfsCache.current.set(id, data)
    return {
      ...baseEvidence,
      metadata: data,
    }
  }, [])

  useEffect(() => {
    let cancelled = false

    const updateEvidences = async () => {
      const evidences = await Promise.all(rawEvidences.map(fetchEvidence))
      if (!cancelled) {
        setEvidences(evidences.sort((a, b) => a.createdAt - b.createdAt))
      }
    }

    updateEvidences()

    return () => {
      cancelled = true
    }
  }, [rawEvidences, fetchEvidence])
}
```

src/lib/ipfs-utils.js
```diff
@@ -14,3 +15,12 @@ export const ipfsGet = async cid => {
     return { error: true }
   }
 }
+
+export const getIpfsCidFromString = string => {
```
bpierre: We might want to rename this `getIpfsCidFromUri(uri)` or `parseIpfsUri(uri)`

src/hooks/useEvidences.js
rperez89: With that promise all, how are we going to handle the timeout for the request? it can't be too much time right? because we can be loading disputes for a long time with the user waiting to load all the disputes

src/hooks/useEvidences.js
```diff
@@ -42,33 +42,49 @@ export default function useEvidences(rawEvidences) {
...
-      const evidences = await Promise.all(rawEvidences.map(fetchEvidence))
-      if (!cancelled) {
-        setEvidences(
-          evidences.sort(
-            (evidenceA, evidenceB) => evidenceA.createdAt - evidenceB.createdAt
-          )
-        )
-      }
+      await Promise.all(
+        rawEvidences.map(async rawEvidence => {
```
bpierre: I guess we could remove the `Promise.all()` and use `forEach()` since we are not doing anything after they complete − or should we use it to export a `loading` boolean?

rperez89: maybe we can remove it since we are fetching one by one right?

bpierre: Yes sounds good!

src/hooks/useEvidences.js
sohkai: > With that promise all, how are we going to handle the timeout for the request?

This should be handled by `ipfsGet()`; if it times out, we return an error.

src/hooks/useEvidences.js
sohkai: > maybe the sync indicator is also needed to let the user know that we are still fetching right?

Yes, this would be nice!

rperez89: Screen Shot 2020-03-03 at 11 12 32 PM
@dizzypaty wondering which ` theme.floating` should we use here for the sync indicator

src/hooks/useEvidences.js
```diff
@@ -77,6 +78,7 @@ export default function useEvidences(rawEvidences) {
           })
         })
       )
+      setFetchingEvidences(false)
```
sohkai: I think we'll need to reset this to be `true` if it isn't already at the start of the `useEffect()`.

If a component using this hook re-renders with more evidence, and this was previously already set to false, it won't appear as syncing.

src/hooks/useEvidences.js
dizzypaty: Hey @rperez89 it seems to me that's the right color to use for the sync indicator. It's not a dark theme color, it's the default one because it needs to catch users attention as it's a feedback indicator, as it's also not persistent.

PR #157, opened 2020-02-14
Tasks: Fix task count

### LINKED ISSUE #153: Tasks: "all tasks" displays more actions than in list
I believe the `Count` should have decreased to just 2?

Screen Shot 2020-02-14 at 12 49 53 AM

rperez89:
This closes #153

### DIFF
--- a/src/hooks/useTasks.js
+++ b/src/hooks/useTasks.js
@@ -32,7 +32,7 @@ function useOpenTasks(tasks, now, courtSettings) {
     : null
 
   return useMemo(() => {
-    if (!tasks) {
+    if (!convertedTasks) {
       return []
     }
 
@@ -83,7 +83,8 @@ function useOpenTasks(tasks, now, courtSettings) {
       }
     }
     return openTasks
-  }, [convertedTasksPhasesKey, tasks]) // eslint-disable-line react-hooks/exhaustive-deps
+    // Since we are using our own generated cache key we don't need to add the convertedTasks to the dependency list.
+  }, [convertedTasksPhasesKey]) // eslint-disable-line react-hooks/exhaustive-deps
 }
 
 function getTaskName(phase) {
@@ -103,25 +104,20 @@ function getTaskName(phase) {
 
 function isAppealTaskOpen(round, currentPhase) {
   if (currentPhase === DisputesTypes.Phase.AppealRuling) {
-    if (round.appeal) {
-      return false
-    }
-    return true
+    return !round.appeal
   }
   if (currentPhase === DisputesTypes.Phase.ConfirmAppeal) {
-    if (Number(round?.appeal?.opposedRuling) !== 0) {
-      return false
+    if (round?.appeal?.opposedRuling) {
+      return Number(round.appeal.opposedRuling) === 0
     }
+
     return true
   }
 }
 
 function isVotingTaskOpen(draft, currentPhase) {
   if (currentPhase === DisputesTypes.Phase.VotingPeriod) {
-    if (draft.commitment) {
-      return false
-    }
-    return true
+    return !draft.commitment
   }
   if (currentPhase === DisputesTypes.Phase.RevealVote) {
     if (draft.outcome || !draft.commitment) {
--- a/src/hooks/useTasks.js
+++ b/src/hooks/useTasks.js
@@ -31,52 +31,57 @@ function useOpenTasks(tasks, now, courtSettings) {
     : null
 
   return useMemo(() => {
+    if (!tasks) {
+      return []
+    }
+
+    const incompleteTasks = tasks.filter(
+      (_, i) => currentRoundPhases[i].phase !== DisputesTypes.Phase.Ended
+    )
+    const incompleteRoundPhases = currentRoundPhases.filter(roundPhase => {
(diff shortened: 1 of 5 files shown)


src/hooks/useTasks.js
```diff
@@ -39,36 +39,42 @@ function useOpenTasks(tasks, now, courtSettings) {
           currentRoundPhases[i].phase === DisputesTypes.Phase.AppealRuling ||
```
sohkai: So these changes looks good, but the `useMemo()` may look simpler / more clear rewritten as:

```js
useMemo(() => {
  if (!tasks) {
    return []
  }

  const incompleteTasks = tasks.filter((_, i) => currentRoundPhases[i].phase !== DisputesTypes.Phase.Ended)
  const openTasks = []
  for (let i = 0; i < incompleteTasks.length; i++) {
    const currentPhase = currentRoundPhases[i].phase

    if (
      currentPhase !== DisputesTypes.Phase.AppealRuling &&
      currentPhase !== DisputesTypes.Phase.ConfirmAppeal
    ) {
      for (let j = 0; j < tasks[i].jurors.length; j++) {
        if (
          isVotingTaskOpen(
            tasks[i].jurors[j],
            currentPhase
          )
        ) {
          openTasks.push({
            number: tasks[i].number,
            state: tasks[i].state,
            createdAt: parseInt(tasks[i].createdAt, 10) * 1000,
            juror: tasks[i].jurors[j].juror.id,
            disputeId: tasks[i].dispute.id,
            commitment: tasks[i].jurors[j].commitment,
            outcome: tasks[i].jurors[j].outcome,
            phase: getTaskName(currentRoundPhases[i].phase),
            phaseType: currentRoundPhases[i].phase,
            dueDate: currentRoundPhases[i].nextTransition,
            open: true,
          })
        }
      }
    } else if (isAppealTaskOpen(tasks[i], currentPhase)) {
      // We are in appeal or confirm and only need to generate a single task
      // (rather than one per juror) if the task is still open
      openTasks.push({
        number: tasks[i].number,
        state: tasks[i].state,
        createdAt: parseInt(tasks[i].createdAt, 10) * 1000,
        juror: 'Anyone',
        disputeId: tasks[i].dispute.id,
        phase: getTaskName(currentRoundPhases[i].phase),
        dueDate: currentRoundPhases[i].nextTransition,
        phaseType: currentRoundPhases[i].phase,
        open: true,
      })
    }
  }

  return openTasks
}
```

Notice how the indenting has been significantly reduced, and we could "lift" the nested ifs out.

(And to double check, on the `currentPhase !== DisputesTypes.Phase.Ended`; this _is_ something we can just filter out at the beginning? If a task's phase is ended, it should never be considered, right?)

src/hooks/useTasks.js
```diff
@@ -39,36 +39,42 @@ function useOpenTasks(tasks, now, courtSettings) {
...
+            openTasks.push({
+              number: tasks[i].number,
+              state: tasks[i].state,
+              createdAt: parseInt(tasks[i].createdAt, 10) * 1000,
+              juror: 'Anyone',
+              disputeId: tasks[i].dispute.id,
+              phase: getTaskName(currentRoundPhases[i].phase),
+              dueDate: currentRoundPhases[i].nextTransition,
+              phaseType: currentRoundPhases[i].phase,
+              open: true,
```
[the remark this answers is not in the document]
rperez89: ohh that is something old from the first implementation where we have the concept of "past tasks" will do a quickly refactor to this, good catch

src/hooks/useTasks.js
```diff
@@ -15,67 +15,75 @@ export default function useTasks() {
...
-  const currentRoundPhasesKey = currentRoundPhases
-    ? currentRoundPhases
+  const convertedTasksPhasesKey = convertedTasks
+    ? convertedTasks
         .map(phase => DisputesTypes.convertToString(phase.phase))
         .join('')
     : null
 
   return useMemo(() => {
+    if (!tasks) {
```
sohkai:
```suggestion
    if (!convertedTasks) {
```

We don't rely on `tasks` anymore in this second `useMemo()`, so we should only operate with data we're actually going to use (rather than a transient dependency).

src/hooks/useTasks.js
```diff
@@ -15,67 +15,75 @@ export default function useTasks() {
...
-                ),
-              })
-            }
-          }
         }
       }
     }
     return openTasks
-  }, [currentRoundPhasesKey, tasks]) // eslint-disable-line react-hooks/exhaustive-deps
+  }, [convertedTasksPhasesKey, tasks]) // eslint-disable-line react-hooks/exhaustive-deps
```
sohkai: Let's add a comment explaining that we're using our own generated cache key and don't need to add `convertedTasks` to the dependency list.

src/hooks/useTasks.js
```diff
@@ -101,7 +109,7 @@ function isAppealTaskOpen(round, currentPhase) {
     return true
   }
   if (currentPhase === DisputesTypes.Phase.ConfirmAppeal) {
-    if (round?.appeal?.opposedRuling !== 0) {
+    if (Number(round?.appeal?.opposedRuling) !== 0) {
```
sohkai: Not quite sure this is what you want; `Number(undefined)` is `NaN`.

PR #158, opened 2020-02-14
Voting: save secret hash to local storage

### LINKED ISSUE #146: Voting: save secret hash to local storage
If this value is available for the juror, we should autofill the secret when they reveal.

rperez89:
This PR closes #146

### DIFF
--- a/src/components/Disputes/panels/CommitPanel.js
+++ b/src/components/Disputes/panels/CommitPanel.js
@@ -15,6 +15,7 @@ import {
 } from '@aragon/ui'
 import useOneTimeCode from '../../../hooks/useOneTimeCode'
 import { useWallet } from '../../../providers/Wallet'
+import { saveCodeInLocalStorage } from '../../../utils/one-time-code-utils'
 
 import IconOneTimeCode from '../../../assets/IconOneTimeCode.svg'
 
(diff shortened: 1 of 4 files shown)


src/components/Disputes/panels/CommitPanel.js
```diff
@@ -54,12 +56,20 @@ const CommitPanel = React.memo(function CommitPanel({
   const handleDownloadCode = useCallback(() => {
     download()
     setCodeSaved(true)
-  }, [download])
+    localStorage.setItem(
+      `oneTimeCode-${connectedAccount}-${dispute.id}`,
```
sohkai: Perhaps we should move this into a utility :).

src/components/Disputes/panels/CommitPanel.js
```diff
@@ -54,12 +56,20 @@ const CommitPanel = React.memo(function CommitPanel({
...
+      oneTimeCode
+    )
+  }, [connectedAccount, dispute.id, download, oneTimeCode])
 
   const handleCopyCode = useCallback(() => {
     setCodeCopied(true)
     toast('One-time-use code copied')
-  }, [toast])
+    localStorage.setItem(
+      `oneTimeCode-${connectedAccount}-${dispute.id}`,
```
sohkai: Since we do this twice, perhaps it would make sense to move to the handler generating the tx?

src/utils/one-time-code-utils.js
```diff
@@ -0,0 +1,18 @@
+export function saveCodeInLocalStorage(
+  connectedAccount,
+  disputeId,
+  oneTimeCode
+) {
+  localStorage.setItem(
+    `oneTimeCode-${connectedAccount}-${disputeId}`,
```
sohkai: In the client we use `:` as delimiters for some localstorage keys (no real reason, but I guess it's more rare for a variable to include `:` than a `-`).

rperez89: I can change if you wants prefer that! sure

PR #165, opened 2020-02-14
loading ring fix

rperez89:
This pr is a fix for the loading ring of the banner that was not showing correctly


src/components/Dashboard/AccountBanner.js
```diff
@@ -83,6 +83,7 @@ function AccountBanner({ status, loading, minActiveBalance, activeBalance }) {
     theme
   )
 
+  if (loading) return <BannerLoadingRing />
```
sohkai: Let's prefer braces :)

PR #166, opened 2020-02-14
Dashboard: Handle fetching errors

fabriziovigevani:


src/components/Dashboard/DashboardStateProvider.js
```diff
@@ -54,7 +54,8 @@ function WithSubscription({ Provider, connectedAccount, children }) {
   } = useJurorDraftsNotRewardedSubscription(account)
 
   const fetching = balancesFetching || appealsFetching || jurorDraftsFetching
-  const errors = [...balanceErrors, ...appealErrors, jurorDraftsError]
+  const errors = [...balanceErrors, ...appealErrors]
+  jurorDraftsError && errors.push(jurorDraftsError)
```
sohkai: This is kind of weird to understand at a glance; might as well just make it an if or move it above and do an `jurorDraftsError ? [jurorDraftsError] : []`

fabriziovigevani:
Yeah since we have multiple subscriptions, in the case one of the queries is off or something it could happen that we get different errors

PR #171, opened 2020-02-15
Set round pill height 20px

fabriziovigevani:
Closes notion item https://www.notion.so/9e7e5cd565b44800a79b5640a5f094a8?v=ef5587b7659946cf806ca4414bcef76a&p=aee66ba6ee584275a381aa4a0bef13e8

The problem appeared only for Firefox browser

### DIFF
--- a/src/components/Disputes/DisputeRoundPill.js
+++ b/src/components/Disputes/DisputeRoundPill.js
@@ -10,22 +10,29 @@ function DisputeRoundPill({ roundId }) {
 
   const label = `Round ${numberToWord(roundId)}`
   return (
-    <span
+    <div
       css={`
-        padding: 1px ${2 * GU}px;
-        border-radius: 100px;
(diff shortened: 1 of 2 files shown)


src/components/Disputes/DisputeRoundPill.js
```diff
@@ -10,22 +10,29 @@ function DisputeRoundPill({ roundId }) {
...
+        border-radius: 100px;
+        height: ${2.5 * GU}px;
       `}
     >
-      {label}
-    </span>
+      <span
+        css={`
+          ${textStyle('label2')};
+          color: ${theme.accent};
```
bpierre: Could we also move it 1px down? This label is always displayed in uppercase, so it will make the font look more centered:

```suggestion
          color: ${theme.accent};
          transform: translateY(1px);
```

PR #172, opened 2020-02-16
Appeal: amend info copy

### LINKED ISSUE #189: "Learn more" in appeal confirmation window does not link anywhere
See screenshot, where it says "Learn more" it looks like a link but doesn't go anywhere.

Also, the copy here is incorrect. There are two possible outcomes for the collateral if the confirming party is ruled against:

- The collateral could be redistributed to the appealing party, if the final ruling is in the appealing party's favor
- The collateral minus a fee could be returned to the confirming party, if the final ruling is in neither the appealing party nor the confirming party's favor (e.g. the appealing party proposes "In favor", the confirming party proposes "refuse to vote", and the final ruling is "against")

Screenshot_2020-02-18 Aragon Court Dashboard(1)

fabriziovigevani:
Noticed that some info sections where not entirely accurate. 

1. Execute ruling info section: Whoever executes the dispute ruling will not get rewards.

Users can earn rewards from executing actions when: 
- Drafting
- Settling penalties

2. When appealing, if the final ruling of a dispute is different from the ruling appealed for, the appealer will lose the entire deposit only if the opposed ruling is the same as  the final ruling. Otherwise, both parties will get their deposits back  `- (round fees )/ 2`.

Fixes #189.

### DIFF
--- a/src/components/Disputes/actions/DisputeAppeal.js
+++ b/src/components/Disputes/actions/DisputeAppeal.js
@@ -1,5 +1,5 @@
 import React, { useCallback } from 'react'
-import { Button, Info, GU } from '@aragon/ui'
+import { Button, GU, Info, Link } from '@aragon/ui'
 import { useWallet } from '../../../providers/Wallet'
 
 function DisputeAppeal({ onRequestAppeal, confirm }) {
@@ -24,8 +24,22 @@ function DisputeAppeal({ onRequestAppeal, confirm }) {
       >
         {actionLabel}
       </Button>
-      <Info mode="description">
-        Anyone holding the required appeal deposit can trigger this action.
+      <Info>
+        <strong>Anyone</strong> can{' '}
+        <strong>
+          lock DAI as collateral to{' '}
+          {confirm ? 'confirm an appeal' : 'propose an appeal'}{' '}
+        </strong>
+        , if they don’t agree with the current outcome. When the final ruling is
+        confirmed, the user who {confirm ? 'confirmed the appeal' : 'appealed'}{' '}
+        gets rewarded if the ruling has switched in their favor. If not, their
+        entire collateral could be re-distributed to the winning party.{' '}
+        {confirm
+          ? 'If an appeal is confirmed, a new adjudication round is initiated and a new jury is drafted.'
+          : ''}
+        <Link href="https://help.aragon.org/article/43-dispute-lifecycle#appeal">
+          Learn more
+        </Link>
       </Info>
     </div>
   )
--- a/src/components/Disputes/actions/DisputeAppeal.js
+++ b/src/components/Disputes/actions/DisputeAppeal.js
@@ -28,17 +28,19 @@ function DisputeAppeal({ onRequestAppeal, confirm }) {
         <strong>Anyone</strong> can{' '}
         <strong>
           lock DAI as collateral to{' '}
-          {confirm ? 'confirm an appeal' : 'propose an appeal'}{' '}
+          {confirm ? 'confirm an appeal' : 'initiate an appeal'}{' '}
         </strong>
-        , if they don’t agree with the current outcome. When the final ruling is
-        confirmed, the user who {confirm ? 'confirmed the appeal' : 'appealed'}{' '}
-        gets rewarded if the ruling has switched in their favor. If not, their
-        entire collateral could be re-distributed to the winning party.{' '}
+        , if they believe the{' '}
+        {confirm ? 'ruling appealed for' : 'current outcome'} is incorrect. When
+        the final ruling is confirmed, the user who{' '}
+        {confirm ? 'confirmed the appeal' : 'appealed'} gets rewarded if the
+        ruling has switched in their favor. If not, their entire collateral
+        could be re-distributed to the winning party.{' '}
         {confirm
-          ? 'If an appeal is confirmed, a new adjudication round is initiated and a new jury is drafted.'
-          : ''}
+          ? 'When an appeal is confirmed, a new adjudication round is initiated and a new jury is drafted. '
+          : 'If an appeal is confirmed, a new adjudication round is initiated and a new jury is drafted. '}
         <Link href="https://help.aragon.org/article/43-dispute-lifecycle#appeal">
-          Learn more
+          Learn more.
         </Link>
       </Info>
     </div>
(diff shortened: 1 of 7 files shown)


src/components/Disputes/panels/AppealPanel.js
```diff
@@ -194,7 +194,7 @@ const AppealPanel = React.memo(function AppealPanel({
       )}
       <Info>
         Please note that if the final ruling outcome is different from your
-        selected appeal, the entire amount of your collateral will be slashed.{' '}
+        selected appeal, the entire amount of your collateral could be slashed.{' '}
```
sohkai: This sounds like we need a "learn more" link here.

src/components/Disputes/actions/DisputeAppeal.js
```diff
@@ -26,14 +25,7 @@ function DisputeAppeal({ onRequestAppeal, confirm }) {
...
       <Info mode="description">
-        Anyone can trigger this action, and whoever does will get some{' '}
-        <span
-          css={`
-            color: ${theme.help};
-          `}
-        >
-          rewards.
-        </span>
+        Anyone holding the required appeal deposit can trigger this action.
```
sohkai:
```suggestion
        Anyone can deposit the required amount to appeal and trigger this action.
```

I notice "anyone" is also in a `<strong>` in the other section; should it also be the case here?

src/components/Disputes/panels/AppealPanel.js
fabriziovigevani: Added

src/components/Disputes/actions/DisputeAppeal.js
fabriziovigevani: Yes, good catch

src/components/Disputes/actions/DisputeDraft.js
```diff
@@ -30,8 +30,8 @@ function DisputeDraft({ disputeId, onDraft }) {
         </Button>
       </div>
       <Info>
-        The evidence submission period is closed. Anyone can now trigger the
-        drafting of jury and earn some rewards.
+        The evidence submission period is closed. <strong>Anyone</strong> can
+        now trigger the drafting of jury and earn some rewards.
```
sohkai:
```suggestion
        now trigger the drafting of a jury and earn some rewards.
```

src/components/Disputes/actions/DisputeAppeal.js
```diff
@@ -24,8 +24,22 @@ function DisputeAppeal({ onRequestAppeal, confirm }) {
       >
         {actionLabel}
       </Button>
-      <Info mode="description">
-        Anyone holding the required appeal deposit can trigger this action.
+      <Info>
+        <strong>Anyone</strong> can{' '}
+        <strong>
+          lock DAI as collateral to{' '}
+          {confirm ? 'confirm an appeal' : 'propose an appeal'}{' '}
```
sohkai: Bit of a nitpick, but "propose an appeal" sounds awkward in English—if there's no taker on the other side, you automatically win. Is this what we use in the user guide and etc?

What do you think about "begin" or "initiate" or "advocate for", as they're more firm than "propose"?

src/components/Disputes/actions/DisputeAppeal.js
```diff
@@ -24,8 +24,22 @@ function DisputeAppeal({ onRequestAppeal, confirm }) {
...
+        </strong>
+        , if they don’t agree with the current outcome. When the final ruling is
```
sohkai:
```suggestion
        if they believe the current outcome is incorrect. When the final ruling is
```

However, in the case of the person confirming the appeal, what they're actually doing is siding with the current outcome, no?

The appealer believes the ruling is incorrect; the confirmer invests against the appealer (and hence believes the current ruling is correct).

src/components/Disputes/actions/DisputeAppeal.js
```diff
@@ -24,8 +24,22 @@ function DisputeAppeal({ onRequestAppeal, confirm }) {
...
+        confirmed, the user who {confirm ? 'confirmed the appeal' : 'appealed'}{' '}
+        gets rewarded if the ruling has switched in their favor. If not, their
+        entire collateral could be re-distributed to the winning party.{' '}
+        {confirm
+          ? 'If an appeal is confirmed, a new adjudication round is initiated and a new jury is drafted.'
```
sohkai:
```suggestion
          ? 'If an appeal is confirmed, a new adjudication round is initiated and a new jury is drafted. '
```

Although in this situation, I think having this line in both cases is useful; it's just switching the "If" for a "When".

src/components/Disputes/actions/DisputeAppeal.js
```diff
@@ -24,8 +24,22 @@ function DisputeAppeal({ onRequestAppeal, confirm }) {
...
+          : ''}
+        <Link href="https://help.aragon.org/article/43-dispute-lifecycle#appeal">
+          Learn more
```
sohkai: Let's add a period afterwards (we should always end the "Learn more"s with a period).

src/components/Disputes/actions/DisputeAppeal.js
fabriziovigevani: Yes, good suggestion

src/components/Disputes/actions/DisputeAppeal.js
fabriziovigevani: I think that that would be the most common case yes.
However the account confirming an appeal can appeal for any ruling that is opposed to the ruling the appealer ruled for. So e.g let's say that the jury vote outcome was `In Favor`, someone doesn't agree with this outcome so he/she appeals `Against`. The one that confirms the appeal can rule for `In favor` (most common case) or `Refuse`. I don't see the reason of someone ruling for the later option  but it could happen.

src/components/Disputes/actions/DisputeAppeal.js
fabriziovigevani: Yes i think it makes sense, as you well say, proposing is a weak word and doesn't reflect properly the intentions of the appealer.

src/components/Disputes/actions/DisputeAppeal.js
fabriziovigevani: Have changed the wording for `if they believe the ruling appealed for is incorrect` in the case that the action is confirm appeal. Let me know what do you think

sohkai: Sounds good; I also don't know about `Refuse` although in the current situation with the first precedence campaign dispute, someone could actually appeal for "refuse" to help invalidate the ruling.

PR #177, opened 2020-02-17
Rewards module: Withdraw funds from treasury

fabriziovigevani:
Since after settling rewards and appeal deposits, tokens are not sent directly to the account but to the treasury, we need an extra action to withdraw them from there.

This PR includes this interaction.


src/utils/dispute-utils.js
```diff
@@ -446,7 +446,8 @@ export function getRoundFees(round, courtConfig) {
   if (round.number === maxRegularAppealRounds)
     return round.jurorsNumber
       .mul(jurorFee)
-      .div(FINAL_ROUND_WEIGHT_PRECISION.mul(finalRoundReduction).div(PCT_BASE))
+      .div(FINAL_ROUND_WEIGHT_PRECISION)
+      .mul(finalRoundReduction.div(PCT_BASE))
```
sohkai: Depending on what we're expecting (and how the contracts are programmed), doing the `.div()` separately may result in more precision.

src/hooks/query-hooks.js
```diff
@@ -24,6 +24,25 @@ export function useJurorDraftQuery(jurorId) {
     : []
 }
 
+/**
+ * Queries if the juror has already recieve rewards
+ *
+ * @param {String} jurorId Address of the juror
+ * @returns {Object} All `jurorId` drafts not yet rewarded
```
sohkai: Maybe I'm missing something... but doesn't this just return a boolean? Should it actually be returning `data.juror.drafts` if that exists and is an array?

src/components/Dashboard/NoRewards.js
```diff
@@ -1,12 +1,20 @@
...
+import { useWallet } from 'use-wallet'
 
+import MessageCard from '../MessageCard'
+import { useJurorDraftRewardedQuery } from '../../hooks/query-hooks'
 import iconNoRewardsSvg from '../../assets/IconNoRewards.svg'
 
 function NoRewards() {
+  const wallet = useWallet()
+  const hasJurorClaimedRewards = useJurorDraftRewardedQuery(
+    wallet?.account.toLowerCase()
```
sohkai: I don't think we handle the case where the `wallet.account` doesn't exist in `useJurorDraftRewardedQuery`. Should it?

As well, do subscriptions require accounts to be in lower case (because that's how they're encoded in the subgraph)? If so, that also sounds like something the query function should handle for us transparently.

src/components/Dashboard/RewardsModule.js
```diff
@@ -31,50 +33,72 @@ const useTotalDisputesFees = (arbitrableFees, appealFees) => {
...
+  // TODO: Handle possible multiple tokens (fee token can change)
+  const treasuryToken = treasury?.find(({ token }) =>
+    addressesEqual(token.id, feeToken.id)
+  )
+  const treasuryBalance = treasuryToken ? treasuryToken.balance : bigNum(0)
+  const totalDisputesFees = totalArbitrableFees.add(totalAppealFees)
+  const totalFees = totalDisputesFees.add(treasuryBalance)
 
   // Form submission
   const handleFormSubmit = async event => {
```
sohkai: This should probably be in a `useCallback()`

src/components/Dashboard/RewardsModule.js
```diff
@@ -31,50 +33,72 @@ const useTotalDisputesFees = (arbitrableFees, appealFees) => {
...
-    try {
-      // Claim all arbitrable fee rewards
-      const arbitrableTxs = []
-      for (const arbitrableFee of rewards.arbitrableFees) {
-        const { disputeId, rounds } = arbitrableFee
-        for (const roundId of rounds) {
-          arbitrableTxs.push(
-            await onSettleReward(disputeId, roundId, wallet.account)
-          )
+      const transactionBag = []
```
sohkai: Don't know if we really want to call this a "bag" (it's an awkward bit of history from the client).

Perhaps `rewardTransactionQueue`?

src/components/Dashboard/RewardsModule.js
```diff
@@ -31,50 +33,72 @@ const useTotalDisputesFees = (arbitrableFees, appealFees) => {
...
+          await onWithdraw(feeToken.id, wallet.account, totalFees)
+        )
+
+        await Promise.all(transactionBag.map(tx => tx.wait()))
+      } catch (err) {
+        console.log(`Error claiming rewards: ${err}`)
+      }
     }
+
+    if (!rewards) return
```
sohkai: We already check this in `claimRewards()`, but do we really need to declare the function if all we're going to do is call it immediately? Why not just inline it?

src/components/Dashboard/RewardsModule.js
```diff
@@ -162,104 +185,129 @@ const DisputesFees = ({
...
+          css={`
+            margin-bottom: ${2 * GU}px;
+          `}
+        />
+        <Button
+          mode="positive"
+          type="submit"
+          wide
+          css={`
+            margin-bottom: ${showInfoSection ? 2 * GU : 0}px;
```
sohkai: This is likely more easily done / clearer by just adding a `margin-top` to the `Info` component, as it's only adding padding if the `Info` exists.

src/components/Dashboard/RewardsModule.js
```diff
@@ -162,104 +185,129 @@ const DisputesFees = ({
...
+          `}
+        >
+          Claim rewards
+        </Button>
+        {showInfoSection && (
+          <Info>
+            This action requires multiple transactions to be signed in Metamask.
```
sohkai:
```suggestion
            This action requires multiple transactions to be signed by your account.
```

@bpierre Do we have any utilities here or in `use-wallet` that help detect which wallet / provider is being used?

bpierre: Yes we have a function for that in this project! https://github.com/aragon/court-dashboard/blob/58dda7601dd870e2f6ec0bb1f68ca989aaac294b/src/ethereum-providers/index.js#L99

To use it:

```jsx
const { activated } = useWallet()
const provider = getProviderFromUseWalletId(activated)
```

src/hooks/query-hooks.js
fabriziovigevani: Yes, the intention is to return a Boolean, the JS docs are off. Initially had implemented it to return the array of drafts already rewarded but then i thought it would be more straight forward to just tell if the array has an item which means that the juror has claimed rewards at some point.

src/components/Dashboard/NoRewards.js
fabriziovigevani: > I don't think we handle the case where the wallet.account doesn't exist in useJurorDraftRewardedQuery. Should it?

Well actually this component is only rendered in the case that an account is connected to the dashboard so the `wallet?.` check doesn't make sense in this case. Will update and add a comment.

> As well, do subscriptions require accounts to be in lower case (because that's how they're encoded in the subgraph)? If so, that also sounds like something the query function should handle for us transparently. 

Yes, that is the reason, will look if urql provides functions before sending the request.

src/components/Dashboard/RewardsModule.js
fabriziovigevani: Now that we are not passing the function via props i think that is not needed any more ?

src/components/Dashboard/RewardsModule.js
fabriziovigevani: We certainly don't need it declared. This was the result of a previous use of `useCallback`

src/components/Dashboard/RewardsModule.js
sohkai: It seems like `handleFormSubmit()` is still being passed to the `form`'s `onsubmit` prop.

But usually, the rule of thumb:
- If the function is dependent on component state / props, probably best to wrap it in a `useCallback()` so it doesn't get re-created on each render
- Otherwise, it could likely be moved to be outside of the component

PR #179, opened 2020-02-18
AccountBanner: new circle graph and info

### LINKED ISSUE #102: Update CircleGraph

### LINKED ISSUE #163: Add more explanation for low ANJ balances
- [ ] Write HelpScout article
- [x] Implement new banner with help text

rperez89:
Fixes #102 and fixes #163.

### DIFF
--- a/src/components/CircleGraph.js
+++ b/src/components/CircleGraph.js
@@ -12,7 +12,7 @@ const LABEL_DEFAULT = value => `${Math.round(value * 100)}`
 
 const { span: AnimatedSpan, circle: AnimatedCircle } = animated
 
-function CircleGraph({ value, label, size, strokeWidth }) {
+function CircleGraph({ value, comparisonOperator, label, size, strokeWidth }) {
   const theme = useTheme()
   const length = Math.PI * 2 * (size - strokeWidth)
   const radius = (size - strokeWidth) / 2
@@ -60,27 +60,28 @@ function CircleGraph({ value, label, size, strokeWidth }) {
           <div
             css={`
               display: flex;
-              align-items: flex-start;
+              align-items: baseline;
+              margin-left: ${comparisonOperator ? `0px ` : `${0.5 * GU}px`};
             `}
           >
+            {comparisonOperator && <div>{comparisonOperator}</div>}
             <AnimatedSpan
               css={`
-                font-size: ${2 * GU}px;
+                font-size: 20px;
               `}
             >
               {progressValue.interpolate(t =>
                 label(Math.min(1, Math.max(0, t)))
               )}
             </AnimatedSpan>
-            <div>
-              <span
-                css={`
-                  font-size: ${1.5 * GU}px;
-                  color: ${theme.surfaceContentSecondary};
-                `}
-              >
-                %
-              </span>
+            <div
+              css={`
+                display: flex;
+                font-size: 12px;
+                color: ${theme.surfaceContentSecondary};
+              `}
+            >
+              %
             </div>
           </div>
         </Main>
--- a/src/components/CircleGraph.js
+++ b/src/components/CircleGraph.js
@@ -67,7 +67,7 @@ function CircleGraph({ value, comparisonOperator, label, size, strokeWidth }) {
             {comparisonOperator && <div>{comparisonOperator}</div>}
             <AnimatedSpan
               css={`
-                font-size: 20px;
+                font-size: 21px;
               `}
             >
               {progressValue.interpolate(t =>
--- a/src/components/CircleGraph.js
+++ b/src/components/CircleGraph.js
@@ -7,25 +7,27 @@ import { GU, useTheme } from '@aragon/ui'
 const STROKE_WIDTH = 4
 
 const SIZE_DEFAULT = 80
-const SUFFIX_DEFAULT = '%'
-const VALUE_DEFAULT = value => {
+const VALUE_DEFAULT = 1
+const LABEL_DEFAULT = value => {
   if (value === 0) {
-    return '0'
+    return { value: '0', suffix: '%' }
   }
   if (Math.round(value * 100) < 1) {
-    return '1'
+    return { prefix: '<', value: '1', suffix: '%' }
   }
-  return String(Math.round(value * 100))
+  return {
+    value: String(Math.round(value * 100)),
+    suffix: '%',
+  }
 }
-const PREFIX_DEFAULT = value => (Math.round(value * 100) < 1 ? `<` : '')
-const LABEL_DEFAULT = value => `${Math.round(value * 100)}`
 
 const { span: AnimatedSpan, circle: AnimatedCircle } = animated
 
-function CircleGraph({ value, prefix, suffix, label, size, strokeWidth }) {
+function CircleGraph({ value, label, size, strokeWidth }) {
   const theme = useTheme()
   const length = Math.PI * 2 * (size - strokeWidth)
   const radius = (size - strokeWidth) / 2
+  const { prefix, value: stringValue, suffix } = label(value)
   return (
     <Spring to={{ progressValue: value }} native>
       {({ progressValue }) => (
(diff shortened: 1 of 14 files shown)


rperez89:
> API of the CircleGraph, but

where should we add that help link? into the help popup?

src/components/Dashboard/AccountBanner.js
```diff
@@ -228,19 +218,11 @@ const BannerWithProbability = ({ activeBalance }) => {
...
-            height={6 * GU}
-            src={anjSpringIcon}
-            alt=""
-          />
-        ) : (
-          <CircleGraph value={draftingProbability} size={6 * GU} />
-        )
+        <CircleGraph
+          value={probablilityTooLow ? 0.01 : draftingProbability}
+          comparisonOperator={probablilityTooLow && '<'}
```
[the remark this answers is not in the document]
rperez89: you mean to change the `comparisonOperator` to something like `customLabel`?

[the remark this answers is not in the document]
bpierre: Note: this is more a discussion for moving the component into aragonUI, I think we could merge this PR without it.

I wonder if it could be a good idea to have a `label` prop as a function returning an object, so that all the styling could be done internally. This object would contain three values:

- `value`, the actual number.
- `prefix`, the optional `<`.
- `suffix`, the `%`.

The default could be defined this way to match what we currently have:

```jsx
CicleGraph.defaultProps = {
  label(value) {
    if (value === 0) {
      return { value: '0', suffix: '%' }
    }
    if (Math.round(value * 100) < 1) {
      return { prefix: '<', value: '1', suffix: '%', }
    }
    return {
      value: String(Math.round(value * 100)),
      suffix: '%',
    }
  }
}
```

What do you guys think?

src/components/CircleGraph.js
```diff
@@ -0,0 +1,117 @@
...
+            css={`
+              display: flex;
+              align-items: baseline;
+              margin-left: ${comparisonOperator ? `0px ` : `${0.5 * GU}px`};
+            `}
+          >
+            {comparisonOperator && <div>{comparisonOperator}</div>}
+            <AnimatedSpan
+              css={`
+                font-size: 20px;
```
bpierre: Should this be `21px` (based on the Figma file)?

src/components/Dashboard/AccountBanner.js
```diff
@@ -176,9 +177,11 @@ const BannerWithProbability = ({ activeBalance }) => {
   // Calculate probability (since the total active balance is asynchronous
   // it can happen that it has not been updated yet when the juror active balance has)
   const draftingProbability = Math.min(1, totalPercentage / 100)
-
   const probablilityTooLow = totalPercentage < 1
-  const chances = totalPercentage > 0 && Math.floor(100 / totalPercentage)
+
+  const chances = probablilityTooLow
+    ? '100+'
+    : totalPercentage > 0 && Math.floor(100 / totalPercentage)
```
bpierre: Isn’t `totalPercentage > 0` always true if `probabilityTooLow` is `false`?

src/components/Dashboard/AccountBanner.js
```diff
@@ -187,60 +190,54 @@ const BannerWithProbability = ({ activeBalance }) => {
...
+          margin-right: ${1 * GU}px;
+        `}
+      >
+        On average, you will be drafted into a jury
+        <span
+          css={`
+            color: ${theme.accent};
+          `}
+        >
+          {` 1 in ${chances} times `}
```
bpierre:
```suggestion
          1 in {chances} times
```

src/components/Dashboard/AccountBanner.js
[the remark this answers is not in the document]
rperez89: i think that is a good approach it makes the implementation  more generic

src/components/Dashboard/AccountBanner.js
rperez89: oh, that is to avoid to divide by 0 given that `totalPercentage` is calculated async

src/components/Dashboard/AccountBanner.js
```diff
@@ -187,61 +190,48 @@ const BannerWithProbability = ({ activeBalance }) => {
...
+          This is a numerical estimate of your likelihood of being selected for
+          arbitration. It’s calculated by dividing your active ANJ balance
+          against the Court's total active ANJ balance during the current term.
+        </p>
+        <p
+          css={`
+            margin-top: ${1 * GU}px;
+          `}
+        >
+          You currently have &lt;1% of all activated ANJ and hence are unlikely
```
sohkai: I think we should only be showing this if the percentage is actually <1%, right?

src/components/Dashboard/AccountBanner.js
sohkai: Perhaps we can restructure this component based on this loading state.

If it's still loading, we render `BannerLoadingRing`, otherwise, we render the `Wrapper` (and find a better name for it :smile:) and that component should handle all logic with the assumption that balances have already been loaded.

In the case of not wanting to divide by 0; we get `Infinity` back, and we can simply use:

```
const chances = Math.floor(100 / totalPercentage)
const chancesText = chances > 100 ? '100+' : chances
```

rperez89: not sure if i understood why we should change that is working now

sohkai: The reason I suggest a restructuring is that in the current `BannerWithProbability ` component, there is a lot of logic dependent on the asynchronously fetched variable, `totalActiveBalanceCurrentTerm`.

The resulting logic ([l172-184](https://github.com/aragon/court-dashboard/blob/ba3ee791a97fb4da87b9ac8e7ef1d86023d01e51/src/components/Dashboard/AccountBanner.js#L172-L184)) would be easier to reason about if we separated the component such that any logic reliant on `totalActiveBalanceCurrentTerm` could assume that it has been resolved.

For example, right now, it's unclear what happens if `getPercentageBN()` is passed `bn(-1)` as the denominator; I assume it handles it, but it puts this bit of code in a weird position in that there's an extra case to think through and be responsible for.

Moreover, this component's render is already in a good position to be logically restructured: we can immediately short circuit and render `BannerLoadingRing` when we know `totalActiveBalanceCurrentTerm` is still being fetched.

rperez89: Gotcha! thanks!

src/components/Dashboard/AccountBanner.js
```diff
@@ -205,21 +205,37 @@ const BannerWithProbability = ({ activeBalance }) => {
...
+                >
+                  You currently have &lt;1% of all activated ANJ and hence are
+                  unlikely to be drafted unless a dispute goes to the final
+                  round or many disputes are created. Activate more ANJ to
+                  increase your chances of being selected as a juror.
+                </p>
+              </>
+            )
+          }
+          return (
```
sohkai: Ah, I wonder if we should just use the same first paragraph in both cases, but only show the second one if you have < 1%?

rperez89: that is not what we have on the  [figma files](https://www.figma.com/file/IdVA6iMn0cWLV5imylwEUP/%E2%9C%85-MASTER%3A-Court-Dashboard?node-id=3450%3A0)

sohkai: Ah, I think that's because it doesn't have the copy adjustments made after the fact.

In the Figma file, both of the initial paragraphs are the same.

rperez89: Screen Shot 2020-03-05 at 3 16 35 PM

not sure which one should i follow because those are differents when you have less than 1% or more than 1

sohkai: Hmm yes, I only saw the left and that one suggested the two had the same first paragraphs.

I think what we have now is clear (I think the copy on the right side is more clear than the left) and let's leave it until more people say they're confused 😅

facuspagnuolo:
> LGTM; as mentioned offline, I'm just wondering if we can get total active balance through the subgraph instead of polling.

Sure, already created [this issue](https://github.com/aragon/court-subgraph/issues/47) to tackle that

PR #186, opened 2020-02-18
Disputes: support final rounds

fabriziovigevani:
Includes all the dispute logic updates to support final rounds.

Also addes some JS docs


src/components/Disputes/DisputeActions.js
```diff
@@ -286,6 +301,7 @@ const useInfoAttributes = (
     negativeBackground,
     phase,
     positiveBackground,
+    status,
     theme.accent,
```
rperez89: should this `theme.accent` be as a dependency here?

src/components/Disputes/actions/DisputeVoting.js
```diff
@@ -2,13 +2,50 @@ import React from 'react'
...
 } from '../../../utils/crvoting-utils'
 
-function DisputeVoting({ isJurorDrafted, onRequestCommit }) {
+function DisputeVoting({
+  draftTermId,
+  isFinalRound,
+  isJurorDrafted,
+  onRequestCommit,
+}) {
+  return isFinalRound ? (
```
rperez89: could  we do something like 
```
 if (isFinalRound) {
   return <VotingFinalRound.... />
}
return <VotingActions ... />
```

src/utils/appeal-utils.js
```diff
@@ -18,6 +18,11 @@ export function transformAppealDataAttributes(appeal) {
       dispute: {
         ...round.dispute,
         lastRoundId: parseInt(round.dispute.lastRoundId, 10),
+        rounds: round.dispute.rounds.map(round => ({
+          ...round,
+          number: parseInt(round.number, 10),
```
rperez89: maybe we can do those parse when we are transforming the dispute atrributes , `transformResponseDisputeAttributes` inside `dispute-utils`

src/components/Disputes/DisputeActions.js
fabriziovigevani: Yes it should, the accent could change (e.g in a dark theme)

src/utils/appeal-utils.js
fabriziovigevani: Not sure i understand what you mean,  the dispute for the appeals don't have the full data so don't think we should use`transformResponseDisputeAttributes`

rperez89: forgot about it , i read dispute-utils in the file name haha

src/components/Disputes/DisputeActions.js
rperez89: nice

src/hooks/useCanJurorVoteFinalRound.js
```diff
@@ -0,0 +1,10 @@
+import { useCourtConfig } from '../providers/CourtConfig'
+import { useActiveBalanceOfAt } from './useCourtContracts'
+
+export default function useCanJurorVoteFinalRound(account, draftTermId) {
```
sohkai:
```suggestion
export default function useCanJurorVoteFinalRound(account, disputeDraftTermId) {
```

I (think I) understand this, but it took a bit to realize we needed the checkpointed balance of a juror because it's based on the dispute's draft term. The original naming is also a bit weird because final rounds are associated with disputes, not draft terms.

src/utils/dispute-utils.js
```diff
@@ -140,6 +146,12 @@ export function getDisputeTimeLine(dispute, courtConfig) {
   return timeLine
 }
 
+/**
+ * @param {Object} dispute The dispute querying current phase and transition of
```
sohkai:
```suggestion
 * @param {Object} dispute The dispute to query the current phase and next transition of
```

src/utils/dispute-utils.js
```diff
@@ -140,6 +146,12 @@ export function getDisputeTimeLine(dispute, courtConfig) {
...
+ * @param {Object} courtConfig The court configuration
+ * @param {Date} nowDate Current date
```
sohkai: `nowDate` seems like it could be renamed to something else; otherwise we could just use `dayjs()` in the function itself?

src/utils/dispute-utils.js
```diff
@@ -208,26 +225,36 @@ export function getPhaseAndTransition(dispute, courtConfig, nowDate) {
   }
 }
 
+/**
+ * Tells the adjudication state of a round at a certain time.
```
sohkai:
```suggestion
 * Tells the adjudication state of a dispute's round at a certain time.
```

src/utils/dispute-utils.js
```diff
@@ -208,26 +225,36 @@ export function getPhaseAndTransition(dispute, courtConfig, nowDate) {
...
+ * @param {Object} dispute Dispute querying the adjudication round of
```
sohkai:
```suggestion
 * @param {Object} dispute Dispute to query the adjudication round of
```

src/utils/dispute-utils.js
```diff
@@ -208,26 +225,36 @@ export function getPhaseAndTransition(dispute, courtConfig, nowDate) {
...
+ * @param {Object} round The round that is being queried
+ * @param {Number} now Current time in ms
```
sohkai: It seems weird that a previous API has the `now` be a Date object whereas here (and in some other functions) it's a number. We should pick one and stick with it across the internal API.

src/utils/dispute-utils.js
```diff
@@ -208,26 +225,36 @@ export function getPhaseAndTransition(dispute, courtConfig, nowDate) {
...
     revealTerms,
     appealTerms,
     appealConfirmationTerms,
-    maxRegularAppealRounds,
   } = courtConfig
-  const numberOfRounds = dispute.rounds.length
+
   const { draftTermId, delayedTerms, number: roundId } = round
 
   const drafTermStartTime = getTermStartTime(draftTermId, courtConfig)
```
sohkai:
```suggestion
  const draftTermStartTime = getTermStartTime(draftTermId, courtConfig)
```

Small misspelling :)

src/utils/dispute-utils.js
```diff
@@ -208,26 +225,36 @@ export function getPhaseAndTransition(dispute, courtConfig, nowDate) {
...
+  const maxAppealReached = hasDisputeReachedMaxAppeals(dispute, courtConfig)
 
+  // Case where we are at the last possible round and voting period has not started yet
   if (
-    DisputesTypes.Phase[round.state] === DisputesTypes.Phase.Invalid &&
-    now < drafTermStartTime
+    now < drafTermStartTime &&
+    (DisputesTypes.Phase[round.state] === DisputesTypes.Phase.Invalid ||
```
sohkai: Does `DisputesTypes.Phase[round.state] === DisputesTypes.Phase.Invalid` hold for `Phase.NotStarted`? This seems like still an invalid state.

src/utils/dispute-utils.js
```diff
@@ -306,6 +334,15 @@ export function getAdjudicationPhase(dispute, round, now, courtConfig) {
   }
 }
 
+/**
+ * @dev Terminology here will be:
+ *            Last round => dispute's last round
+ *            Final round => max possible round for a dispute (when the max appeals for a given dispute is reached)
```
sohkai: Nitpick: indentation here seems a bit weird.

I always have a hard time choosing how to indent these multi-line JSDoc strings, but usually 2 spaces past the start of the previous line looks OK (it is currently 6).

src/utils/dispute-utils.js
```diff
@@ -306,6 +334,15 @@ export function getAdjudicationPhase(dispute, round, now, courtConfig) {
   }
 }
 
+/**
+ * @dev Terminology here will be:
+ *            Last round => dispute's last round
```
sohkai:
```suggestion
 *            Last round => last round actually reached in a dispute
```

This might be clearer

src/utils/dispute-utils.js
```diff
@@ -306,6 +334,15 @@ export function getAdjudicationPhase(dispute, round, now, courtConfig) {
...
+ *            Final round => max possible round for a dispute (when the max appeals for a given dispute is reached)
+ * @param {Object} courtConfig The court configuration
+ * @param {Object} round The round to get the phases from
+ * @param {Object} currentPhase The dispute current phase
```
sohkai:
```suggestion
 * @param {Object} currentPhase The dispute's current phase
```

src/utils/dispute-utils.js
```diff
@@ -403,32 +440,49 @@ function getRoundPhasesAndTime(courtConfig, round, currentPhase) {
     },
   ]
 
+  // If it's not the last round means that it was appealed and confirm appealed
```
sohkai: I wonder if we can come up with a different wording for "appealed and confirm appealed", like "the dispute reached the next appeal round"?

src/utils/dispute-utils.js
```diff
@@ -403,32 +440,49 @@ function getRoundPhasesAndTime(courtConfig, round, currentPhase) {
...
+  // When round has not yet ended
+  return roundPhasesAndTime.slice(
+    currentPhase.maxAppealReached ? 1 : 0,
+    currentPhaseIndex + 1
+  )
 }
 
+/**
+ *
+ * @param {Object} dispute The dispute in cuestion
```
sohkai:
```suggestion
 * @param {Object} dispute The dispute in question
```

src/utils/dispute-utils.js
```diff
@@ -437,10 +491,32 @@ function getEvidenceSubmissionEndTime(dispute, courtConfig) {
   return getTermStartTime(firstRound.draftTermId, courtConfig)
 }
 
+/**
+ * Tells wether the dispute has reached the maximum number of rounds possible
```
sohkai:
```suggestion
 * Tells whether the dispute has reached the maximum number of rounds possible
```

src/utils/dispute-utils.js
```diff
@@ -437,10 +491,32 @@ function getEvidenceSubmissionEndTime(dispute, courtConfig) {
...
+ * @param {Object} dispute The dispute in cuestion
```
sohkai:
```suggestion
 * @param {Object} dispute The dispute in question
```

src/utils/dispute-utils.js
```diff
@@ -437,10 +491,32 @@ function getEvidenceSubmissionEndTime(dispute, courtConfig) {
...
+ */
 export function getDisputeLastRound(dispute) {
   return dispute.rounds[dispute.lastRoundId]
 }
 
+/**
+ *
+ * @param {Object} round Round to calculate fees from
+ * @param {Object} courtConfig The court configuration
+ * @returns {BigNum} The total `round` fees
```
sohkai:
```suggestion
 * @returns {BigNum} The total fees for the round
```

src/utils/dispute-utils.js
```diff
@@ -464,6 +540,13 @@ export function getRoundFees(round, courtConfig) {
     .mul(round.jurorsNumber)
 }
 
+/**
+ *
+ * @param {BigNum} minActiveBalance The minimum active balance required to become an active juror
+ * @param {BigNum} penaltyPct Permyriad of min active tokens balance to be locked to each drafted juror
```
sohkai: > Permyriad of min active tokens balance

I really don't know what this means haha, it looks like "pyramid" to me but that can't be right!

src/utils/dispute-utils.js
```diff
@@ -464,6 +540,13 @@ export function getRoundFees(round, courtConfig) {
...
+ * @param {BigNum} weight Weight computed for a juror on a round
+ * @returns {BigNum} The amount locked to a drafted juror
```
sohkai:
```suggestion
 * @returns {BigNum} The amount that will be locked each time a juror is drafted
```

src/utils/appeal-utils.js
```diff
@@ -18,6 +18,11 @@ export function transformAppealDataAttributes(appeal) {
...
+          jurorsNumber: bigNum(round.jurorsNumber),
```
sohkai: Just wondering, is it possible to reach a number that native JS numbers won't work? E.g. is this value always increasing, even if past jurors deactivate?

I think it is highly unlikely for us to get above 2^53 active jurors.

src/hooks/useCourtContracts.js
```diff
@@ -351,6 +351,32 @@ export function useAppealFeeAllowance(owner) {
...
+      retryMax(() => jurorRegistryContract.activeBalanceOfAt(juror, termId))
+        .then(balance => {
+          setActiveBalance(balance)
+        })
+        .catch(err => {
+          console.error(`Error fetching active balance for juror : ${err}`)
+        })
+    }
+
+    getActiveBalanceOfAt()
```
sohkai: We should use the ["cancel" pattern](https://github.com/aragon/court-dashboard/blob/development/src/hooks/useTokenBalanceToUsd.js#L89) here, where we avoid setting state if the hook has been re-run.

If we're using timers to retry, we should also clear them in the return of the effect.

src/hooks/useCourtContracts.js
```diff
@@ -351,6 +351,32 @@ export function useAppealFeeAllowance(owner) {
...
+
+  useEffect(() => {
+    const getActiveBalanceOfAt = async () => {
+      if (!jurorRegistryContract) return
+
+      retryMax(() => jurorRegistryContract.activeBalanceOfAt(juror, termId))
+        .then(balance => {
+          setActiveBalance(balance)
+        })
+        .catch(err => {
```
sohkai: I'm not sure this will work as intended; `retryMax()` only retries if the inner function rejects with an error.

Here, we catch any errors in the promise and log it without re-throwing it.

src/hooks/useCanJurorVoteFinalRound.js
fabriziovigevani: Yeah, thats correct. 
This makes me realize that we probably don't need a hook for this and we can do it directly on the `DisputeVoting` component where we explain the use of the hook

src/utils/dispute-utils.js
fabriziovigevani: Yes good catch. I'm refactoring this bit of code and adding more comments

src/utils/dispute-utils.js
fabriziovigevani: Haha i wasn't familiar with the word either until i looked at the contracts :D 
For what i understand it means `a % of..` 

Will change the term :)

rperez89: Looool

src/utils/dispute-utils.js
fabriziovigevani: Yeah, changed the naming as well as the type in favor of using `dayjs()`. I think it makes more sense to stick with passing the date as parameter since we are passing the result of `useNow` in some places e.g in dispute hooks.

Also changed `useNow` to set a `dayjs` object as opposed to the `Date` object so we can use it transparently.

src/utils/dispute-utils.js
fabriziovigevani: Yep totally makes sense.

src/utils/appeal-utils.js
fabriziovigevani: For normal rounds this number won't be too high. It's calculated based on the previous round `jurorsNumber` multiplied by a `appealStepFactor`.
E.g. Current court on mainnet is configured with 3 jurors for first round and an `appealStepFactor` of 3 with a max appeal rounds of 4 so at most it will be 234.

For final rounds the `jurorsNumber` is the number of times the minimum required active balance is held in the registry at the round start term, multiplied  by a final round precision factor of `1000`. 

With current court configurations i think i may be also unlikely for it to get above 2 ^ 53 so we could not convert it to a big number. 
What do you think?

src/hooks/useCourtContracts.js
fabriziovigevani: Well, yes this was originally the intention. What do you think we should do when the max retry attempts has been reached? 

Should we re throw the error and let it be handled by the global error?

src/hooks/useCourtContracts.js
fabriziovigevani: Yeah good point, will add the pattern to all these type of functions.

src/utils/appeal-utils.js
sohkai: Yes, I think it's reasonable to assume we don't need to convert it to a big number then. We will have much bigger problems if we reach 2^53 jurors.

src/hooks/useCourtContracts.js
sohkai: Maybe we can set an error state in this hook, so the component outside of it can choose to respond?

src/utils/dispute-utils.js
```diff
@@ -543,7 +547,7 @@ export function getRoundFees(round, courtConfig) {
 /**
  *
  * @param {BigNum} minActiveBalance The minimum active balance required to become an active juror
- * @param {BigNum} penaltyPct Permyriad of min active tokens balance to be locked to each drafted juror
+ * @param {BigNum} penaltyPct Percentage of min active tokens balance to be locked to each drafted juror
```
sohkai:
```suggestion
 * @param {BigNum} penaltyPct Percentage (against PCT_BASE) of min active tokens balance to be locked to each drafted juror
```

"Percentage" usually implies something smaller than 100, but we have 10^16 values here :)

src/hooks/useCourtContracts.js
```diff
@@ -386,27 +419,34 @@ export function useTotalActiveBalancePolling(termId) {
...
-          fetchTotalActiveBalance()
-        })
-    }, 1000)
-  }, [jurorRegistryContract, termId])
-
   useEffect(() => {
+    let cancelled = false
+
+    const fetchTotalActiveBalance = () => {
+      timeoutId.current = setTimeout(() => {
```
sohkai: Nitpick: there's not much point in using a ref now, since we've enclosed this within the `useEffect()`. We can just assign a local variable and clear it in the return.

src/hooks/useCourtContracts.js
```diff
@@ -386,27 +419,34 @@ export function useTotalActiveBalancePolling(termId) {
...
+          .then(balance => {
+            if (!cancelled) {
+              setTotalActiveBalance(balance)
+              clearTimeout(timeoutId.current)
+              fetchTotalActiveBalance()
+            }
+          })
+          .catch(err => {
+            console.error(`Error fetching balance: ${err} retrying...`)
+            fetchTotalActiveBalance()
```
sohkai: We forgot to guard this one with a `if (!cancelled)`.

However, what we probably want to do anyway is use a `.finally()` on the promise as we do this on both the resolve and reject cases.

src/hooks/useCourtContracts.js
```diff
@@ -410,54 +427,103 @@ export function useAppealFeeAllowance(owner) {
...
+    return () => {
+      cancelled = true
+    }
   }, [disputeManagerAddress, feeTokenContract, owner])
 
   return allowance
 }
 
-export function useTotalActiveBalancePolling(termId) {
+export function useActiveBalanceOfAt(juror, termId) {
```
sohkai: I mentioned this in https://github.com/aragon/court-dashboard/pull/179#pullrequestreview-369563941, but this seems like something the subgraph should provide rather than us having to poll for.

Is it because we could be arbitrarily asking for balances on any term, so keeping the full history in the subgraph is less ideal?

In any case, I also wonder if we should only do the polling if the requested `termId` is the current term. You would never have a previous term's active balance change, right?

src/hooks/useCourtContracts.js
fabriziovigevani: Yes!! Very good point

src/hooks/useCourtContracts.js
fabriziovigevani: > this seems like something the subgraph should provide rather than us having to poll for.

Yes, yesterday we had a call with @facuspagnuolo about this and next week we'll have this data available in the subgraph, see https://github.com/aragon/court-subgraph/issues/47.

> In any case, I also wonder if we should only do the polling if the requested termId is the current term

Yes, note that `useActiveBalanceOfAt` does not poll but only queries.

I think that your suggestion is the way to go, `useTotalActiveBalancePolling` could be replaced by  `useTotalActiveBalance`. And we already have this implemented by @rperez89  at https://github.com/aragon/court-dashboard/pull/229/files/8c46338753c545029b88e1dd79d071817e03998b#r395011351

src/hooks/useCourtContracts.js
```diff
@@ -365,11 +368,17 @@ export function useAppealDeposits(disputeId, roundId) {
...
+                amounts: [appealDeposit, confirmAppealDeposit],
+                error: null,
+              })
             }
           })
           .catch(err => {
-            console.error(`Error fetching appeal deposits: ${err}`)
+            setAppealDeposits(appealDeposits => ({
+              ...appealDeposits,
+              error: `Error fetching appeal deposits: ${err}`,
```
sohkai: A bit of API nitpick with the errors: in other hooks we're using a boolean value.

We effectively get the same with a string (any error state returned denotes an error if it's truthy), but we may want to consolidate how we design around errors.

Because we are in the same "context" here (i.e. we control both the place where the error is generated and where it'll be consumed), I think it's better to start simpler (ie. prefer bools to long strings). Returning the full error string likely isn't useful to any consumer—if we wanted to log it or send a metric event, we would do so at the error generation site.

In the future, if a consumer needs to branch based on the error type, we could return a short string or symbol.

src/hooks/useCourtContracts.js
```diff
@@ -494,25 +520,27 @@ export function useTotalActiveBalancePolling(termId) {
...
-              fetchTotalActiveBalance()
             }
           })
           .catch(err => {
             console.error(`Error fetching balance: ${err} retrying...`)
-            fetchTotalActiveBalance()
+          })
+          .finally(() => {
+            if (!cancelled) {
+              clearTimeout(timeoutId)
```
sohkai: I wonder if we need to do this; the related timer should have already been finished at this point (it is what invokes this function)?

src/hooks/useCourtContracts.js
fabriziovigevani: That totally makes sense. Thanks for the explanation.

src/hooks/useCourtContracts.js
fabriziovigevani: Ah yes good point, it was needed in the prior implementation but indeed we don't need it now.

PR #192, opened 2020-02-19
Disputes: apply override to voided disputes

fabriziovigevani:


src/voided-disputes.js
```diff
@@ -0,0 +1,34 @@
+import { networks, RINKEBY_COURT, RINKEBY_USABILITY_COURT } from './networks'
+import { getNetworkName } from './lib/web3-utils'
+import env from './environment'
+
+const DEFAULT_VOID_TEXT = ''
+const DEFAULT_VOID_LINK = ''
+
+export const VoidedDisputes = {
```
sohkai: Let's upper case this constant, otherwise it looks like a component or class

src/components/Disputes/DisputeDetail.js
```diff
@@ -64,50 +65,65 @@ const DisputeDetail = React.memo(function DisputeDetail({ match }) {
...
-            <Box heading="Dispute timeline" padding={0}>
-              {disputeFetching ? (
-                <div css="height: 200px" />
-              ) : (
-                <DisputeTimeline dispute={dispute} />
-              )}
-            </Box>
-          </React.Fragment>
+
+      {(() => {
```
sohkai: Nitpick, but we could lift `DisputeInfoComponent` out of the render so we don't need the IIFE here:

```js
</Bar>
{dispute?.status === DisputeStatus.Voided ? DisputeInfoComponent : (
  ...
)}
```

src/utils/dispute-utils.js
```diff
@@ -52,6 +61,20 @@ export const transformResponseDisputeAttributes = dispute => {
       }
     }),
   }
+
+  if (transformedDispute.status !== DisputesTypes.Status.Voided) {
+    return transformedDispute
+  }
+
+  // If the dispute is voided we will override certain data
+  return overrideVoidedDispute(transformedDispute)
```
sohkai: What do you think about moving all the transforms for the override into `overrideVoidedDispute`?

That way we could have something like:

```
const transformedDispute = {
  ... // normal values, as before
}

const isDisputeVoided = ...

return isDisputeVoted ? overrideVoidedDispute(transformedDispute) : transformedDispute
```

And `overrideVoidedDispute()` would handle the overrides for the status as well.

src/voided-disputes.js
```diff
@@ -0,0 +1,34 @@
...
+  rpc: new Map([[networks.rpc.court, []]]),
+  ropsten: new Map([[networks.ropsten.court, []]]),
+  rinkeby: new Map([
+    [RINKEBY_COURT, []],
+    [RINKEBY_USABILITY_COURT, []],
+  ]),
+  mainnet: new Map([
+    [
+      networks.mainnet.court,
+      [
```
sohkai: What do you think about making this a map of `dispute id -> voided information`?

src/utils/dispute-utils.js
```diff
@@ -4,22 +4,31 @@ import { getTermStartTime } from './court-utils'
...
 const juryDraftingTerms = 3
 
 export const transformResponseDisputeAttributes = dispute => {
-  return {
+  const voidedDisputes = getVoidedDisputesByCourt()
+  const isDisputeVoided = voidedDisputes.some(
+    voidedDispute => voidedDispute.id === dispute.id
+  )
+
+  const disputeStatus = isDisputeVoided
```
rperez89: could we create a function for this so we don't use those nested `?``:` ?

fabriziovigevani: Actually with the comment from @sohkai above this is resolved

src/components/Disputes/DisputeVoided.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react'
+
+function DisputeVoided({ id }) {
+  return (
+    <div>
+      <h2>Void Notice for Dispute #{id}</h2>
```
rperez89: here i think that we need to check if the voided text is an url and if it is we should display a link to the blogspot or whatever cc @sohkai

fabriziovigevani: Actually we are going to have a text and a link for each voided dispute in the `voided-disputes` file. 
And we will render the data like this https://www.figma.com/file/IdVA6iMn0cWLV5imylwEUP/%E2%9C%85-MASTER-Court-Dashboard?node-id=3930%3A344

src/components/Disputes/DisputeInfo.js
```diff
@@ -228,12 +236,12 @@ function DisputeHeader({ id, dispute }) {
               />
             )}
           </h1>
-          {transaction && (
+          {dispute?.status !== Status.Voided && transaction ? (
```
sohkai:
```suggestion
          {Boolean(dispute?.status !== Status.Voided && transaction) && (
```

And we can remove the ternary `null` afterwards

src/components/Disputes/DisputeInfo.js
```diff
@@ -75,15 +75,6 @@ const DisputeInfo = React.memo(function({
           if (loading) {
```
sohkai: We can probably move `loading` down to be after `isDisputeVoided` as well

src/voided-disputes.js
```diff
@@ -0,0 +1,37 @@
+import { networks, RINKEBY_COURT, RINKEBY_USABILITY_COURT } from './networks'
+import { getNetworkName } from './lib/web3-utils'
+import env from './environment'
+
+const DEFAULT_VOID_TEXT = 'This is a text'
```
sohkai:
```suggestion
const DEFAULT_VOID_TEXT = 'This dispute has been voided and discontinued'
```

src/voided-disputes.js
```diff
@@ -0,0 +1,37 @@
...
+            link: DEFAULT_VOID_LINK,
+            // TODO: Add texts and links when available
+          },
+        ].map(dispute => [dispute.id, dispute])
+      ),
+    ],
+  ]),
+}
+
+export function getVoidedDisputesByCourt() {
```
sohkai: Let's add a build-time flag that, if enabled, just returns an empty `Map` here.

src/voided-disputes.js
```diff
@@ -0,0 +1,37 @@
+import { networks, RINKEBY_COURT, RINKEBY_USABILITY_COURT } from './networks'
+import { getNetworkName } from './lib/web3-utils'
+import env from './environment'
+
+const DEFAULT_VOID_TEXT = 'This is a text'
+const DEFAULT_VOID_LINK = ''
+
+export const VOIDED_DISPUTES = {
```
sohkai: Should we avoid exporting this for now, since all internal code should be using `getVoidedDisputesByCourt()`?

src/components/Disputes/DisputeCard.js
```diff
@@ -15,18 +17,12 @@ function DisputeCard({ dispute, onSelectDispute }) {
...
     <CardItem onClick={handleClick}>
-      <div
-        css={`
-          display: flex;
-          align-items: center;
-          justify-content: space-between;
-        `}
-      >
+      <div>
         <DisputeStatus dispute={dispute} />
```
sohkai: Bit of a nitpick, but does this need to be enclosed in this div if we don't apply any styling to it?

src/components/Disputes/DisputeCard.js
```diff
@@ -39,32 +35,38 @@ function DisputeCard({ dispute, onSelectDispute }) {
...
-            height: ${27 * 3}px; // 27px * 3 = line-height * 3 lines
-            display: -webkit-box;
-            -webkit-box-orient: vertical;
-            -webkit-line-clamp: 3;
-          `}
-        />
+        {dispute.status !== Status.Voided && (
+          <>
+            <DisputeText
+              text={description}
```
sohkai: Wondering if we should still communicate something here, since still have the card.

src/components/Disputes/DisputeInfo.js
fabriziovigevani: In this case i don't think there's a difference since we won't know if the dispute is voided or not until `loading` ends. The loading is not for the `ipfs content` but for the `graph subscription`.

src/voided-disputes.js
fabriziovigevani: Ah yes, initially i was importing it from another file and forgot to remove the export after implementing `getVoidedDisputesByCourt `

src/components/Disputes/DisputeCard.js
fabriziovigevani: Actually the `<DisputeCard />` is styling it to be `100% width`. If we remove the div, it would be applied to the dispute status tag

src/components/Disputes/DisputeInfo.js
sohkai: Gotcha, this is specifically in the case where you open a URL with a dispute id and we're waiting to fetch information?

But in either case, we know the dispute id immediately, so it seems like we should be able to immediately prefer showing the voided state?

PR #193, opened 2020-02-19
Tasks: void task based on dispute id

rperez89:


src/hooks/useTasks.js
```diff
@@ -84,7 +87,7 @@ function useOpenTasks(tasks, now, courtSettings) {
     }
     return openTasks
     // Since we are using our own generated cache key we don't need to add the convertedTasks to the dependency list.
-  }, [convertedTasksPhasesKey]) // eslint-disable-line react-hooks/exhaustive-deps
+  }, [convertedTasksPhasesKey, voidedDisputes]) // eslint-disable-line react-hooks/exhaustive-deps
```
sohkai: `voidedDisputes` isn't tracked by React, so it doesn't need to be in the dependency list.

src/hooks/useTasks.js
```diff
@@ -35,10 +37,11 @@ function useOpenTasks(tasks, now, courtSettings) {
     if (!convertedTasks) {
       return []
     }
-
     const openTasks = []
     const incompleteTasks = convertedTasks.filter(
```
sohkai: Looking at this again, what do you think about lifting `incompleteTasks` into its own `useMemo()`, and then just using that directly in this later `useMemo()`?

src/hooks/useTasks.js
rperez89: mmhh added with the auto-save, let me check

src/hooks/useTasks.js
rperez89: Could you explain why that would be better? since  that new useMemo will also have as a dependency the `convertedTasksPhasesKey` like the the memo where it is right now

PR #194, opened 2020-02-19
Tasks: void task based on dispute id

rperez89:


src/hooks/useTasks.js
```diff
@@ -31,15 +33,22 @@ function useOpenTasks(tasks, now, courtSettings) {
...
+  const incompleteTasks = useMemo(() => {
+    if (!convertedTasks) {
+      return null
+    }
+    return convertedTasks.filter(
+      task =>
+        !voidedDisputes.has(task.dispute.id) &&
+        task.phase !== DisputesTypes.Phase.Ended
+    )
+  }, [convertedTasks]) // eslint-disable-line react-hooks/exhaustive-deps
```
fabriziovigevani: Was thinking that it would be better to make `convertedTasksPhasesKey ` the dependency here so we don't filter every second

src/hooks/useTasks.js
```diff
@@ -84,7 +93,7 @@ function useOpenTasks(tasks, now, courtSettings) {
     }
     return openTasks
     // Since we are using our own generated cache key we don't need to add the convertedTasks to the dependency list.
-  }, [convertedTasksPhasesKey]) // eslint-disable-line react-hooks/exhaustive-deps
+  }, [convertedTasksPhasesKey, incompleteTasks]) // eslint-disable-line react-hooks/exhaustive-deps
```
fabriziovigevani: Then here we would only use `incompleteTasks` as dependency
cc @sohkai

src/hooks/useTasks.js
rperez89: yeah, that was something that we were discussing yesterday night with @sohkai

rperez89: cc @sohkai

sohkai: Coming back a bit late on this, but yes, after discussion it sounds like we should've been using `convertedTasksPhasesKey` as @fabriziovigevani suggested.

rperez89: hahah yes, will create later a pr for updating this

PR #198, opened 2020-02-20
Dispute Detail: Display error on invalid dispute

### LINKED ISSUE #180: Disputes: invalid dispute id in URL results in always loading dispute
E.g. see https://court-rinkeby.aragon.org/disputes/100

The dispute detail ends up in an always "loading" state, but we should explain to the user that this dispute is not available.

fabriziovigevani:
closes #180 

TODO: 

- [x] Merge new #203 changes


src/components/Disputes/DisputeDetail.js
```diff
@@ -79,43 +69,52 @@ const DisputeDetail = React.memo(function DisputeDetail({ match }) {
...
-                ) : (
-                  <DisputeTimeline dispute={dispute} />
-                )}
-              </Box>
-            </React.Fragment>
-          }
+      {noDispute ? (
+        <ErrorLoading
+          subject="dispute"
+          errors={[`Dispute #${disputeId} does not exist`]}
```
rperez89: should we check this error msg with @dizzypaty ?

dizzypaty: > should we check this error msg with @dizzypaty ?

@rperez89 – See designs for this error below: 

dispute-not-found

Figma file: ✅ MASTER: Court Dashboard > Page: - Error pages
https://www.figma.com/proto/IdVA6iMn0cWLV5imylwEUP/%E2%9C%85-MASTER-Court-Dashboard?node-id=4158%3A52&viewport=170%2C1002%2C0.25&scaling=min-zoom

src/GlobalErrorHandler.js
```diff
@@ -55,11 +57,15 @@ class GlobalErrorHandler extends React.Component {
 
     return error ? (
       <GlobalErrorScreen>
-        <GenericError
-          detailsTitle={error.message}
-          detailsContent={errorStack}
-          reportCallback={SENTRY_DSN ? this.handleReportClick : null}
-        />
+        {error instanceof DisputeNotFound ? (
```
rperez89: Not sure if we should do this, this is the global error the dispute not found is not a global error is an specific error and should not be handled by the global error which intention is to capture all the errors that were not captured by any specific component error. So here i would suggest to re use the `GlobalErrorScreen` and pass a children with the specific error. But not handle the dispute error inside the global error

src/hooks/useDisputes.js
```diff
@@ -55,10 +55,35 @@ export default function useDisputes() {
...
+    ? convertToString(Object.values(disputePhase)[0])
+    : ''
+
+  return useMemo(() => {
+    if (fetching || (dispute && !disputeProcessed)) {
+      return { fetching: true }
+    }
+
+    return {
+      dispute: dispute
```
rperez89: should we use a function here to avoid returning with inline ternary operation?

src/hooks/useJurorSubscriptionFees.js
```diff
@@ -0,0 +1,44 @@
+import { useEffect, useState } from 'react'
```
rperez89: i think that this file does not belong to this pr right?

src/GlobalErrorHandler.js
fabriziovigevani: Not sure i follow. I think that this component is not meant to only capture unhandled generic errors but also any error that we would like to handle (i take that by `Global` doesn't reference a generic error but the error hub of the app). Doesn't make sense to me to pass another component as children that will use the `GlobalErrorScreen` when we are already using it here.

src/hooks/useDisputes.js
fabriziovigevani: I don't think this is a ternary operation?

src/hooks/useJurorSubscriptionFees.js
fabriziovigevani: Oops yeah definitely

src/GlobalErrorHandler.js
rperez89: you don't need the globalErrorHandler at all this is a dispute detail error the same than any of the dispute detail error, think about if not we are going to have some dispute errors here another dispute details error inside the dispute detail component, what you need is the `<GlobalErrorScreen/>` not the handler, also i think that you created the `disputeNotFound` inside a generic error folder instead inside the `Dispute` Folder as the other dispute errors

src/hooks/useDisputes.js
rperez89: `? ` `:` is the ternary operator

src/GlobalErrorHandler.js
rperez89: take a look [here](https://reactjs.org/docs/error-boundaries.html) if you want to what is the idea for this component

fabriziovigevani: I think we do need the global error handler. We need to treat this error different because we are going to display it in a [different way](https://github.com/aragon/court-dashboard/pull/198#discussion_r385010775) than the other types of dispute errors that are local to the dispute detail. Not sure if you are suggesting to use  `<GlobalErrorScreen />`  inside the dispute detail, but that would not work because it would render as a child of `<MainView />`.

src/hooks/useDisputes.js
fabriziovigevani: Ah thought you meant nested operations. What's the problem to use it like this? Don't see a benefit of using a function for this simple operation.

src/GlobalErrorHandler.js
rperez89: ohh is a full screen error this one! so yes you are right, so let's change the dispute error inside the dispute component folder maybe?

src/hooks/useDisputes.js
rperez89: i always try not to compute inside the return statement, but if you prefer like this is up to you

src/GlobalErrorHandler.js
fabriziovigevani: I think that since it's a global error and it's not a component that will be a rendered as a child of the main `Disputes` component it may make more sense to leave it in the generic `errors` folder what do you think ?

rperez89: mmm but is not a global error for the whole app, is an error only for disputes so i think that should be in the ddispute folder

fabriziovigevani: Yeah okay that's also true.

src/components/Disputes/DisputeNotFoundError.js
```diff
@@ -0,0 +1,48 @@
...
+      </h1>
+      <div
+        css={`
+          margin-bottom: ${3 * GU}px;
+          text-align: center;
+          color: ${theme.surfaceContentSecondary};
+          ${textStyle('body2')};
+        `}
+      >
+        It looks like there’s no dispute associated with the “DisputeID #
```
sohkai: Let's remove the `the` here

src/hooks/useDisputes.js
```diff
@@ -55,10 +55,35 @@ export default function useDisputes() {
...
+  const disputeProcessed = useProcessedDispute(dispute)
+
+  const disputePhase = getPhaseAndTransition(dispute, courtConfig, now)
+  const disputePhaseKey = disputePhase
+    ? convertToString(Object.values(disputePhase)[0])
+    : ''
+
+  return useMemo(() => {
+    if (fetching || (dispute && !disputeProcessed)) {
+      return { fetching: true }
```
sohkai: Not sure how the rest of the project's hooks have been standardized on, but in this case, it might be nice to return an array instead, rather than two different type schemas.

So for example:

```
return [null, true] // fetching

return [{ ... }, false] // has information
```

fabriziovigevani: When the hook returns lots of different data we usually just use an object.

In this case i think your suggestion makes sense.

> But if all the other code is structured like how this hook is currently written, with an implicitly false fetching key when data exists

It's not, we can change it without extra effort

PR #199, opened 2020-02-20
Tasks: due date fix

### LINKED ISSUE #181: Tasks: due date says "today" even though it's tomorrow
E.g. if you go to https://court.aragon.org/dashboard or https://court.aragon.org/tasks today, you'll see:

Screen Shot 2020-02-18 at 3 18 05 PM

There's more than one day remaining so it's a bit odd it says "today". Given that "today" is also very subjective, I wonder if it's better to say "in the next day" until we get closer, e.g. less than 12 hours (or we use the browser's local time zone to check if the end date is actually in the same day as the person)?

------

Spec:

- If end date is in the same day (as the browser's time zone), say "today"
- If end date is the next day, say "in the next day"
- Otherwise, say "in X days"

rperez89:
this closes #181

### DIFF
--- a/src/components/Tasks/TaskStatus.js
+++ b/src/components/Tasks/TaskStatus.js
@@ -3,7 +3,7 @@ import { textStyle, GU } from '@aragon/ui'
 import dayjs from 'dayjs'
 import { datesDiff } from '../../utils/date-utils'
 
-function getEndDay(dueDate) {
+function getDueDayText(dueDate) {
   const diff = datesDiff(dayjs(), dayjs(dueDate))
 
   if (diff === 0) {
@@ -25,7 +25,7 @@ const TaskStatus = React.memo(function TaskStatus({ dueDate }) {
         margin-left: ${GU * 0.7}px;
       `}
       >
-        {`Open: ${getEndDay(dueDate)}`}
+        {`Open: ${getDueDayText(dueDate)}`}
       </span>
     </div>
   )
--- a/src/components/Tasks/TaskStatus.js
+++ b/src/components/Tasks/TaskStatus.js
@@ -1,10 +1,9 @@
 import React, { useEffect, useState } from 'react'
 import { textStyle, GU } from '@aragon/ui'
-import dayjs from 'dayjs'
-import { datesDiff } from '../../utils/date-utils'
+import dayjs from '../../lib/dayjs'
 
 function getDueDayText(dueDate) {
-  const diff = datesDiff(dayjs(), dayjs(dueDate))
+  const diff = dayjs(dueDate).diff(dayjs(), 'day')
 
   if (diff === 0) {
     return ' Due Today'
@@ -19,7 +18,11 @@ const TaskStatus = React.memo(function TaskStatus({ dueDate }) {
   const [refresh, setRefresh] = useState(false)
 
   const now = new Date()
-  const refreshTime = dueDate - now
+
+  const endofTime = dayjs()
+    .endOf('day')
+    .add(1, 'millisecond')
+  const refreshTime = endofTime - now
 
   useEffect(() => {
     if (refreshTime <= 0) {
--- a/src/components/Tasks/TaskStatus.js
+++ b/src/components/Tasks/TaskStatus.js
@@ -10,9 +10,9 @@ function getEndDay(dueDate) {
     return ' Due Today'
   }
   if (diff === 1) {
-    return 'In the next day'
+    return 'Due In the next day'
   }
-  return `In ${diff} days`
+  return `Due In ${diff} days`
 }
 
 const TaskStatus = React.memo(function TaskStatus({ dueDate }) {
--- a/src/components/Tasks/TaskStatus.js
+++ b/src/components/Tasks/TaskStatus.js
@@ -10,7 +10,7 @@ function getDueDayText(dueDate) {
     return ' Due Today'
   }
   if (diff === 1) {
-    return 'Due In the next day'
+    return 'Due in the next day'
   }
   return `Due In ${diff} days`
 }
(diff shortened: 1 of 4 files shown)


src/components/Tasks/TaskStatus.js
```diff
@@ -1,7 +1,21 @@
...
+    return ' Due Today'
+  }
+  if (diff === 1) {
+    return 'Due In the next day'
+  }
+  return `Due In ${diff} days`
+}
+
+const TaskStatus = React.memo(function TaskStatus({ dueDate }) {
   return (
```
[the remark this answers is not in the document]
rperez89: Not sure if we should use the `useNow` here since we are not showing a timer just a label, and i don't think the user will be looking this screen for hours. it's more like the user will enter to the task list take a look and go to complete some takss cc @bpierre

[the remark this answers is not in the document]
rperez89: pinging @bpierre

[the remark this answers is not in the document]
bpierre: Yes I think we could set a timeout for the next refresh instead, that would re-render in `startOfNextDay - now` every time?

src/components/Tasks/TaskStatus.js
```diff
@@ -16,6 +16,22 @@ function getDueDayText(dueDate) {
 }
 
 const TaskStatus = React.memo(function TaskStatus({ dueDate }) {
+  const [refresh, setRefresh] = useState(false)
+
+  const now = new Date()
+  const refreshTime = dueDate - now
```
fabriziovigevani: Since the refreshing is relative to the remaining days we should calculate this as 

`const refreshTime = (dueDate - now) % DAY_IN_SECONDS` where `DAY_IN_SECONDS = 86400`

rperez89: I think that the timeout should be until local `00:00 hs` and then that number is updated

fabriziovigevani: Yes totally true 👍

src/components/Tasks/TaskStatus.js
```diff
@@ -1,7 +1,37 @@
-import React from 'react'
+import React, { useEffect, useState } from 'react'
 import { textStyle, GU } from '@aragon/ui'
+import dayjs from 'dayjs'
```
sohkai: Wondering, should we make sure to always import `dayjs` from `lib/dayjs`? I assume only that import has the extended functionality?

src/components/Tasks/TaskStatus.js
```diff
@@ -1,7 +1,37 @@
...
+import { datesDiff } from '../../utils/date-utils'
+
+function getDueDayText(dueDate) {
+  const diff = datesDiff(dayjs(), dayjs(dueDate))
+
+  if (diff === 0) {
+    return ' Due Today'
+  }
+  if (diff === 1) {
+    return 'Due In the next day'
```
sohkai:
```suggestion
    return 'Due in the next day'
```

src/components/Tasks/TaskStatus.js
```diff
@@ -1,7 +1,37 @@
...
+  }
+  return `Due In ${diff} days`
```
sohkai:
```suggestion
  return `Due in ${diff} days`
```

src/utils/date-utils.js
```diff
@@ -8,3 +8,24 @@ const KNOWN_FORMATS = {
 export function dateFormat(date, format) {
   return dayjs(date).format(KNOWN_FORMATS[format] || format)
 }
+
+export function datesDiff(startDate, endDate) {
```
sohkai: There's a somewhat easier way to get this:

```
if (endDate.diff(startDate, 'hour', true) < 24) {
  return endDate.get('date') === startDate.get('date') ? 0 : 1
}
return endDate.diff(startDate, 'day')
```

src/components/Tasks/TaskStatus.js
rperez89: yes totally

src/utils/date-utils.js
rperez89: but first we check that the startDate is the same than the endDate, if that condition is false then we check if the hours diff is less than 24 hours then is the next day and if non of the others is true we are returning the days diff

sohkai: The main reason I like the simplification is that this check:

> first we check that the startDate is the same than the endDate

is made redundant by `endDate.diff(startDate, 'hour', true) < 24`. If this is true, it already encompasses the year and month matching, and so we only need to check the day.

sohkai: Actually we've found [`isSame()`](https://day.js.org/docs/en/query/is-same) and should use that :smile:

src/components/Tasks/TaskStatus.js
```diff
@@ -19,7 +18,11 @@ const TaskStatus = React.memo(function TaskStatus({ dueDate }) {
   const [refresh, setRefresh] = useState(false)
 
   const now = new Date()
-  const refreshTime = dueDate - now
+
+  const endofTime = dayjs()
+    .endOf('day')
+    .add(1, 'millisecond')
+  const refreshTime = endofTime - now
```
sohkai: Not sure on this; we might want to use `diff(now, 'millisecond')` or convert the dayjs variable into an actual Date (it might implicitly do this, but it feels better to know we're using two native Date objects than one dayjs object and one Date object).

It might also be clearer as `const startOfNextDay = dayjs().add(1, 'day').startOf('day')`

src/components/Tasks/TaskStatus.js
```diff
@@ -11,10 +44,10 @@ function TaskStatus() {
         margin-left: ${GU * 0.7}px;
       `}
       >
-        Open: Due today
+        {`Open: ${getDueDayText(dueDate)}`}
```
sohkai: I think we should be able to do this:

```suggestion
        Open: {getDueDayText(dueDate)}
```

PR #203, opened 2020-02-25
Global error

rperez89:


src/components/GenericError.js
```diff
@@ -0,0 +1,139 @@
...
+  ButtonBase,
+  Link,
+  GU,
+  IconDown,
+  RADIUS,
+  textStyle,
+  useTheme,
+} from '@aragon/ui'
+
+const SUPPORT_URL = 'https://github.com/aragon/aragon/issues/new'
```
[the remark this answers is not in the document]
rperez89: totally haha

src/components/GlobalErrorScreen.js
```diff
@@ -3,19 +3,34 @@ import PropTypes from 'prop-types'
 import { Card, GU, useTheme, useViewport } from '@aragon/ui'
 import globalError from '../assets/globalError.svg'
 import backgroundError from '../assets/backgroundError.svg'
+import logo from '../assets/logo.svg'
 
 function GlobalErrorScreen({ children }) {
+  const theme = useTheme()
   return (
     <div
+      id="testt2"
```
fabriziovigevani: Should we remove these ids?

src/components/GlobalErrorScreen.js
```diff
@@ -3,19 +3,34 @@ import PropTypes from 'prop-types'
...
+        border-top-color: ${theme.accent};
       `}
     >
+      <img
+        css={`
+          position: absolute;
+          top: 16px;
+          left: 16px;
+        `}
+        src={logo}
```
fabriziovigevani:
```suggestion
        src={logo}
        alt=""
```

rperez89: why is that needed?

fabriziovigevani: It's usually needed for alternative text when the image can't load or something. If we use a decorative image like a logo we should use `alt=""`

src/GlobalErrorHandler.js
```diff
@@ -0,0 +1,70 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import * as Sentry from '@sentry/browser'
+import GenericError from './components/GenericError'
+// import { DAONotFound } from './errors'
```
sohkai: This looks like it could be removed :)

src/GlobalErrorHandler.js
```diff
@@ -0,0 +1,70 @@
...
+        .replace(/^\n+|\n+$/g, '')
+        .replace(/^ {4}/gm, ''),
+    })
+
+    // Once this point is reached, the app can not recover because the routing
+    // system, being below this component in the tree, is not functional
+    // anymore. To make hash changes work despite this (e.g. by pressing the
+    // back button in the browser), the page need to be reloaded.
+    window.removeEventListener('hashchange', this.handleHashchange)
+    window.addEventListener('hashchange', this.handleHashchange)
```
sohkai: Not sure if this is still relevant here, since we're not using hashes for routing here.

src/GlobalErrorHandler.js
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener('hashchange', this.handleHashchange)
+  }
+
+  handleReportClick = () => {
```
sohkai: Let's just always report new errors caught here to sentry, and use the button to show the report dialog.

src/components/GenericError.js
```diff
@@ -0,0 +1,139 @@
...
+
+const GenericError = React.memo(function GenericError({
+  detailsTitle,
+  detailsContent,
+  reportCallback,
+}) {
+  const theme = useTheme()
+  const [opened, setOpened] = useState(false)
+  const toggle = useCallback(() => {
+    setOpened(!opened)
```
sohkai: These can usually be streamlined to using the functional setter, which then allows you to never re-create this function:

```
const toggle = useCallback(() => {
  setOpened(opened => !opened)
}, [])
```

src/components/GlobalErrorScreen.js
```diff
@@ -0,0 +1,110 @@
...
+        border-top-color: ${theme.accent};
+      `}
+    >
+      <img
+        css={`
+          position: absolute;
+          top: 16px;
+          left: 16px;
+        `}
+        src={logo}
```
sohkai: Might be nice to turn this into a link that also does the refresh / go back.

PR #204, opened 2020-02-27
Rewards module: Court subscriptions

fabriziovigevani:
Includes juror's subscription fees.

Subscriptions fees are fetched directly from the contract.
As opposed to dispute fees, when claiming subscription fees, these go directly to the juror's wallet


src/components/Dashboard/RewardsModule.js
```diff
@@ -15,46 +15,74 @@ import NoRewards from './NoRewards'
...
+    totalAppealFees,
+    totalSubscriptionFees,
+  } = useTotalFeeRewards(
+    feeRewards?.arbitrableFees,
+    feeRewards?.appealFees,
+    subscriptionFees
   )
 
+  // We'll get the total juror's balance held in the treasury
   // TODO: Handle possible multiple tokens (fee token can change)
```
rperez89: is this TODO still valid?

src/components/Dashboard/RewardsModule.js
```diff
@@ -75,7 +103,7 @@ const RewardsModule = React.memo(function RewardsModule({
       }
 
       // Claim all appeal fee rewards
-      for (const appealFee of rewards.appealFees) {
+      for (const appealFee of feeRewards.appealFees) {
```
rperez89: cool, didn't know this way to do the fors haha

src/components/Dashboard/RewardsModule.js
```diff
@@ -108,33 +146,38 @@ const RewardsModule = React.memo(function RewardsModule({
         if (loading) return <Loading height={150} />
```
rperez89: let's try to use braces (brett told me in another PR)

src/components/Dashboard/RewardsModule.js
```diff
@@ -108,33 +146,38 @@ const RewardsModule = React.memo(function RewardsModule({
...
 
         if (!hasRewardsToClaim) return <NoRewards />
```
rperez89: same here

src/hooks/useCourtContracts.js
```diff
@@ -258,6 +259,54 @@ export function useRewardActions() {
...
+  const hasJurorClaimed = useCallback(
+    (juror, periodId) => {
+      return courtSubscriptionsContract.hasJurorClaimed(juror, periodId)
+    },
+    [courtSubscriptionsContract]
+  )
+
+  return {
+    claimFees,
+    getters: useMemo(
```
rperez89: should we move this useMemo outside the return? so it's cleaner

src/hooks/useJurorSubscriptionFees.js
```diff
@@ -0,0 +1,44 @@
...
+
+        const jurorSubscriptionsFees = []
+        for (let periodId = 0; periodId < currentPeriodId; periodId++) {
+          const jurorShare = await getters.getJurorShare(
+            wallet.account,
+            periodId
+          )
+
+          if (
+            jurorShare[1].gt(0) &&
```
rperez89: maybe add some comments here explaining why we are getting the jurorsShare at `[1]`?

src/components/Dashboard/RewardsModule.js
fabriziovigevani: Yes it is, this would be the case were we have multiple fee tokens

src/hooks/useJurorSubscriptionFees.js
fabriziovigevani: Yes good idea

src/hooks/useJurorSubscriptionFees.js
```diff
@@ -0,0 +1,45 @@
...
+    const fetchSubscriptionFees = async () => {
+      if (!getters) {
+        return
+      }
+
+      try {
+        const currentPeriodId = await getters.getCurrentPeriodId()
+
+        const jurorSubscriptionsFees = []
+        for (let periodId = 0; periodId < currentPeriodId; periodId++) {
```
sohkai: Just wondering if this is going to be _really_ expensive after a while, if the periodId is unboundedly increasing.

If it is, this sounds like perhaps something for us to cache in the subgraph instead? (E.g. keeping track of which periods are relevant for a juror)

src/hooks/useJurorSubscriptionFees.js
```diff
@@ -0,0 +1,45 @@
...
+          }
+        }
+
+        setSubscriptionFees(jurorSubscriptionsFees)
+      } catch (err) {
+        console.error(`Error fethcing juror subscription fees: ${err}`)
+      }
+    }
+
+    fetchSubscriptionFees()
```
sohkai: As mentioned offline, we should make sure to clean up and cancel these async actions in a `useEffect()`, to make sure there aren't any funny race conditions when setting state.

src/components/Dashboard/RewardsModule.js
```diff
@@ -15,46 +15,74 @@ import NoRewards from './NoRewards'
...
 
-  return { totalArbitrableFees, totalAppealFees }
+    return { totalArbitrableFees, totalAppealFees, totalSubscriptionFees }
+  }, [appealFees, arbitrableFees, subscriptionFees])
 }
 
-// Ruling Fees => ANJ fees
-// Dispute fees + Appeal Fees => DAI fees
+// anjRewards => ANJ => Settled onSettleReward
+// feeRewards => DAI => settled onSettleRewards and onSettleAppealDeposit
```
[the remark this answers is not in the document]
fabriziovigevani: Correct, you need to invoke both of them in case you have regular dispute fee rewards and  also appeal fee rewards.

src/components/Dashboard/RewardsModule.js
```diff
@@ -105,36 +143,46 @@ const RewardsModule = React.memo(function RewardsModule({
...
+                  />
+                )}
+                {totalSubscriptionFees.gt(0) && (
+                  <SubscriptionFeeRewards totalFees={totalSubscriptionFees} />
+                )}
+                <TotalFees
+                  totalFees={totalFeeRewards}
+                  requiresMultipleTxs={
+                    totalDisputesFees.gt(0) ||
+                    (subscriptionFees.lenght > 0 && treasuryBalance.gt(0))
```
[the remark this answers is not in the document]
fabriziovigevani: Yep true

src/components/Dashboard/RewardsModule.js
```diff
@@ -15,46 +15,74 @@ import NoRewards from './NoRewards'
...
   )
   const treasuryBalance = treasuryToken ? treasuryToken.balance : bigNum(0)
+
+  // Total dispute fees are conformed by appeal fees and arbitrable fees (fees paid by the creator of the dispute for the first round
+  // and fees paid by appealers for subsequent rounds)
   const totalDisputesFees = totalArbitrableFees.add(totalAppealFees)
-  const totalFees = totalDisputesFees.add(treasuryBalance)
+
+  // All dispute fees are sent to the treasury after being settled
+  const totalTreasuryFees = totalDisputesFees.add(treasuryBalance)
```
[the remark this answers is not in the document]
fabriziovigevani: Yes that's correct. Since we are triggering every tx right after the other without waiting for them to be mined we must assume that all of them will succeed and try to withdraw the total amount. 

In the case that one of the settlement tx fails probably withdrawing will also fail. I think it's not too big of a deal since they can try again. 

And when we have implemented the signer experience we will be able to handle this scenarios accordingly.

src/hooks/useCourtContracts.js
```diff
@@ -384,3 +435,5 @@ export function useTotalActiveBalancePolling(termId) {
 
   return totalActiveBalance
 }
+
+export function useJurorSubscriptionShare(juror, periodId) {}
```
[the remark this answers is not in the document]
fabriziovigevani: Should be removed.

src/hooks/useJurorSubscriptionFees.js
fabriziovigevani: As discussed offline with @facuspagnuolo, we'll be able to keep track of period Ids where jurors have already claimed so we can skip the `hasJurorClaimed` contract call, see https://github.com/aragon/court-subgraph/issues/48.

It would also be useful to have available in the subgraph the `periodDuration` configuration so we can re-calculate every time there's a new period.

The only contract call that we cannot bypass is the `getJurorShare` call since there's no way to save this information in the subgraph for each juror.

src/components/Dashboard/RewardsModule.js
```diff
@@ -85,18 +113,28 @@ const RewardsModule = React.memo(function RewardsModule({
...
+      // Claim subscription fees
+      for (const subscriptionFee of subscriptionFees) {
+        rewardTransactionQueue.push(
+          await onClaimSubscriptionFees(subscriptionFee.periodId)
+        )
+      }
 
       await Promise.all(rewardTransactionQueue.map(tx => tx.wait()))
+
+      setSubscriptionFees([])
```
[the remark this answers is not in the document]
fabriziovigevani: Yes this is definitely a possible scenario. 

>  But once their transactions were mined, and the subgraph updates, we would be able to start from n+1 (in the example, the third transaction)?

And yes this is what would happen in the case that we have rewards that are not subscriptions.

For court subscriptions, since we are fetching the data once on page load for now, in the scenario described above the page would need to be reloaded again. Once we have https://github.com/aragon/court-dashboard/pull/204#discussion_r395118931 implemented this shouldn't be a problem.

src/components/Dashboard/RewardsModule.js
```diff
@@ -88,28 +116,42 @@ const RewardsModule = React.memo(function RewardsModule({
...
+            await onClaimSubscriptionFees(subscriptionFee.periodId)
+          )
+        }
 
         await Promise.all(rewardTransactionQueue.map(tx => tx.wait()))
+
+        setSubscriptionFees([])
       } catch (err) {
         console.error(`Error claiming rewards: ${err}`)
+        Sentry.captureException(err)
```
sohkai: Just wondering; do we need to guard this with a check that Sentry is actually initialized?

fabriziovigevani: I don't think we need to, just tried as well capturing an error without initializing, with no issues

PR #212, opened 2020-02-28
Sidepanel: Close after tx signed

### LINKED ISSUE #211: Close panel after user signs transaction
See https://www.notion.so/9e7e5cd565b44800a79b5640a5f094a8?v=d3b5e76c06634dafa4b7083edbaca881&p=2da0ba795edc4c4290605d573ff0f31d

fabriziovigevani:
closes #211 

Once we have an implementation of the `Signerpanel` we can show proper messages in case the tx failed

### DIFF
--- a/src/components/Disputes/panels/AppealPanel.js
+++ b/src/components/Disputes/panels/AppealPanel.js
@@ -112,8 +112,9 @@ const AppealPanel = React.memo(function AppealPanel({
         dispute.lastRoundId,
         appealOption.outcome
       )
-      await tx.wait()
+
       onDone()
+      await tx.wait()
     } catch (err) {
       console.log('Error submitting tx: ', err)
     }
--- a/src/components/Disputes/panels/AppealPanel.js
+++ b/src/components/Disputes/panels/AppealPanel.js
@@ -116,7 +116,7 @@ const AppealPanel = React.memo(function AppealPanel({
       onDone()
       await tx.wait()
     } catch (err) {
-      console.log('Error submitting tx: ', err)
+      console.error('Error submitting tx: ', err)
     }
   }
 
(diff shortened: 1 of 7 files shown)


src/components/Disputes/panels/AppealPanel.js
```diff
@@ -112,8 +112,9 @@ const AppealPanel = React.memo(function AppealPanel({
         dispute.lastRoundId,
         appealOption.outcome
       )
-      await tx.wait()
+
```
rperez89: are we adding those blank lines between for some reason?

src/components/Disputes/panels/RevealPanel.js
```diff
@@ -39,9 +39,10 @@ const RevealPanel = React.memo(function RevealPanel({
         jurorDraft.commitment,
         password
       )
+
```
rperez89: same question here

src/components/Disputes/panels/AppealPanel.js
fabriziovigevani: Ah yes, i added them for better readability

src/components/Disputes/panels/AppealPanel.js
```diff
@@ -112,8 +112,9 @@ const AppealPanel = React.memo(function AppealPanel({
...
       onDone()
+      await tx.wait()
     } catch (err) {
       console.log('Error submitting tx: ', err)
```
rperez89: should we use `console.error` for errors?

fabriziovigevani: Yes, i think we should. Initially this was kind of a placeholder to catch the error thinking that in the future we would be handling this in the signer panel.

rperez89: cool, so in the meanwhile we can change those `console.log` for `console.error` right?

PR #218, opened 2020-03-03
Use useCallback for form submissions

fabriziovigevani:
Memoizing all form submission functions  see https://github.com/aragon/court-dashboard/pull/177#discussion_r387005036

### DIFF
--- a/src/components/Disputes/panels/AppealPanel.js
+++ b/src/components/Disputes/panels/AppealPanel.js
@@ -23,11 +23,11 @@ const AppealPanel = React.memo(function AppealPanel({
   onDone,
 }) {
   const { feeToken } = useCourtConfig()
+  const { account: connectedAccount } = useWallet()
   const [selectedOutcome, setSelectedOutcome] = useState({
     value: -1,
     error: null,
   })
-  const { account: connectedAccount } = useWallet()
 
   const handleOutcomeSelected = useCallback(newOutcome => {
     setSelectedOutcome({ value: newOutcome })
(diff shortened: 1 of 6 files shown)


src/components/Disputes/panels/AppealPanel.js
```diff
@@ -205,4 +215,19 @@ const AppealPanel = React.memo(function AppealPanel({
   )
 })
 
-export default AppealPanel
+export default ({ dispute, confirm, ...props }) => {
```
sohkai: Might want to move the `useMemo` to here and give this a component a name

src/components/Disputes/panels/AppealPanel.js
```diff
@@ -58,66 +58,76 @@ const AppealPanel = React.memo(function AppealPanel({
...
-      return true
-    }
-
-    return false
-  }
-
   // For submission
-  const handleAppeal = async event => {
-    try {
+  const handleAppeal = useCallback(
```
sohkai: Nitpick: might be nice to move the other `useCallback()` function (`handleOutcomeSelected()`) close to this function; in other places we've taken the habit of grouping "alike" hooks together, since they usually do similar things:

1. `useState()` / custom state hooks: these provide state and "context" for logical operations later in the component
1. `useMemo()`, `useRef()`: these are usually derived from props or state, and continue providing state and context
1. `useCallback()`: these are usually reactive handlers passed into props
1. `useEffect()`: these usually derive from the above, and can be viewed as "automatic triggers" (similar in concept to render, which is always last; it _reacts_ based on all the above changing).

src/components/Disputes/panels/AppealPanel.js
```diff
@@ -215,19 +215,17 @@ const AppealPanel = React.memo(function AppealPanel({
...
 
-  return (
-    <AppealPanel
-      dispute={dispute}
-      appeal={appeal}
-      confirm={confirm}
-      {...props}
-    />
-  )
+  return useMemo(() => {
```
sohkai: Ah interesting, I meant `React.memo()` not `useMemo()` previously, but I've never considered using a `useMemo()` for the rendered component before.

Although in this case, I think it'd be functionally equivalent to use `React.memo()` (and it'd be a bit cleaner).

PR #219, opened 2020-03-03
Uniswap price info

### LINKED ISSUE #119: Use Uniswap for price information
Our current API is centralized, unlikely to list ANJ, and has rate limiting :(.

fabriziovigevani:
closes #119


src/hooks/useANJBalanceToUsd.js
```diff
@@ -0,0 +1,101 @@
...
+async function getANJTradeDetails(tokenAddress, anjTokenAddress, tradeAmount) {
+  const anjMarketDetails = await getANJMarketDetails(
+    tokenAddress,
+    anjTokenAddress
+  )
+
+  return getTradeDetails(TRADE_EXACT.OUTPUT, tradeAmount, anjMarketDetails)
+}
+
+export default function useANJBalanceToUsd(amount) {
```
sohkai: Ah, so there's generally some confusion over how Uniswap and its SDK work in this regard.

This hook is getting the converted **execution** amount if a real trade happened right now with ANJ<>DAI on the given amount (`getTradeDetails()`).

Since we're not interested in doing the actual conversion, we don't care about slippage (or liquidity, or any of the other Uniswap concerns)—we only care about the **spot price**. That can be obtained via just `getMarketDetails()`, and its `.marketRate`. We then multiply the spot price with the ANJ amount given to get the converted value.

src/utils/known-tokens.js
```diff
@@ -0,0 +1,10 @@
+import env from '../environment'
+
+export const KNOWN_DAI_ADDRESS_BY_ENV = new Map([
+  ['1', '0x6b175474e89094c44da98b954eedeac495271d0f'],
+  ['4', '0xe9A083D88Eed757B1d633321Ce0519F432c6284d'],
```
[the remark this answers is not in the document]
fabriziovigevani: You meant Rinkeby no?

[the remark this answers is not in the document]
sohkai: Ah I always thought `4` was Kovan; nevermind it is Rinkeby!

src/hooks/useANJBalanceToUsd.js
```diff
@@ -0,0 +1,88 @@
...
+
+        const convertedAmount = formatUnits(amount.mul(rate).div(precision), {
+          digits: anjToken.decimals,
+        })
+
+        if (!cancelled) {
+          setConvertedAmount(convertedAmount)
+        }
+      } catch (err) {
+        console.log('err', err)
```
sohkai: Might as well use an `console.error()` with something more descriptive, e.g. `Could not fetch Uniswap price for ANJ`.

src/hooks/useANJBalanceToUsd.js
```diff
@@ -0,0 +1,88 @@
...
+    [tokenAddress, anjTokenAddress].map(address => {
+      // In the case of ETH, undefined should be passed as the reserves data.
+      if (addressesEqual(address, ETH_FAKE_ADDRESS)) {
+        return undefined
+      }
+
+      const tokenReserves = getTokenReserves(address, env('CHAIN_ID'))
+
+      if (!tokenReserves) {
+        throw new Error('Could not fetch reserves')
```
rperez89: Not sure if here should throw an error that will be catched by the `GlobalError` component, i think that we should not  show a global error if we just couldn't fetch from uniswap

src/hooks/useANJBalanceToUsd.js
```diff
@@ -0,0 +1,88 @@
...
+      }
+      return tokenReserves
+    })
+  )
+
+  return getMarketDetails(tokenData, anjData)
+}
+
+export default function useANJBalanceToUsd(amount) {
```
rperez89: What about doing this a generic hook like `useTokenBalanceToUsd` since we are going to use it also for ANT and DAI?

src/utils/known-tokens.js
```diff
@@ -0,0 +1,10 @@
+import env from '../environment'
+import { getNetworkName } from '../lib/web3-utils'
+
+export const KNOWN_DAI_ADDRESS_BY_ENV = {
```
rperez89: What do you think about making this generic? so `Known_tokens_By_Env` and not just with the address maybe also with the token id, decimals and symbol ?

src/utils/known-tokens.js
```diff
@@ -0,0 +1,10 @@
...
+  mainnet: '0x6b175474e89094c44da98b954eedeac495271d0f',
+}
+
+export function getDaiAddress() {
```
rperez89: And also make this generic

src/hooks/useANJBalanceToUsd.js
fabriziovigevani: In this case it would be catched by the `try {} catch() {}` in `useANJBalanceToUsd `

src/hooks/useANJBalanceToUsd.js
fabriziovigevani: Yeah true i completely forgot about that

src/utils/known-tokens.js
fabriziovigevani: Yep, we can make it generic and add the `ANT` as well 👍

PR #226, opened 2020-03-05
Use cryptocompare api for non ANJ tokens

fabriziovigevani:
See https://github.com/aragon/court-dashboard/pull/219#discussion_r387924367

For other tokens we can directly use the `cryptocompare` api price info.
Also made `known-tokens` generic.

### DIFF
--- a/src/hooks/useTokenBalanceToUsd.js
+++ b/src/hooks/useTokenBalanceToUsd.js
@@ -3,12 +3,13 @@ import { getMarketDetails, getTokenReserves } from '@uniswap/sdk'
 import { useCourtConfig } from '../providers/CourtConfig'
 
 import env from '../environment'
-import { getDaiAddress } from '../utils/known-tokens'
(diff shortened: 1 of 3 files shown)


src/hooks/useTokenBalanceToUsd.js
```diff
@@ -86,3 +93,42 @@ export default function useANJBalanceToUsd(amount) {
...
+
+/**
+ * Convert a token into a USD price
+ *
+ * @param {String} symbol The symbol of the token to convert from.
+ * @param {Number} decimals The amount of decimals for the token.
+ * @param {BigNumber} balance The balance to convert into USD.
+ * @returns { Number } The balance value in USD
+ */
+export default function useTokenBalanceToUsd(symbol, decimals, balance) {
```
sohkai: Hmm do we use this? Do we show any other tokens in the dashboard?

rperez89: yes ANT and DAI

PR #227, opened 2020-03-05
Clock module

fabriziovigevani:
**This PR is subject to change depending on the designs**

Includes court module and `heartbeat` interaction.

Made some components used by the `account module` more generic so we could reuse them with the `clock module`.

Will integrate this logic with the balance module to tell if a balance should be updated on a separate PR

**Preview:**

Screen Shot 2020-03-06 at 3 04 57 PM

Screen Shot 2020-03-06 at 3 05 10 PM


src/components/ClockModule.js
```diff
@@ -0,0 +1,281 @@
...
+    setOpened(false)
+  }, [])
+
+  const handleOnClick = useCallback(async () => {
+    try {
+      const tx = await onHeartbeat(neededTransitions)
+      await tx.wait()
+    } catch (err) {
+      console.error(err)
+    }
```
bpierre: Note to self: remember to use [`addActivity()`](https://github.com/aragon/court-dashboard/blob/activity/src/components/Activity/ActivityProvider.js#L109) here, whenever #100 or this PR gets merged.

src/components/ClockModule.js
```diff
@@ -0,0 +1,281 @@
...
+              width={4.5 * GU}
+              height={4.5 * GU}
+            />
+          </div>
+        }
+        content={
+          <div
+            css={`
+              line-height: 1;
+              width: 120px;
```
bpierre:
```suggestion
              width: ${15 * GU}px;
```

src/components/ClockModule.js
```diff
@@ -0,0 +1,281 @@
...
+
+              & > time {
+                & > span:first-child {
+                  display: none;
+                }
```
bpierre: This should probably be added as a mode into `Timer`, as the internal HTML structure could break at any time.

src/components/ClockModule.js
```diff
@@ -0,0 +1,281 @@
...
+                & * {
+                  line-height: 1;
+                }
```
bpierre: Same here, ideally we should either add a mode on the component or try to find another way to achieve this.

src/components/ClockModule.js
```diff
@@ -0,0 +1,281 @@
...
+                  `}
+                >
+                  {isSynced ? 'Court term synced' : 'Term out of sync'}
+                </span>
+              </>
+            ) : (
+              <div
+                css={`
+                  background: ${theme.background};
+                  height: 40px;
```
bpierre:
```suggestion
                  height: ${5 * GU}px;
```

src/components/ClockModule.js
```diff
@@ -0,0 +1,281 @@
...
+                      ? theme.contentSecondary
+                      : theme.negative};
+                    line-height: 1;
+                  `}
+                >
+                  {isSynced ? 'Court term synced' : 'Term out of sync'}
+                </span>
+              </>
+            ) : (
+              <div
```
bpierre: Nice touch! :heart:

src/components/ClockModule.js
```diff
@@ -0,0 +1,281 @@
...
+              />
+            )}
+          </div>
+        }
+        onClick={toggle}
+      />
+
+      <HeaderPopover
+        animateHeight={false}
+        heading="Court cLock"
```
bpierre:
```suggestion
        heading="Court clock"
```

src/components/ClockModule.js
```diff
@@ -0,0 +1,281 @@
...
+          >
+            {isSynced ? (
+              <span
+                css={`
+                  ${textStyle('body3')};
+                  color: ${theme.contentSecondary};
+                `}
+              >
+                Term is the time unit for measuring Aragon Court phases. Each
+                term lasts {formatDuration(courtConfig.termDuration / 1000)}
```
bpierre:
```suggestion
                term lasts {formatDuration(courtConfig.termDuration / 1000)}.
```

src/components/ClockModule.js
```diff
@@ -0,0 +1,281 @@
...
+              </span>
+            ) : (
+              <Button
+                label="Call Heartbeat"
```
bpierre: Not sure if “heartbeat” is a name or a noun here, but if it’s the latter we shouldn’t capitalize it:

```suggestion
                label="Call heartbeat"
```

src/components/Header/Header.js
```diff
@@ -30,13 +30,16 @@ const Header = React.memo(function Header({ compactMode, toggleMenuPanel }) {
       <div
         css={`
           display: flex;
-          flex-grow: 1;
-          justify-content: center;
+          align-self: stretch;
+
+          & > * {
+            align-self: stretch;
+          }
```
bpierre: Is this needed?

src/components/Header/Header.js
```diff
@@ -30,13 +30,16 @@ const Header = React.memo(function Header({ compactMode, toggleMenuPanel }) {
...
         `}
       >
-        {!compactMode && <Clock />}
+        {!compactMode && <ClockModule />}
+        <Account />
```
bpierre: Should we rename this `AccountModule` for consistency?

fabriziovigevani: Yep good idea

src/components/ClockModule.js
fabriziovigevani: I think in this case is intended to be a name? cc. @dizzypaty

src/components/Header/Header.js
fabriziovigevani: Not at all, forgot to remove it when updating the inner components. The `align-self` property on the parent component can be replaced with `height: 100%`

src/components/ClockModule.js
fabriziovigevani: Yeah that's true! I can add the mode in `@aragon/ui` and i'll add a note here to use it when its available :)

fabriziovigevani: PR: https://github.com/aragon/aragon-ui/pull/734

src/components/ClockModule.js
fabriziovigevani: Added in https://github.com/aragon/court-dashboard/pull/227/commits/f1b5e8d2f1ef093f28f39f709b132f82f3a1851a

src/components/ClockModule.js
fabriziovigevani: From an offline chat with @dizzypaty, we are rewording most labels including the `Call heartbeat ` to `Update term`. Here is the commit https://github.com/aragon/court-dashboard/pull/227/commits/e80c35b1e5454cfa489dceaba8fdb4d31a08b08e

src/components/Activity/activity-types.js
```diff
@@ -101,12 +102,12 @@ const ACTIVITY_TYPES = new Map(
         `,
       }
     },
-    revealVote({ roundId, disputeId }) {
+    heartbeat({ transitions }) {
       return {
-        title: 'Reveal vote',
-        icon: iconAnj,
+        title: 'Heartbeat',
```
bpierre: Shouldn’t this be “Update term”, based on https://github.com/aragon/court-dashboard/pull/227#discussion_r390451950?

fabriziovigevani: Yeah totally! Good catch!

src/components/ClockModule.js
```diff
@@ -0,0 +1,286 @@
...
+        outline: 0;
+      `}
+    >
+      <HeaderModule
+        icon={
+          <div
+            css={`
+              background-color: ${theme.background};
+              display: flex;
+              border-radius: 100px;
```
sohkai, line 78: Wondering: are we just trying to turn this into a circle? We should use `border-radius: 50%` instead if so :)

src/components/ClockModule.js
```diff
@@ -0,0 +1,286 @@
...
+          <div
+            css={`
+              line-height: 1;
+              width: ${15 * GU}px;
+
+              & span {
+                line-height: 1;
+              }
+
+              // TODO: Use showIcon new prop when available, see: https://github.com/aragon/aragon-ui/pull/734
```
sohkai, line 99: cc @bpierre have we released aragonUI since this was introduced?

bpierre: Nope, I’m going to update this after the new version gets released. :+1:

PR #229, opened 2020-03-10
Dashboard: stats widget

rperez89:


src/hooks/query-hooks.js
```diff
@@ -49,3 +49,13 @@ export function useFirstANJActivationQuery(jurorId, { pause = false }) {
...
 }
+
+export function useActiveJurorsNumber() {
+  const [result] = useQuery({
+    query: ActiveJurors,
+  })
+
+  const { jurors } = result.data || {}
+
+  return jurors ? jurors.length : null
```
sohkai: Could we just use the null coalescing operator here (`?.`)?

src/hooks/query-hooks.js
```diff
@@ -1,7 +1,7 @@
 import { useQuery } from 'urql'
 
 import { JurorDrafts, JurorDraftRewarded } from '../queries/jurorDrafts'
-import { FirstANJActivationMovement } from '../queries/balances'
+import { FirstANJActivationMovement, ActiveJurors } from '../queries/balances'
```
sohkai: Nitpick: might as well alphabetize :)

src/hooks/useCourtContracts.js
```diff
@@ -442,3 +451,57 @@ export function useTotalActiveBalancePolling(termId) {
 
   return totalActiveBalance
 }
+
+export function useTotalANTStaked() {
```
sohkai: I think this is currently only reporting the **ANJ** staked (`JurorsRegistry` doesn't know anything about ANT).

For mainnet, total **ANT** staked comes from two places: the Network DAO's Agent, and the Network Fundraising's Vault. You can see how [anj.aragon.org](https://github.com/aragon/anj.aragon.org/blob/master/src/web3-contracts.js#L359) reports this figure.

For rinkeby / other environments, I would just leave this value at 0.

src/hooks/useCourtContracts.js
```diff
@@ -442,3 +451,57 @@ export function useTotalActiveBalancePolling(termId) {
...
+        .catch(err => {
+          console.error(`Error fetching ANT staked: ${err}`)
+        })
+    }
+
+    getTotalActiveBalance()
+  }, [jurorRegistryContract])
+
+  return totalActiveBalance
+}
```
sohkai: There are a couple of updates we should do on the changes to this file. This is a good chance for @fabriziovigevani to teach what he's done in #186 :).

One thing I will say though; it would be nice for these stats to be updated every 10-15s :).

src/hooks/useCourtStats.js
```diff
@@ -0,0 +1,48 @@
...
+    token: { ...getKnownToken('ANJ'), icon: IconANJ },
+  },
+  {
+    label: 'Total Staked ANT',
+    token: { ...getKnownToken('ANT'), icon: IconANT },
+  },
+  { label: 'Total Active Jurors' },
+]
+
+function useCourtStats() {
```
sohkai: It would be nice to add a bit of JSDoc to hooks like this, especially for hooks where we're returning tuples with the loading state in the second index.

src/components/Dashboard/CourtStats.js
```diff
@@ -1,20 +1,152 @@
 import React from 'react'
-import { Box, Distribution } from '@aragon/ui'
+import { Box, GU, textStyle, useTheme } from '@aragon/ui'
+import useCourtStats from '../../hooks/useCourtStats'
+import { formatUnits } from '../../lib/math-utils'
+import {
+  useANJBalanceToUsd,
+  useTokenBalanceToUsd,
+} from '../../hooks/useTokenBalanceToUsd'
```
sohkai: Let's put the hook imports together; right now they're separated by the `lib/math-utils` import

src/components/Dashboard/CourtStats.js
```diff
@@ -1,20 +1,152 @@
...
+import Loading from './Loading'
 
-const CourtStats = () => {
+const splitAmount = amount => {
```
sohkai: This looks like it might've been taken from Finance, or another app who didn't get the memo on the aragonDS upgrade.

Wondering if we should:

1. Make this an actual component (e.g. `SplitAmount`)
2. Update the css to use `GU` and `textStyle()`

src/components/Dashboard/CourtStats.js
```diff
@@ -1,20 +1,152 @@
...
+  )
+}
+
+function CourtStats() {
+  const theme = useTheme()
+  const [stats, fetching] = useCourtStats()
+
+  return (
+    <Box heading="Court Metrics" padding={3 * GU}>
+      {(() => {
```
sohkai: In this case, it might be simpler as:

```js
{fetching ? (
  <Loading height={86} />
): (
  stats.map((stat, index) => {
    ...
  })
)}

File: src/components/Dashboard/CourtStats.js [hunk 9]
```diff
@@ -1,20 +1,152 @@
...
+              )}
+            </div>
+          )
+        })
+      })()}
     </Box>
   )
 }
 
+function Stat({ value }) {
```
sohkai: Let's just inline this small component

src/networks.js
```diff
@@ -16,3 +16,20 @@ export const networks = {
...
+
+export const networkConfigs = {
+  rpc: {
+    nodes: {
+      defaultEth: 'http://localhost:8545',
+    },
+  },
+  main: {
+    nodes: {
+      defaultEth: 'wss://mainnet.eth.aragon.network/ws',
```
sohkai: It turns out [Ethers.js didn't have support for websockets until last week](https://github.com/ethers-io/ethers.js/issues/141#issuecomment-598686017), and this is still in its v5 beta.

For now, since we don't need events or need to query that often, let's continue using the http endpoint:

```suggestion
      defaultEth: 'https://mainnet.eth.aragon.network/',
```

src/components/Dashboard/CourtStats.js
```diff
@@ -1,20 +1,152 @@
...
+          </span>
+        </div>
+      </div>
+      <span
+        css={`
+          ${textStyle('body2')}
+          color: ${theme.positive};
+        `}
+      >
+        ${symbol === 'ANJ' ? AnjUsdValue(value) : TokenUsdValue(token, value)}
```
sohkai: On mainnet, neither of the tokens were returning the USD values.

src/components/Activity/assets/activity-icon-execute-ruling.svg
```diff
@@ -1 +1 @@
....15-.528.166-.106.007-.22.005-.35-.015-.033-.004-.064-.009-.099-.016l-.051-.01-.054-.013a1.928 1.928 0 01-.233-.074 1.918 1.918 0 01-.044-.24z" fill="#fff"/><path opacity=".668" d="M16.747 10.343l-3.142 2.76a.624.624 0 01-.879-.054l-.215-.243a.618.618 0 01.013-.833l3.231-2.84a.624.624 0 01.832.092l.215.243a.618.618 0 01-.055.875z" fill="url(#paint3_linear)"/></g><defs><linearGradient id="paint0_linear" x1="7.442" y1="33.717" x2="30.981" y2="15.121" gradientUnits="userSpaceOnUse"><stop stop-color="#FFB36D"/><stop offset="1" stop-color="#F88"/></linearGradient><linearGradient id="paint1_linear" x1="5.585" y1="16.086" x2="7.48" y2="12.153" gradientUnits="userSpaceOnUse"><stop stop-color="#fff" stop-opacity=".01"/><stop offset="1" stop-color="#fff"/></linearGradient><linearGradient id="paint2_linear" x1=".588" y1="9.161" x2="4.652" y2="1.038" gradientUnits="userSpaceOnUse"><stop stop-color="#fff" stop-opacity=".01"/><stop offset="1" stop-color="#fff"/></linearGradient><linearGradient id="paint3_linear" x1="4.792" y1="13.948" x2="8.857" y2="5.825" gradientUnits="userSpaceOnUse"><stop stop-color="#fff" stop-opacity=".01"/><stop offset="1" stop-color="#fff"/></linearGradient></defs></svg>
```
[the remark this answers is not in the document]
rperez89: yes!

src/utils/known-tokens.js
```diff
@@ -2,10 +2,40 @@ import env from '../environment'
 import { getNetworkType } from '../lib/web3-utils'
 
 export const KNOWN_TOKEN_BY_ENV = {
+  ANJ: {
+    mainnet: {
+      address: '0xcD62b1C403fa761BAadFC74C525ce2B51780b184',
+      decimals: 18,
+      symbol: 'ANJ',
+    },
+    rpc: {
```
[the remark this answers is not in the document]
rperez89: oh that was i mistake to test it locally

src/hooks/useTokenBalanceToUsd.js
```diff
@@ -102,7 +102,7 @@ export function useANJBalanceToUsd(amount) {
  * @param {BigNumber} balance The balance to convert into USD.
  * @returns { Number } The balance value in USD
  */
-export default function useTokenBalanceToUsd(symbol, decimals, balance) {
+export function useTokenBalanceToUsd(symbol, decimals, balance) {
```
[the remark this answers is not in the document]
rperez89: yeah, but which one would you make the default here?

src/networks.js
```diff
@@ -16,3 +16,20 @@ export const networks = {
   },
   main: { court: '0xee4650cBe7a2B23701D416f58b41D8B76b617797' },
 }
+
+export const networkConfigs = {
+  rpc: {
```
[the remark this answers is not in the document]
rperez89: do we have one for ropsten?

src/hooks/useCourtContracts.js
rperez89: ohh yes we got confused with this one! thxs!!

src/hooks/useCourtContracts.js
fabriziovigevani: Yes! From https://github.com/aragon/court-dashboard/pull/186#discussion_r395175180 i think we could deprecate `useTotalBalancePolling` (since we are only using it for the account banner which passes the current term) and only use this function implemented by @rperez89. 

You can see how the interval and cancel pattern has been implemented at https://github.com/aragon/court-dashboard/blob/2238e9eeed14604282fb199c431fc49917753dbd/src/hooks/useCourtContracts.js#L516
What we could also do is pass the polling frequency as parameter to the function.

src/components/Dashboard/CourtStats.js
rperez89: for me is way more cleaner to read as it is now, what do yo think?

src/utils/known-tokens.js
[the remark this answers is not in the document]
sohkai: That sounds reasonable too. Are these configs only used for fetching price (would be nice to comment this or make that assumption a bit more obvious)?

src/networks.js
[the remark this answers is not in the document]
sohkai: We do as well expose a ropsten network.

Do we support ropsten elsewhere in the dashboard? If not, let's leave that as an item for the future :).

src/hooks/useTokenBalanceToUsd.js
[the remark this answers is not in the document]
sohkai: Ah, I hadn't realized this had more than one hook!

Perhaps we could rename this file, or put the other hook in a different file. It's a bit weird for the hook of the same file name not be the default export.

src/components/Dashboard/CourtStats.js
sohkai: Let's leave it as is!

src/utils/known-tokens.js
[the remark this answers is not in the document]
rperez89: yes is only for fetching the price!, cool will comment something

src/networks.js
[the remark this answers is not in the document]
rperez89: https://github.com/aragon/court-dashboard/blob/b38ee3c8a2129234bfce59f7869b117999db06ac/src/networks.js#L11 yes we expose ropsten!

src/hooks/useTokenBalanceToUsd.js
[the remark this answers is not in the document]
rperez89: sure!

src/hooks/useCourtContracts.js
```diff
@@ -564,44 +566,56 @@ export function useActiveBalanceOfAt(juror, termId) {
...
     let timeoutId
 
-    const fetchTotalActiveBalance = () => {
+    if (getNetworkType() === 'local') {
+      setError(true)
+      return
+    }
+    if (!antContract) {
+      return
+    }
```
fabriziovigevani: I think the first if is not needed. 

What we can do instead is return in the second condition `!antContract` setting the state of `total ANTStaked` to 0

What do you tihnk?

```suggestion
    if (!antContract) {
      return setState(bigNum(0))
    }
```

src/hooks/useCourtContracts.js
```diff
@@ -564,44 +566,56 @@ export function useActiveBalanceOfAt(juror, termId) {
...
-export function useTotalActiveBalancePolling(termId) {
-  const jurorRegistryContract = useCourtContract(
-    CourtModuleType.JurorsRegistry,
-    jurorRegistryAbi
-  )
-  const [totalActiveBalance, setTotalActiveBalance] = useState(bigNum(-1))
+export function useTotalANTStakedPolling(timeout = 1000) {
+  const [totalANTStaked, setTotalANTStaked] = useState(bigNum(-1))
+  const [error, setError] = useState(false)
+  const [controlledTimeout, setControlledTimeout] = useState(0)
```
fabriziovigevani: Here we could use `useRef(0)` instead and update it as `controlledTimeout.current = timeout`. This way we don't do an unnecessary re-render the first time we get the total staked.

src/components/Dashboard/CourtStats.js
```diff
@@ -1,20 +1,131 @@
...
+                }
+              `}
+            >
+              <span
+                css={`      
+               ${textStyle('body2')}
+               color: ${theme.surfaceContentSecondary};
+               display:block;
+               margin-bottom:${1 * GU}px;
+             `}
```
fabriziovigevani: @bpierre mentioned in previos occasions that we should use a semicolon after using `textStyle` so the linter can recognize it and ident the props properly

src/hooks/useCourtContracts.js
rperez89: on this one i would preffer to let as it is now since is clear that we are not supporting the local environment

src/endpoints.js
```diff
@@ -26,7 +26,8 @@ function getAPIBase() {
 
 export default function endpoints() {
   const [API_BASE_HTTP, API_BASE_WS] = getAPIBase()
-  const networkType = getNetworkType(CHAIN_ID)
+  const networkType =
```
sohkai: Generally the local chain's type is `"private"`.

src/hooks/query-hooks.js
```diff
@@ -51,3 +51,13 @@ export function useFirstANJActivationQuery(jurorId, { pause = false }) {
...
 }
+
+export function useActiveJurorsNumber() {
+  const [{ data, error }] = useQuery({
+    query: ActiveJurors,
+  })
+
+  const { jurors } = data || {}
+
+  return [jurors?.length, error]
```
sohkai: Could we use `data?.jurors?.length`?

src/networks.js
```diff
@@ -10,15 +11,47 @@ export const RINKEBY_STAGING_COURT =
   '0x52180Af656A1923024D1ACcF1D827AB85cE48878'
 
 export const networks = {
-  rpc: { court: '0xD833215cBcc3f914bD1C9ece3EE7BF8B14f841bb' },
```
sohkai: A bit of a nitpick, but let's order these networks as:

1. Mainnet
2. Rinkeby
3. Ropsten
4. Local

This more or less reflects their relative importance to us as environments.

src/networks.js
```diff
@@ -10,15 +11,47 @@ export const RINKEBY_STAGING_COURT =
...
+  local: { court: '0xD833215cBcc3f914bD1C9ece3EE7BF8B14f841bb' },
   ropsten: { court: '0x3b26bc496aebaed5b3E0E81cDE6B582CDe71396e' },
   rinkeby: {
     // Use the 'usability' Court address if declared
```
sohkai: We might need to update this comment now that we're also including the staging environment on rinkeby

src/networks.js
```diff
@@ -10,15 +11,47 @@ export const RINKEBY_STAGING_COURT =
...
+  main: {
+    nodes: {
+      defaultEth: 'https://mainnet.eth.aragon.network/',
+    },
+  },
+  rinkeby: {
+    nodes: {
+      defaultEth: 'https://rinkeby.eth.aragon.network/',
+    },
+  },
```
sohkai: We can also add ropsten: `'https://ropsten.eth.aragon.network/'`

src/web3-contracts.js
```diff
@@ -20,3 +22,19 @@ export function useContract(address, abi, signer = true) {
     )
   }, [abi, account, address, ethers, signer])
 }
+
+export function useContractReadOnly(address, abi) {
+  const ethEndpoint = networkConfigs[getNetworkType()].nodes.defaultEth
+
+  const ethProvider = useMemo(
+    () => (ethEndpoint ? new Providers.JsonRpcProvider(ethEndpoint) : null),
```
sohkai: A note for later: I think for now this is totally fine, since this provider uses http, but when we switch to websockets, we'll most likely want to initialize a singleton provider to maintain a single websocket connection for all contracts.

src/hooks/useCourtStats.js
```diff
@@ -0,0 +1,126 @@
...
+      error,
+    ]
+  }, [error, rewards])
+}
+/**
+ * Hook to get the dashboard stats ANJ active balance, ANT total stake and the active jurors number
+ * @returns {Array} First item an array with the stats and the second one a loading state
+ */
+function useCourtStats() {
+  const timeout = 5000
```
sohkai: Let's lift this out to be a constant at the module level.

It might also be fine up'ing it to 15s, since block times are generally around that length!

src/flagged-disputes/voided-disputes.js
```diff
@@ -8,7 +8,7 @@ import { getNetworkType } from '../lib/web3-utils'
 import env from '../environment'
 
 const VOIDED_DISPUTES = {
-  rpc: new Map([[networks.rpc.court, new Map([])]]),
+  local: new Map([[networks.local.court, new Map([])]]),
```
fabriziovigevani: Could we also update the one at `precedence-campaign-disputes.js` that it was recently merged?

src/components/Dashboard/CourtStats.js
```diff
@@ -1,20 +1,131 @@
...
+        </span>
+        {!error && (
+          <div
+            css={`
+              display: inline-flex;
+            `}
+          >
+            <img
+              css={`
+                margin-right: 5px;
```
sohkai: Should this be 0.5GU?

src/hooks/subscription-hooks.js
```diff
@@ -320,3 +327,30 @@ export function useTasksSubscription() {
...
+    modules,
+    CourtModuleType.JurorsRegistry
+  )
+
+  const [{ data, error }] = useSubscription({
+    query: JurorsRegistryModule,
+    variables: { id: jurorRegistryAddress },
+  })
+
+  const juroRegistryStats = data?.jurorsRegistryModule || null
```
sohkai:
```suggestion
  const jurorRegistryStats = data?.jurorsRegistryModule || null
```

sohkai:
```suggestion
  const jurorRegistryStats = data?.jurorsRegistryModule || null
```

src/hooks/useCourtStats.js
rperez89: sure!

src/web3-contracts.js
rperez89: ok, let's keep that in mind

src/hooks/useCourtContracts.js
```diff
@@ -564,44 +566,57 @@ export function useActiveBalanceOfAt(juror, termId) {
...
+  const antContract = useContractReadOnly(antAddress, tokenAbi)
+
+  const controlledTimeout = useRef(0)
 
   useEffect(() => {
     let cancelled = false
     let timeoutId
 
-    const fetchTotalActiveBalance = () => {
+    if (getNetworkType() === 'local') {
```
sohkai: Wonder if we can leave a comment for this—do we not have the contracts on the local environment?

src/hooks/useCourtContracts.js
```diff
@@ -564,44 +566,57 @@ export function useActiveBalanceOfAt(juror, termId) {
...
-    jurorRegistryAbi
-  )
-  const [totalActiveBalance, setTotalActiveBalance] = useState(bigNum(-1))
+export function useTotalANTStakedPolling(timeout = 1000) {
+  const [totalANTStaked, setTotalANTStaked] = useState(bigNum(-1))
+  const [error, setError] = useState(false)
+  const { address: antAddress } = getKnownToken('ANT') || {}
+  const antContract = useContractReadOnly(antAddress, tokenAbi)
+
+  const controlledTimeout = useRef(0)
```
sohkai: Let's add a comment to say that we start with 0 because we want to immediately make the call

src/hooks/useCourtContracts.js
```diff
@@ -576,12 +576,14 @@ export function useTotalANTStakedPolling(timeout = 1000) {
   const { address: antAddress } = getKnownToken('ANT') || {}
   const antContract = useContractReadOnly(antAddress, tokenAbi)
 
+  // We are starting in 0 in order to inmediately make the fetch call
```
sohkai:
```suggestion
  // We are starting in 0 in order to immediately make the fetch call
```

src/providers/CourtConfig.js
```diff
@@ -1,15 +1,15 @@
...
-import environment from '../environment'
+// import environment from '../environment'
 import { useCourtConfigSubscription } from '../hooks/subscription-hooks'
-import { getNetworkType } from '../lib/web3-utils'
-import { networks } from '../networks'
+// import { getNetworkType } from '../lib/web3-utils'
+import { getNetwork } from '../networks'
 
-const CHAIN_ID = environment('CHAIN_ID')
+// const CHAIN_ID = environment('CHAIN_ID')
```
fabriziovigevani: Let's remove this comments

src/hooks/useCourtContracts.js
```diff
@@ -578,7 +582,7 @@ export function useTotalANTStakedPolling(timeout = 1000) {
     let cancelled = false
     let timeoutId
 
-    if (getNetworkType() === 'local') {
+    if (getInternalNetworkName() === 'local') {
```
fabriziovigevani: Should we use `isLocalOrUnknownNetwork ` ?

src/endpoints.js
```diff
@@ -27,7 +27,7 @@ function getAPIBase() {
 export default function endpoints() {
   const [API_BASE_HTTP, API_BASE_WS] = getAPIBase()
   const networkType =
-    getNetworkType(CHAIN_ID) === 'local' ? 'rpc' : getNetworkType(CHAIN_ID)
+    getNetworkType(CHAIN_ID) === 'private' ? 'rpc' : getNetworkType(CHAIN_ID)
```
fabriziovigevani: Same here ? `isLocalOrUnknownNetwork `

PR #238, opened 2020-03-17
settings: network preferences

rperez89:
I created this branch from the court stats one since i needed some stuffs from there, so it will be good to review first #229 

Also leaving here a screenshot for further discussions about improvements/changes

Screen Shot 2020-03-17 at 12 55 56 PM


src/networks.js
```diff
@@ -10,37 +11,29 @@ export const RINKEBY_USABILITY_COURT =
 export const RINKEBY_STAGING_COURT =
   '0x52180Af656A1923024D1ACcF1D827AB85cE48878'
 
-export const networks = {
+export const networkConfigs = {
   main: {
     court: '0xee4650cBe7a2B23701D416f58b41D8B76b617797',
```
sohkai, line 16: If it's not too much work, it would be great to key these under a `addresses` key like we have in [the client](https://github.com/aragon/aragon/blob/master/src/network-config.js#L13).

src/local-settings.js
```diff
@@ -0,0 +1,38 @@
+import env from './environment'
+import { IPFS_ENDPOINT } from './endpoints'
+
+const DEFAULT_ETH_NODE = 'DEFAULT_ETH_NODE'
```
sohkai, line 4: Like in the client, I'd recommend adding a `DASHBOARD_` prefix to these keys and allowing them to be configured via environment variables.

Afterwards, we should also document these like we do in the client: https://github.com/aragon/aragon/blob/master/docs/CONFIGURATION.md

src/lib/web3-utils.js
```diff
@@ -1,13 +1,15 @@
...
-
+import { InvalidURI, InvalidNetworkType, NoConnection } from '../errors'
 export const soliditySha3 = solidityKeccak256
 export const hash256 = keccak256
 export const DEFAULT_LOCAL_CHAIN = 'private'
 export const ETH_FAKE_ADDRESS = `0x${''.padEnd(40, '0')}`
 
 const ETH_ADDRESS_SPLIT_REGEX = /(0x[a-fA-F0-9]{40}(?:\b|\.|,|\?|!|;))/g
 const ETH_ADDRESS_TEST_REGEX = /(0x[a-fA-F0-9]{40}(?:\b|\.|,|\?|!|;))/g
+const websocketRegex = /^wss?:\/\/.+/
```
sohkai, line 12: Actually in this case I think we should be using http or https—we don't use the websockets provider at the moment.

src/lib/web3-utils.js
```diff
@@ -148,3 +159,41 @@ export function transformAddresses(str, callback) {
...
+ * @returns {Promise} Resolves if the ETH node is compatible, otherwise throws:
+ *    - InvalidURI: URI given is not compatible (e.g. must be WebSockets)
+ *    - InvalidNetworkType: ETH node connected to wrong network
+ *    - NoConnection: Couldn't connect to URI
+ */
+export async function checkValidEthNode(uri) {
+  // Must be websocket connection
+  const isLocalOrUnknown = isLocalOrUnknownNetwork(env('CHAIN_ID'))
+  if (!isLocalOrUnknown) {
+    if (!websocketRegex.test(uri)) {
```
sohkai, line 176: I think we should do this check regardless of the network being local or unknown—we only support http/https right now.

src/errors.js
```diff
@@ -1,6 +1,14 @@
...
 
+export const extendError = (name, { defaultMessage }) =>
+  class extends Error {
+    name = name
+    constructor(message = defaultMessage) {
+      super(message)
+    }
+  }
+
 export class DisputeNotFound extends Error {
```
sohkai, line 12: Maybe we can slightly extend the `extendError()` options, such that we can also use it for `DisputeNotFound`?

We could support an arbitrary object and then apply `Object.entries(options).forEach((key, value) => this[key] = value)`.

src/Routes.js
```diff
@@ -18,3 +21,7 @@ export default function Routes() {
     </Switch>
   )
 }
+
+export function getPreferencesSearch(screen) {
```
sohkai, line 25: We didn't use react-router in the client, but I wonder if there's a "framework appropriate" way to handle this with react-router.

src/components/Dashboard/Balance.js
```diff
@@ -47,7 +47,7 @@ const Balance = React.memo(function Balance({
         `}
       >
         {loading ? (
-          <Loading height={86} />
+          <Loading height={150} />
```
sohkai, line 50: I think @fabriziovigevani changed this recently because it was off balance, wanted to ping to re-confirm.

src/components/MainView.js
```diff
@@ -1,13 +1,39 @@
...
 function MainView({ children }) {
   const { width: vw, below } = useViewport()
   const compactMode = below('medium')
-
   const [menuPanelOpen, setMenuPanelOpen] = useState(!compactMode)
+  const history = useHistory()
+  const search = useLocation().search
+  const locator = useLocation()
+
+  function parsePreferences(search = '') {
```
sohkai, line 19: As mentioned earlier, there is likely a better way to handle query parameters like this within react-router.

src/hooks/useCourtContracts.js
```diff
@@ -625,3 +633,57 @@ export function useTotalANTStakedPolling(timeout = 1000) {
...
+    if (!jurorRegistryContract) {
+      return
+    }
+    const getTotalActiveBalance = async () => {
+      retryMax(() => jurorRegistryContract.totalActiveBalance())
+        .then(totalActive => {
+          setTotalActiveBalance(totalActive)
+        })
+        .catch(err => {
+          console.error(`Error fetching ANT staked: ${err}`)
```
sohkai, line 681:
```suggestion
          console.error(`Error fetching total active ANJ: ${err}`)
```

src/hooks/useCourtContracts.js
```diff
@@ -625,3 +633,57 @@ export function useTotalANTStakedPolling(timeout = 1000) {
 
   return [totalANTStaked, error]
 }
+
+export function useTotalANTStaked() {
```
sohkai, line 637: Just wondering—weren't both of these implemented in the PR for the stats?

In any case, they are both missing the "cancel" pattern :).

src/components/GlobalPreferences/GlobalPreferences.js
```diff
@@ -0,0 +1,208 @@
...
+const NETWORK_INDEX = 0
+
+const AnimatedDiv = animated.div
+
+function GlobalPreferences({ compact, onClose, onNavigation, sectionIndex }) {
+  const theme = useTheme()
+
+  useEsc(onClose)
+
+  const tabItems = VALUES.filter((_, index) => index === NETWORK_INDEX)
```
sohkai, line 33: We don't really need this bit—this was only used in the client to filter some tabs during onboarding.

src/components/GlobalPreferences/GlobalPreferences.js
```diff
@@ -0,0 +1,208 @@
...
+      items={sectionIndex !== null}
+      from={{ opacity: 0, enterProgress: 0, blocking: false }}
+      enter={{ opacity: 1, enterProgress: 1, blocking: true }}
+      leave={{ opacity: 0, enterProgress: 1, blocking: false }}
+      config={springs.smooth}
+    >
+      {show =>
+        show &&
+        /* eslint-disable react/prop-types */
+        // z-index 2 on mobile keeps the menu above this preferences modal
```
sohkai, line 166: We probably need to update this comment since the z-index seems to be 1 now.

src/components/GlobalPreferences/GlobalPreferences.js
```diff
@@ -0,0 +1,208 @@
...
+  )
+
+  useEffect(() => {
+    if (!path) {
+      setSectionIndex(null)
+      return
+    }
+    const index = PATHS.findIndex(item => path.startsWith(item))
+
+    if (index !== NETWORK_INDEX) {
```
sohkai, line 101: We should be able to remove this bit, since we're not filtering for just the Network tab.

src/components/GlobalPreferences/GlobalPreferences.js
```diff
@@ -0,0 +1,208 @@
...
+            </Bar>
+          )}
+          <main>{sectionIndex === NETWORK_INDEX && <Network />}</main>
+        </React.Fragment>
+      </Layout>
+    </div>
+  )
+}
+
+function useGlobalPreferences({ path = {}, onScreenChange }) {
```
sohkai, line 84:
```suggestion
function useGlobalPreferences({ path = '', onScreenChange }) {
```

It looks like `path` is used directly like a string here.

src/components/GlobalPreferences/GlobalPreferences.js
```diff
@@ -0,0 +1,208 @@
...
+          css={`
+            color: ${theme.surfaceIcon};
+          `}
+        />
+      </ButtonIcon>
+    </div>
+  )
+}
+
+function AnimatedGlobalPreferences(props) {
```
sohkai, line 144:
```suggestion
function AnimatedGlobalPreferences({ onScreenChange, path }) {
```

And then we can remove the `props.` usage later.

src/components/GlobalPreferences/GlobalPreferences.js
```diff
@@ -0,0 +1,208 @@
...
+              ),
+            }}
+            css={`
+              position: fixed;
+              top: 0;
+              bottom: 0;
+              left: 0;
+              right: 0;
+              overflow: auto;
+              min-width: 360px;
```
sohkai, line 187: This should be a GU (please also update this in the client!)

src/components/GlobalPreferences/GlobalPreferences.js
```diff
@@ -0,0 +1,208 @@
...
+              padding-bottom: ${2 * GU}px;
+              border-top: 2px solid ${theme.accent};
+              background: ${theme.surface};
+              ${breakpoint('medium', `padding-bottom:0;`)}
```
sohkai, line 191: Ah... yes... this the GlobalPreferences is probably the last bit in the client that hasn't been cleaned up with the design system.

We should use `useViewport()` instead of `breakpoint()`! (And please also update the client when doing so!)

src/components/GlobalPreferences/Network/Network.js
```diff
@@ -0,0 +1,182 @@
...
+              `}
+            >
+              {(() => {
+                if (networkError instanceof InvalidNetworkType) {
+                  return `Node must be connected to ${sanitizeNetworkType(
+                    networkType
+                  )}`
+                }
+                if (networkError instanceof InvalidURI) {
+                  return 'Must provide WebSocket endpoint to node'
```
sohkai, line 67:
```suggestion
                  return 'Must provide Http endpoint to node'
```

As mentioned earlier

src/components/GlobalPreferences/Network/Network.js
```diff
@@ -0,0 +1,182 @@
...
+              ${textStyle('body2')};
+              color: ${theme.contentSecondary};
+            `}
+          />
+        </Label>
+        <Button mode="strong" onClick={handleNetworkChange} wide={compact}>
+          Save changes
+        </Button>
+      </Box>
+      <Box heading="Troubleshooting">
```
sohkai, line 93: "Clearing cache" is an unnecessary action in this app.

However, we should talk with @dizzypaty about a "reset" button for the network state instead (which will effectively clear cache).

src/components/GlobalPreferences/Network/Network.js
```diff
@@ -0,0 +1,182 @@
...
+    setDefaultEthNode(ethNode)
+    setIpfsGateway(ipfsGateway)
+    window.location.reload()
+  }, [ethNode, ipfsGateway])
+
+  const handleClearCache = useCallback(() => {
+    window.localStorage.clear()
+  }, [])
+
+  const handleKeyPress = useCallback(
```
sohkai, line 145: I'm not sure, but do we have a `useEnterKey()` type of hook (I saw that we had one for `useEscKey()`)?

src/components/GlobalPreferences/Network/Network.js
```diff
@@ -0,0 +1,182 @@
...
+    handleNetworkChange,
+  } = useNetwork()
+  const theme = useTheme()
+
+  const { layoutName } = useLayout()
+  const compact = layoutName === 'small'
+
+  return (
+    <React.Fragment>
+      <Box heading="Node settings">
```
sohkai, line 41: We should add a setting for the subgraph API endpoint (and like the ETH node, do some sanity checks against it to make sure it works).

src/components/MainView.js
```diff
@@ -25,6 +51,16 @@ function MainView({ children }) {
     setMenuPanelOpen(!compactMode)
   }, [compactMode])
 
+  if (preferenceOption.path) {
+    return (
+      <GlobalPreferences
```
sohkai, line 56: Will this appear even if there is a global error?

src/local-settings.js
rperez89: we have all the env variables [here](https://github.com/aragon/court-dashboard/blob/a739f7896cb846c0a2e4302fdcb7279c9b28c18e/src/environment.js#L31), do you think that we need to change those names?

src/hooks/useCourtContracts.js
rperez89: yes, both of those functions are old ones.. just removed them

src/errors.js
rperez89: could you give me an example about how are you thinking to do this?

src/components/GlobalPreferences/GlobalPreferences.js
rperez89: At some point i remember not sure if it was from patty or pierre that we use the GU for paddings, margins , etc but not for `height` `width`

src/components/GlobalPreferences/Network/Network.js
rperez89: just to undertand why do we want to let the user change the data source of the dashboard?

rperez89: just to understand why do we want to let the user change the data source of the dashboard?

src/components/MainView.js
rperez89: did a research and asked Pierre about this and didn't find a better way to handle the parameters, seems that we need to handle the query parameters by ourselves

src/components/MainView.js
rperez89: not sure if i follow here, we handle the global errors [here](https://github.com/aragon/court-dashboard/blob/57a4000e3a7cff114a47d95e8a3ec9f26aa6ed02/src/App.js#L26)

src/local-settings.js
sohkai: Nevermind me—this looks all good as-is!

src/components/GlobalPreferences/Network/Network.js
sohkai: In case they wanted to run their own hosted instance of the subgraph, just like the ETH node or IPFS gateway.

src/components/MainView.js
sohkai: Yep—this question was to check whether we would be able to change the network settings in the global error.

For example, if I change one of the network errors, but then get a fatal global error, it would be ideal to reset the settings there. Otherwise, you're more or less stuck until someone from support manually intervenes to help you clear storage.

src/lib/web3-utils.js
```diff
@@ -9,7 +9,7 @@ export const ETH_FAKE_ADDRESS = `0x${''.padEnd(40, '0')}`
 
 const ETH_ADDRESS_SPLIT_REGEX = /(0x[a-fA-F0-9]{40}(?:\b|\.|,|\?|!|;))/g
 const ETH_ADDRESS_TEST_REGEX = /(0x[a-fA-F0-9]{40}(?:\b|\.|,|\?|!|;))/g
-const websocketRegex = /^wss?:\/\/.+/
+const httpRegex = /^(http|https):\/\//
```
sohkai, line 12:
```suggestion
const httpRegex = /^(https?):\/\//
```

😄

src/lib/web3-utils.js
```diff
@@ -172,10 +172,9 @@ export function transformAddresses(str, callback) {
 export async function checkValidEthNode(uri) {
   // Must be websocket connection
   const isLocalOrUnknown = isLocalOrUnknownNetwork(env('CHAIN_ID'))
-  if (!isLocalOrUnknown) {
-    if (!websocketRegex.test(uri)) {
-      throw new InvalidURI('The URI must use the WebSocket protocol')
-    }
+
+  if (!httpRegex.test(uri)) {
+    throw new InvalidURI('The URI must use the Http protocol')
```
sohkai, line 177:
```suggestion
    throw new InvalidURI('The URI must use the HTTP protocol')
```

src/errors.js
sohkai: Nevermind, what I was thinking doesn't work! Trying to plug this extendibility into the constructor isn't worth it.

src/networks.js
rperez89: is it ok for you if i change this in a separated pr only for that?

sohkai: Yep—let's do it!

src/hooks/usePreferences.js
```diff
@@ -0,0 +1,38 @@
+import { useCallback, useRef } from 'react'
+import { useHistory, useLocation } from 'react-router-dom'
+import { getPreferencesSearch } from '../Routes'
+
+/**
+ * Hook to interact with the preferences
+ * @returns {Array} first parameter open preferences handler, second parameter close preferences handler
```
sohkai, line 7:
```suggestion
 * @returns {Array} [open preferences handler, close preferences handler, current preferences screen]
```

I _think_ this is fine for JSDoc... do you think it makes the return description clearer?

src/hooks/usePreferences.js
```diff
@@ -0,0 +1,38 @@
...
+ *  third parameter the selected preference option
+ */
+export default function usePreferences() {
+  /* We need to keep track of the path where the preference was called 
+    in order to return to the same path when the preference modal is closed
```
sohkai, line 12: Let's just use inline comments for this, since it's not a long comment

src/hooks/usePreferences.js
```diff
@@ -0,0 +1,38 @@
...
+  */
+  const { pathname: previousPath } = useLocation()
```
sohkai, line 14: Could we name this `basePath` instead? Technically you're still on the same path, just one with an additional query parameter.

src/hooks/usePreferences.js
```diff
@@ -0,0 +1,38 @@
...
+  const history = useHistory()
+
+  const preferenceOption = useRef('')
+
+  const handleOpenPreferences = useCallback(
+    screen => {
+      preferenceOption.current = screen
```
sohkai, line 21: I might rename all instances of `preferenceOption` to `preferenceScreen`, to align the argument names. "Option" is a bit overloaded when talking about preferences.

src/hooks/usePreferences.js
rperez89: yeah i think that is ok the way you suggested

src/components/GlobalPreferences/Network/Network.js
rperez89: https://github.com/aragon/court-dashboard/issues/273

src/hooks/useKeyboardArrows.js
```diff
@@ -39,3 +39,19 @@ export function useEsc(cb) {
     return () => window.removeEventListener('keydown', handlekeyDown)
   }, [handlekeyDown])
 }
+
+export function useEnterKey(cb, validated) {
```
sohkai, line 43: Even if it means another `useCallback()` in the component itself, I think we should leave the `validated` bit to the callback handler—best not to add too much surface area to this hook.

src/hooks/usePreferences.js
```diff
@@ -1,38 +1,53 @@
...
+
+// Preferences base query string
+const GLOBAL_PREFERENCES_QUERY_PARAM = '?preferences='
+
+export function getPreferencesSearch(screen) {
+  return `${GLOBAL_PREFERENCES_QUERY_PARAM}${screen}`
+}
+
+function parsePreferences(search = '') {
+  const [, path = ''] = search.split(GLOBAL_PREFERENCES_QUERY_PARAM)
```
sohkai, line 12: We may want to use [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) to do this parsing instead—I'm always a bit queasy doing a naive string match on `?<param>=` 😄

src/hooks/usePreferences.js
```diff
@@ -1,38 +1,53 @@
...
-    in order to return to the same path when the preference modal is closed
-  */
-  const { pathname: previousPath } = useLocation()
+export default function usePreferences(appCrashed = false) {
+  // We need to keep track of the path where the preference was called in order to return to the same path when the preference modal is closed
+  const { pathname, search } = useLocation()
   const history = useHistory()
 
-  const preferenceOption = useRef('')
+  const searchParmFromUrl = parsePreferences(search)
```
sohkai, line 27:
```suggestion
  const searchParamFromUrl = parsePreferences(search)
```

src/hooks/usePreferences.js
```diff
@@ -1,38 +1,53 @@
...
+  const handleOpenPreferences = screen => {
+    preferenceScreen.current = screen
+    const fullPath = basePath + getPreferencesSearch(preferenceScreen.current)
+    history.push(fullPath)
+    if (appCrashed) {
+      window.location.reload()
+    }
+  }
 
-  const handleClosePreferences = useCallback(() => {
```
sohkai: To be clear, in the previous comment about `useCallback()`s, I think it is better to leave these callbacks memoized!

I would assume all return parameters from hooks are memoized correctly already (that is their responsibility), and use them in components without wrapping them into another `useMemo()` or `useCallback()`.

src/hooks/usePreferences.js
```diff
@@ -1,38 +1,53 @@
...
-      const fullPath = previousPath + getPreferencesSearch(screen)
-      history.push(fullPath)
-    },
-    [history, previousPath]
-  )
+  const handleOpenPreferences = screen => {
+    preferenceScreen.current = screen
+    const fullPath = basePath + getPreferencesSearch(preferenceScreen.current)
+    history.push(fullPath)
+    if (appCrashed) {
```
sohkai, line 38: Just wondering—wouldn't this lose the error context? I think we should open the preferences _inside_ the global error handler, rather than refreshing immediately and then opening preferences.

src/components/GlobalPreferences/Network/Network.js
```diff
@@ -0,0 +1,170 @@
...
+              ${textStyle('body2')};
+              color: ${theme.contentSecondary};
+            `}
+          />
+        </Label>
+        <Button mode="strong" onClick={handleNetworkChange} wide={compact}>
+          Save changes
+        </Button>
+      </Box>
+      <Box heading="Troubleshooting">
```
sohkai, line 93: In terms of this section, to get this merged for now, I would recommend the following copy changes:

```
Troubleshooting

Press this button to reset the network settings to their defaults.

<button>Reset network settings</button>

// remove Info component
```

src/hooks/useKeyboardArrows.js
rperez89: not sure if i follow the idea, i am pretty sure that a hook can not be called inside a useCallback

src/hooks/usePreferences.js
rperez89: This is something that we talked today with @bpierre , that if we are calling the preferences from the global error we should redirect to the home `/` and do a reload

src/hooks/useKeyboardArrows.js
sohkai: Ah, what I meant here is that we should keep the API of `useEnteryKey()` as

```
function useEnterKey(cb)
```

And let the component handle its own validation logic—even if it means inside that component, there's a small event handler wrapped with `useCallback()`.

src/local-settings.js
```diff
@@ -20,6 +20,12 @@ function setLocalSetting(confKey, value) {
   return window.localStorage.setItem(storageKey, value)
 }
 
+export function clearLocalStorageNetworkSettings() {
+  // TODO - remove subgrapk key onces available
```
sohkai, line 24:
```suggestion
  // TODO - remove subgraph key once available
```

src/components/GlobalPreferences/Network/Network.js
```diff
@@ -0,0 +1,164 @@
...
+        >
+          <span>
+            Press this button to reset the network settings to their defaults.
+          </span>
+        </div>
+        <Button
+          css={`
+            margin-bottom: ${2 * GU}px;
+          `}
+          onClick={clearLocalStorageNetworkSettings}
```
sohkai, line 109: Ahh, we should also refresh the app here! Otherwise this won't do anything until you do haha.

src/components/MainView.js
```diff
@@ -38,7 +51,11 @@ function MainView({ children }) {
           flex-shrink: 0;
         `}
       >
-        <Header compactMode={compactMode} toggleMenuPanel={toggleMenuPanel} />
+        <Header
```
sohkai, line 54: Ahh, something I also just noticed, we should remove the bit of padding or margin separating the activity panel with the preferences panel surface:

Screen Shot 2020-04-08 at 3 25 39 PM

(The orange higlight)

src/components/Errors/GlobalErrorScreen.js
```diff
@@ -23,18 +20,14 @@ function GlobalErrorScreen({ children }) {
...
-      <div
+      <img
         css={`
           position: absolute;
-          width: 100%;
-          display: flex;
-          justify-content: space-between;
-          padding: ${2 * GU}px ${2 * GU}px 0;
+          top: 16px;
+          left: 16px;
```
sohkai: Might as well make these GUs :)

PR #239, opened 2020-03-18
Update now deployments to build against staging in PRs

rperez89:
Now configs for the new QA environment


now-usability.json
```diff
@@ -0,0 +1,14 @@
+{
+    "version": 2,
+    "public": true,
+    "scope": "aragon",
+    "alias": "court-rinkeby.aragon.org",
```
facuspagnuolo: Should we call this `court-usability.aragon.org` to avoid confusing it with the actual Court Rinkeby instance?

sohkai:
> If that's correct, could it happen that two different PRs overlap each other with the staging instance trying to deploy different versions?

No; `now` will deploy to a different staging URL per-branch.

now-usability.json
sohkai: So far we've been pretending usability is the rinkeby instance, and that we'd do a switchover later at some point.

Perhaps it would be better as `court-mock.aragon.org` or `mock.court.aragon.org` instead though.

.github/workflows/usability.yml
```diff
@@ -0,0 +1,32 @@
+
```
sohkai: Let's remove this extra line :)

now-usability.json
facuspagnuolo: I was suggesting the other cause we followed that terminology for all the other pieces of each env already 😅 

I prefer leaving it as `rinkeby` if you prefer not using `usability` then

sohkai:
`court-staging.aragon.org` should be pointing to Zeit now.

PR #240, opened 2020-03-19
dispute: error screens for dispute detail and timeline

### LINKED ISSUE #221: Error handling for dispute detail graph query
Right now we are not handling the errors that might happen when querying the graph

rperez89:
This pr includes Error screen for the dispute detail when we have an error from the graphql and also an error screen for the timeline

This closes #221 

Screen Shot 2020-03-18 at 11 16 41 PM


src/assets/LogoAccent.svg
```diff
@@ -1 +1 @@
...
...6-2.019 3.296a6.878 6.878 0 00-1.014.445c.935-1.76 1.849-3.077 2.741-3.95a4.453 4.453 0 01.95-1.705zm1.776 1.804l-2.18 1.027c.621.01 1.114-.102 1.48-.338.365-.236.599-.466.7-.689zm-.804-.374c-.013-.449-.071-.7-.176-.757-.116-.051-.184-.163-.116-.367a.614.614 0 00-.038-.434L7.818 3l.666.419a.465.465 0 01-.145.409c.145.156.145.359-.023.592.004.058.008.115.01.17.002.085-.001.205-.01.362h-.76z" fill="#fff" fill-opacity=".6"/><path d="M15.338 11.363c-.094 1.722.27 2.7 1.089 2.935 1.23.353 3.415.4 3.275 3.07.111.354.213.69.305 1.008-.36.386-.752.741-1.17 1.064a4.235 4.235 0 00-1.747-2.072c-2.83-1.78-3.415-3.782-1.752-6.005zm-7.961 5.939c1.652 4.14 4.297 1.488 6.34 2.457.934.444 1.444 1.258 1.529 2.442-.797.185-1.627.283-2.48.283a10.88 10.88 0 01-4.625-1.025c-.005-1.246-.26-2.632-.764-4.157zm3.134-11.194c.429-.118.78-.385 1.051-.8-.017.247-.08.46-.186.638 1.239-.222 2.2-.971 2.884-2.247-1.261-.685-2.561-.218-3.258.634-.465.567-.628 1.159-.49 1.775z" fill="#fff" fill-opacity=".6"/><defs><linearGradient id="paint0_linear" x1="37.619" y1="-10.8" x2="-.457" y2="18.042" gradientUnits="userSpaceOnUse"><stop stop-color="#FFB36D"/><stop offset="1" stop-color="#F88"/></linearGradient></defs></svg>
```
fabriziovigevani: In the future we can just do `svgo ./file-path` so we don't have this diffs :D

src/hooks/useDisputes.js
```diff
@@ -19,6 +20,7 @@ const DISPUTE_PROCESSED_DEFAULT = {
   error: false,
```
fabriziovigevani: I think this should be `''` now

src/hooks/useDisputes.js
```diff
@@ -61,25 +63,46 @@ export default function useDisputes() {
...
+          }
+        : null,
+      fetching,
+      error,
+      !!graphErrorMessage,
+    ]
+  }, [disputeProcessed, disputePhaseKey, error, graphFetching]) // eslint-disable-line react-hooks/exhaustive-deps
 }
 
 function useProcessedDispute(dispute, fetching) {
```
fabriziovigevani: I'd like to refactor this to how it was implemented before (without setting a default state of `DISPUTE_PROCESSED_DEFAULT` object) because logically if the dispute does not exist it could generate more confusion than the benefits it provides and we could have race conditions between when the dispute is done fetching from the graph and it exists,  and we are still fetching from IPFS.

We could set the default state as `null` and calculate if the dispute is fetching under this condiitons:

- we are still fetching the data from the subgraph. 
- we are done fetching form subgraph
    - if the dispute exists then we are stil fetching if `disputeProcessed != null`
    - if dispute does not exists we just return

Also in this way we separate the fetching logic from this hook.

src/hooks/useDisputes.js
```diff
@@ -61,25 +63,46 @@ export default function useDisputes() {
...
+        ? {
+            ...disputeProcessed,
+            ...disputePhase,
+          }
+        : null,
+      fetching,
+      error,
+      !!graphErrorMessage,
+    ]
+  }, [disputeProcessed, disputePhaseKey, error, graphFetching]) // eslint-disable-line react-hooks/exhaustive-deps
```
fabriziovigevani: I think we may want to temporarily remove the `eslint-disable` rule and check if we may have more props to add here

src/hooks/useDisputes.js
rperez89: cc @sohkai as he suggested to use the default state

src/hooks/useDisputes.js
sohkai: If you have the editor integration set up, this should be easy to do: just remove the comment, run prettier, and then add the comment back.

src/hooks/useDisputes.js
sohkai: The new version looks good! And agreed, having it be `null` makes it more clear it's still being fetched.

Two notes:

1. We forgot to use the `cancel` pattern in `useProcessedDispute()`
2. Are we handling errors returned from `useProcessedDispute()` correctly?

fabriziovigevani: > We forgot to use the cancel pattern in useProcessedDispute()

Ah yes good point!

> Are we handling errors returned from useProcessedDispute() correctly?

Yes, the handling is made here https://github.com/aragon/court-dashboard/blob/55271e90738eebf71e192f56c4d95f0cbc7757cf/src/hooks/useDisputes.js#L81

Will do a PR with the cancel pattern.

PR #241, opened 2020-03-19
Email: add generic-action template

sohkai:
See https://www.figma.com/file/CS1ubWwI37gkjj8Uq3xdKH/Restart-precedence-campaing-%E2%80%93-Email-notifications?node-id=4%3A802

Screen Shot 2020-03-19 at 3 29 28 PM

- [x] Update `icon-generic.png` (needs a re-export from someone who has write access to Figma)

-----

Also updates some templates' parameterization to be more consistent with other template (e.g. always use `Html` prefix for attributes that are html). Additionally strips whitespace resulting from multi-line strings in the text templates.


emails/template-utils.js
```diff
@@ -36,6 +36,11 @@ function asset(path) {
   return `${assetBasePath}/${path}`
 }
 
+function stripWhitespace(str) {
```
bpierre: Should this be something like `trimStartMultiline`, to express that it’s at the start and multiline? Also trying to imitate https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trimStart

sohkai: Updated in 5186287.

Made it so we trimmed all starting and ending spaces off of new lines :).

PR #242, opened 2020-03-19
Sentry: update initialization and usage

sohkai:
Slims down the logic a bit, and consolidates initialization logic into its own file.

### DIFF
--- a/src/sentry.js
+++ b/src/sentry.js
@@ -1,0 +1,16 @@
+import * as Sentry from '@sentry/browser'
+import env from './environment'
+import { getNetworkType } from './lib/web3-utils'
+
+const SENTRY_DSN = env('SENTRY_DSN')
+export const sentryEnabled = Boolean(SENTRY_DSN && env('ENABLE_SENTRY'))
+
+export default function initializeSentry() {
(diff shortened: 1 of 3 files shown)


src/sentry.js
```diff
@@ -0,0 +1,16 @@
+import * as Sentry from '@sentry/browser'
```
sohkai: Wondering where we should put this file; I was mainly thinking between this and something like `configuration/sentry.js`.

fabriziovigevani: I think we can leave it here for now and if we have the need to move all config files (endpoints, networks, etc) we can do so later?

sohkai:
> Seeing that we are starting to report errors from different places in the codebase, does it make sense to provide a reportError() function from sentry.js that does the captureException()?

We would have to export both `captureMessage()` and `captureException()` (or a single wrapper that detected types). I think consolidating into `sentry.js` would make sense if we were planning on adding additional logic to the errors (e.g. filtering, parsing, etc.), but I don't see too many uses cases of this at the moment.

We could also directly import the `captureException()` directly instead of `import * as Sentry`.

Thoughts?

PR #244, opened 2020-03-20
Precedence campaign banner

### LINKED ISSUE #223: Precedence campaign banner
See https://www.notion.so/aragonone/Precedence-campaign-banner-837e078414914e938d26db97871b975d

fabriziovigevani:
closes #223 

**We need to figure out on Monday:**

- A way to flag a dispute as part of the P.Campaign from the data itself.
- If a user hasn't marked the banner / modal as read, we will show the banner until they do.
  When they have already read the modal content, should we show somewhere in the Dispute detail UI another link  (not so highlighted as the Banner) so users can re-open and re-read the modal content again?

Preview: 

Screen Shot 2020-03-23 at 2 46 25 PM

Screen Shot 2020-03-23 at 2 43 05 PM

Screen Shot 2020-03-23 at 2 44 50 PM


src/components/Disputes/PrecedenceCampaign/PCBanner.js
```diff
@@ -0,0 +1,74 @@
+import React, { useCallback, useState } from 'react'
+import { Button, GU, useTheme } from '@aragon/ui'
+import Modal from './PCModal'
+
+const KEY_PREFIX = 'PC_DISPUTE_BANNER_READ'
+
+function PCBanner({ disputeId }) {
```
rperez89: could we avoid the abbreviations?

src/components/Disputes/PrecedenceCampaign/PCModal.js
```diff
@@ -0,0 +1,119 @@
+import React from 'react'
+import { Button, GU, Modal, textStyle, useTheme, useViewport } from '@aragon/ui'
+
+import pcModalSvg from './assets/PCModal.svg'
+
+function PCModal({ opened, onClose }) {
```
rperez89: Same here!

src/components/Disputes/PrecedenceCampaign/assets/PCModal.svg
```diff
@@ -0,0 +1 @@
...856c3.082 0 6.184.911 8.791 2.608.427-.586.892-1.128 1.28-1.567 2.074-2.351 5.147-4.011 8.268-3.723 4.439.411 7.822 4.47 12.232 5.109 4.139.602 8.026-1.925 12.116-2.797 4.343-.925 9.218.209 12.281 3.421 3.063 3.211 3.887 8.546 1.435 12.247-3.17 4.796-10.313 5.309-15.586 3-5.264-2.308-9.286-6.677-13.706-10.358-2.869-2.384-6.146-4.587-9.868-4.841-2.326-.159-4.216.558-6.077 1.427 2.52 2.51 4.178 5.999 4.178 10.358 0 9.456-6.844 14.108-6.844 14.108v-6.211c1.89-.536 3.277-2.25 3.277-4.305 0-2.056-1.387-3.77-3.277-4.306a4.378 4.378 0 00-1.221-.19c-.892 0-1.706.265-2.404.709-3.131-2.251-4.1-6.233-4.1-6.233-6.358 7.682-16.071 7.668-16.071 7.668-3.005-12.869 7.696-16.124 15.296-16.124z" fill="#352C47"/><rect x=".11" y="366.713" width="178.142" height="174.378" rx="10" fill="#EFF2FF"/><defs><linearGradient id="paint0_linear" x1="-17.639" y1="292.18" x2="209.423" y2="79.977" gradientUnits="userSpaceOnUse"><stop stop-color="#FFB36D"/><stop offset="1" stop-color="#F88"/></linearGradient><linearGradient id="paint1_linear" x1="229.406" y1="116.955" x2="297.801" y2="35.317" gradientUnits="userSpaceOnUse"><stop stop-color="#FFB36D"/><stop offset="1" stop-color="#F88"/></linearGradient></defs></svg>
```
rperez89: should we optimize this svg?

src/flagged-disputes/precedence-campaign-disputes.js
```diff
@@ -0,0 +1,19 @@
+import { networks, RINKEBY_COURT, RINKEBY_USABILITY_COURT } from '../networks'
+import { getNetworkType } from '../lib/web3-utils'
+
+const PRECEDENCE_CAMPAIGN_DISPUTES = {
+  rpc: new Map([[networks.rpc.court, []]]),
+  ropsten: new Map([[networks.ropsten.court, []]]),
+  rinkeby: new Map([
+    [RINKEBY_COURT, []],
+    [RINKEBY_USABILITY_COURT, []],
```
rperez89: We also need to support STAGING

src/components/Disputes/PrecedenceCampaign/PCBanner.js
fabriziovigevani: Yes good point!

src/components/Disputes/PrecedenceCampaign/assets/PCModal.svg
fabriziovigevani: It's already optimized :D Github shows the whole svg wrapped but it's all in one line

src/utils/dispute-utils.js
```diff
@@ -53,6 +54,13 @@ export const transformDisputeDataAttributes = dispute => {
     }),
   }
 
+  // If the dispute is part of the precedence campaign we will flag it as such
+  const precedenceCamapignDisputes = getPrecedenceCampaignDisputesByCourt()
+  const isPartOfPrecedenceCampaign = precedenceCamapignDisputes.some(
+    disputeId => disputeId === dispute.id
```
sohkai: Not a big deal, but we might as well use a Map here like the voided ones. If not for performance, then just for consistency :).

src/components/Disputes/PrecedenceCampaign/PrecedenceCampaignBanner.js
```diff
@@ -0,0 +1,74 @@
...
+
+function PrecedenceCampaignBanner({ disputeId }) {
+  const theme = useTheme()
+  const [bannerVisible, setBannerVisible] = useState(
+    localStorage.getItem(`${KEY_PREFIX}${disputeId}`) === null
+  )
+  const [modalOpened, setModalOpened] = useState(false)
+
+  const closeBanner = useCallback(() => {
+    localStorage.setItem(`${KEY_PREFIX}${disputeId}`, 'true')
```
sohkai: Maybe be useful to have this as `${KEY_PREFIX}_${disputeId}`.

But it may also be useful to use a small API around this, that does serialization, like [`StoredList`](https://github.com/aragon/aragon/blob/master/src/StoredList.js).

src/components/Disputes/PrecedenceCampaign/PrecedenceCampaignBanner.js
```diff
@@ -0,0 +1,74 @@
...
+    setBannerVisible(false)
+  }, [disputeId])
+
+  const openModal = useCallback(() => setModalOpened(true), [])
+  const markRead = useCallback(() => {
+    closeBanner()
+    setModalOpened(false)
+  }, [closeBanner, setModalOpened])
```
sohkai: No need to have the `setModalOpened` in the dependencies here :).

src/components/Disputes/PrecedenceCampaign/PrecedenceCampaignModal.js
```diff
@@ -0,0 +1,119 @@
...
+    ? height - 40
+    : Math.max(700, Math.min(620, height - 40))
+
+  return (
+    <div>
+      <Modal
+        closeButton={false}
+        padding={0}
+        visible={opened}
+        width={Math.min(1055, width - 40)}
```
sohkai: Interesting, I wonder if there's a better way to do this cc @bpierre

src/components/Disputes/PrecedenceCampaign/PrecedenceCampaignModal.js
```diff
@@ -0,0 +1,119 @@
...
+        css={`
+          z-index: 4;
```
sohkai: I'm not sure, but do we need this z-index? Modal should already be rendering in a portal that's of higher z-index precedence than the main view via aragonUI.

src/components/Disputes/PrecedenceCampaign/PrecedenceCampaignModal.js
```diff
@@ -0,0 +1,119 @@
...
+              `}
+            >
+              <p
+                css={`
+                  ${textStyle('body2')};
+                  color: ${theme.contentSecondary};
+                  text-transform: uppercase;
+                `}
+              >
+                Aragon court
```
sohkai: Let's capitalize the name: **Court** 😄

src/components/Disputes/PrecedenceCampaign/PrecedenceCampaignModal.js
```diff
@@ -0,0 +1,119 @@
...
+              </h1>
+              <span
+                css={`
+                  ${textStyle(verticalMode ? 'body2' : 'body1')}
+                  line-height: 2;
+                  display: block;
+                `}
+              >
+                All information, arguments, and evidence presented in disputes
+                related to precedence campaign portray scenarios constructed by
```
sohkai: to **the** precedence

src/flagged-disputes/precedence-campaign-disputes.js
```diff
@@ -0,0 +1,25 @@
...
+  RINKEBY_USABILITY_COURT,
+} from '../networks'
+import { getNetworkType } from '../lib/web3-utils'
+
+const PRECEDENCE_CAMPAIGN_DISPUTES = {
+  rpc: new Map([[networks.rpc.court, ['0']]]),
+  ropsten: new Map([[networks.ropsten.court, []]]),
+  rinkeby: new Map([
+    [RINKEBY_COURT, []],
+    [RINKEBY_STAGING_COURT, []],
```
sohkai: Just for testing purposes, could we also include dispute #0 on staging?

src/components/Disputes/PrecedenceCampaign/PrecedenceCampaignModal.js
fabriziovigevani: Sure, in any case note that the whole text is transformed to uppercase :D

src/components/Disputes/PrecedenceCampaign/PrecedenceCampaignBanner.js
fabriziovigevani: Oops

src/components/Disputes/PrecedenceCampaign/PrecedenceCampaignModal.js
fabriziovigevani: Yeah the problem is that it would render relative to the inner page but form your comment i'm realizing that we probably want to set that `z-index` at the  `<Root />` component

fabriziovigevani: Well actually if we do that, we have conflicts with the SideBar and Header's shadows

src/components/Disputes/PrecedenceCampaign/PrecedenceCampaignModal.js
bpierre: I think that’s how we generally do it, using `Math.min` / `Math.max` to add constraints to the modal dimensions.

In this specific case we could also use `max-width: 1055px; width: calc(100vw - 40px)` if we wanted to stay in CSS only, but I think that doing it in JS makes sense since we are doing other calculations.

Otherwise I think we could use more `GU` here :-)

src/components/Disputes/PrecedenceCampaign/PrecedenceCampaignModal.js
bpierre: Ah yes we have a new `Root.Provider` defined [on the content area](https://github.com/aragon/court-dashboard/blob/development/src/components/MainView.js#L65), so we need to escape it.

It would be nice to have named `Root.Provider` instances so we could choose :-)

PR #245, opened 2020-03-20
Add Refuse to vote hint

### LINKED ISSUE #234: Disputes: add hint for "Refuse to vote"
- [ ] Where to place hint
- [ ] Hint copy

fabriziovigevani:
closes #234 

~~Couldn't find a way to prevent the `<Button/>` from triggering it's `:active`  pseudo-class.
Suggestions are welcome :)~~

- Relocated the hint
- Fixed Dispute detail responsive issues

**Preview:** 

Screen Shot 2020-04-01 at 12 51 58 PM

Screen Shot 2020-04-01 at 12 52 43 PM


src/components/Disputes/actions/DisputeVoting.js
```diff
@@ -85,7 +85,27 @@ function VotingActions({ canJurorVote, onRequestCommit }) {
...
+            `}
+          >
+            Refuse to vote
+          </span>
+          <div
+            onClick={event => {
+              event.stopPropagation()
+            }}
+          >
+            <Help hint="Why would I refuse to vote?">
```
sohkai: @bpierre I wonder if we can add the ability to activate `Help` components on hover for cases like this (and disable their onclicks).

It does not seem like a great pattern to have to click to open this Help tooltip if it's inside a button.

rperez89: Agree on that i think that it might not make sense to click on a button tooltip

bpierre: I’m wondering if it makes sense to have this opener inside of the button? Even if we do it on hover, it wouldn’t work for touch-only screens where we would still have to click on it. /cc @dizzypaty

src/components/Disputes/DisputeInfo.js
```diff
@@ -329,9 +332,12 @@ function DisputeActionText({ dispute, isFinalRulingEnsured }) {
...
 const Row = styled.div`
   display: grid;
-  grid-template-columns: 1fr minmax(200px, auto);
-  grid-gap: ${5 * GU}px;
-  margin-bottom: ${2 * GU}px;
+
+  ${({ compactMode }) => `
+    grid-gap: ${(compactMode ? 2.5 : 5) * GU}px;
+    margin-bottom: ${compactMode ? 0 : 2 * GU}px;
+    grid-template-columns: ${compactMode ? 'auto' : '1fr minmax(200px, auto)'};
```
sohkai, line 339: Wonder if we should use a GU for the 200px :)

src/components/Disputes/actions/DisputeVoting.js
```diff
@@ -53,41 +53,53 @@ function VotingFinalRound({ draftTermId, onRequestCommit }) {
...
     <div>
       <div
         css={`
           display: flex;
           justify-content: space-between;
+          flex-direction: ${compactMode ? 'column' : 'row'};
           width: 100%;
           margin-bottom: ${1.5 * GU}px;
+
+          & > button {
```
sohkai, line 71: This is a bit brittle, since `Button` doesn't necessarily render to a `button` element.

It probably still makes sense to make this a custom component, something like:

```js
const VotingButton = styled(Button)`
  width: calc((100% - ${2 * GU}px) / 3);
  margin-bottom: ${({ compactMode }) => `${compactMode ? 1 : 0) * GU}px`)
`
```

src/components/Disputes/actions/DisputeVoting.js
```diff
@@ -102,8 +114,56 @@ function VotingActions({ canJurorVote, onRequestCommit }) {
...
+      `}
+    >
+      <div
+        css={`
+          width: ${width};
+          color: ${theme.help};
+        `}
+      >
+        <Container
+          style={{
```
sohkai, line 138: Does `Label` not support the `classNames` prop to allow us to use the `css` prop?

src/components/Disputes/actions/DisputeVoting.js
```diff
@@ -102,8 +114,56 @@ function VotingActions({ canJurorVote, onRequestCommit }) {
...
+          >
+            Why refuse to vote?
+          </span>
+          <Help hint="">
+            You can refuse to vote for many reasons, for example if you consider
+            that the evidence was not conclusive enough or the description was
+            incoherent. In any case, you won’t be penalized at this stage for
+            selecting any of these options. Remember that you should vote the
+            way that you think a majority of jurors will vote, since you will be
+            penalized if your vote is in the minority.
```
sohkai, line 157: > In any case, you won’t be penalized at this stage for selecting any of these options.

Just sanity checking—do we really want to say this? Sure, they won't get penalized _now_, but it's not like this is not an action without repercussion.

-------

Maybe it would be nice to move this into its own paragraph:

> Remember that you should vote the way that you think a majority of jurors will vote, since you will be penalized if your vote is in the minority.

src/components/Disputes/actions/DisputeVoting.js
```diff
@@ -102,8 +114,56 @@ function VotingActions({ canJurorVote, onRequestCommit }) {
...
+      css={`
+        display: flex;
+        align-items: center;
+        justify-content: ${compactMode ? 'center' : 'flex-end'};
+        margin-bottom: ${1.5 * GU}px;
+      `}
+    >
+      <div
+        css={`
+          width: ${width};
```
sohkai, line 133: Do we need to set the width of this component, if we've already applied `flex-end`?

src/components/Disputes/actions/DisputeVoting.js
```diff
@@ -53,41 +53,53 @@ function VotingFinalRound({ draftTermId, onRequestCommit }) {
...
+          & > button {
+            width: ${buttonWidth};
+            margin-bottom: ${(compactMode ? 1 : 0) * GU}px;
+          }
         `}
       >
-        <VotingButton
+        <Button
           mode="positive"
           wide
```
sohkai, line 79: We can most likely remove the `wide` declarations from each of the Buttons.

src/components/Disputes/actions/DisputeVoting.js
fabriziovigevani: When using the `css` prop directly i get a weird `Container is not defined` error

src/components/Disputes/actions/DisputeVoting.js
fabriziovigevani: Yes we do, because we want the hint to be centered and aligned relative to the refuse to vote hint on big scale displays

src/components/Disputes/actions/DisputeVoting.js
fabriziovigevani: > Just sanity checking—do we really want to say this? Sure, they won't get penalized now, but it's not like this is not an action without repercussion.

Yeah i think we could remove this part. For users that are new to the court could easily get the wrong even though we state below that they could be penalized.

> Maybe it would be nice to move this into its own paragraph:

👍

src/components/Disputes/actions/DisputeVoting.js
```diff
@@ -149,19 +147,28 @@ const RefuseToVoteHint = ({ compactMode, width }) => {
...
-            selecting any of these options. Remember that you should vote the
-            way that you think a majority of jurors will vote, since you will be
-            penalized if your vote is in the minority.
+            <p>
+              You can refuse to vote for many reasons, for example if you
+              consider that the evidence was not conclusive enough or the
+              description was incoherent.
+            </p>
+            <p>
+              Remember that you should vote the way that you think a majority of
```
sohkai, line 156:
```suggestion
              Remember that you should vote the way that you think the majority of
```

src/components/Disputes/actions/DisputeVoting.js
sohkai: Interesting... I wonder if this is because we're using `div` as a string.

Something we could do is use the [`as` prop in `styled-components`](https://styled-components.com/docs/api#as-polymorphic-prop) to swap between the rendered element. In the compact form, we may as well keep the `cursor: pointer` style.

fabriziovigevani: Nice

PR #246, opened 2020-03-21
Disputes: show juror weight hint for jurors drafted more than once

### LINKED ISSUE #236: Dispute: add hint for jurors being drafted more than once
See https://www.notion.so/aragonone/Communicate-to-jurors-if-they-ve-been-drafted-twice-for-the-same-dispute-vote-weight-lock-amount-20925e81957547ce9ccf5371069e5d76

fabriziovigevani:
closes #236 

When implementing this, i wondered whether we should also show this hint in the voting phase.
What do you think?

**Preview:** 

Screen Shot 2020-03-20 at 9 13 21 PM
Screen Shot 2020-03-20 at 9 13 26 PM

### DIFF
--- a/src/components/Disputes/DisputeCurrentRuling.js
+++ b/src/components/Disputes/DisputeCurrentRuling.js
@@ -117,15 +117,15 @@ function DisputeCurrentRuling({ dispute }) {
                       margin-right: ${0.5 * GU}px;
                     `}
                   >
-                    {showMyWeight ? 'Your v' : 'V'}oting weight
+                    {showMyWeight ? 'Your voting weight' : 'Voting weight'}
                   </span>
-                  <Help hint="What's the voting weight?">
+                  <Help hint="">
                     {showMyWeight ? 'You have been' : 'The same juror can be'}{' '}
                     drafted multiple times to arbitrate the same dispute for the
                     same round. When that happens,{' '}
                     {showMyWeight ? 'your' : 'their'} voting weight{' '}
                     {showMyWeight ? 'is' : 'will be'} proportional to the number
-                    of times {showMyWeight ? 'you are' : 'is'} drafted.
+                    of times {showMyWeight ? 'you' : 'they'} are drafted.
                   </Help>
                 </div>
               )}
--- a/src/components/Disputes/DisputeCurrentRuling.js
+++ b/src/components/Disputes/DisputeCurrentRuling.js
@@ -123,7 +123,7 @@ function DisputeCurrentRuling({ dispute }) {
                     the same dispute for the same round. When that happens,
                     their voting weight will be proportional to the number of
                     times is drafted, as well as the % of ANJ locked in the
-                    Active balance.{' '}
+                    Active balance.
                   </Help>
                 </div>
               )}
--- a/src/components/Disputes/DisputeCurrentRuling.js
+++ b/src/components/Disputes/DisputeCurrentRuling.js
@@ -56,9 +56,10 @@ function DisputeCurrentRuling({ dispute }) {
           percentage: weight,
         }))}
         renderFullLegendItem={({ color, item, index, percentage }) => {
-          // We'll show the juror voting weight hint only if the juror participated in the  current ruling
-          // and has been drafted more than once
-          const showMyVotingWeight = index === 0 && myWeight > 1
+          // We'll show the juror voting weight hint if any juror has been drafted more than once for this last round
+          const showVotingWeightHint =
+            index === 0 && lastRound.jurorsNumber !== lastRound.jurors.length
+          const showMyWeight = myWeight > 1
 
           return (
             <div
@@ -80,7 +81,7 @@ function DisputeCurrentRuling({ dispute }) {
                     background: ${color};
                     width: 8px;
                     height: 8px;
-                    margin-right: 8px;
+                    margin-right: ${1 * GU}px;
                     border-radius: 50%;
                   `}
                 />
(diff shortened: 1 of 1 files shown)


src/components/Disputes/DisputeCurrentRuling.js
```diff
@@ -55,42 +55,81 @@ function DisputeCurrentRuling({ dispute }) {
...
+                      margin-right: ${0.5 * GU}px;
+                    `}
+                  >
+                    Your voting weight
+                  </span>
+                  <Help>
+                    The same juror can be drafted multiple times to arbitrate
+                    the same dispute for the same round. When that happens,
+                    their voting weight will be proportional to the number of
+                    times is drafted, as well as the % of ANJ locked in the
```
sohkai:
```suggestion
                    times they are drafted, as well as the % of ANJ locked in the
```

I don't follow the second half of this sentence ("as well as the % of ANJ locked in the Active balance"). This is irrelevant to the voting weight, right? It only helps determine how many times they could be drafted.

src/components/Disputes/DisputeCurrentRuling.js
```diff
@@ -55,42 +55,81 @@ function DisputeCurrentRuling({ dispute }) {
...
-                width: 8px;
-                height: 8px;
-                margin-right: 8px;
-                border-radius: 50%;
-              `}
-            />
+        renderFullLegendItem={({ color, item, index, percentage }) => {
+          // We'll show the juror voting weight hint only if the juror participated in the  current ruling
+          // and has been drafted more than once
+          const showMyVotingWeight = index === 0 && myWeight > 1
```
sohkai: Not sure if this was discussed with @dizzypaty, but I think this information is useful to everyone involved, if there is a round where someone was drafted more than once.

We could change the text to be this if you are the actual juror who got drafted more than once though.

src/components/Disputes/DisputeCurrentRuling.js
fabriziovigevani: Yeah i think that part was meant as extra information but it should be changed to the `amount` rather than `%` of ANJ, or we could drop it entirely.

src/components/Disputes/DisputeCurrentRuling.js
```diff
@@ -55,42 +55,81 @@ function DisputeCurrentRuling({ dispute }) {
...
+                  display: flex;
+                  align-items: center;
+                `}
+              >
+                <div
+                  css={`
+                    background: ${color};
+                    width: 8px;
+                    height: 8px;
+                    margin-right: 8px;
```
rperez89: Let's use GU for margins

src/components/Disputes/DisputeCurrentRuling.js
fabriziovigevani: Yeah good point!

src/components/Disputes/DisputeCurrentRuling.js
```diff
@@ -116,14 +117,15 @@ function DisputeCurrentRuling({ dispute }) {
                       margin-right: ${0.5 * GU}px;
                     `}
                   >
-                    Your voting weight
+                    {showMyWeight ? 'Your v' : 'V'}oting weight
```
sohkai: Haha what do you think about just having two different strings?

Better to not be too clever, and it's also generally easier for collecting strings for internationalization :)

src/components/Disputes/DisputeCurrentRuling.js
```diff
@@ -116,14 +117,15 @@ function DisputeCurrentRuling({ dispute }) {
...
-                    their voting weight will be proportional to the number of
-                    times is drafted, as well as the % of ANJ locked in the
-                    Active balance.
+                  <Help hint="What's the voting weight?">
+                    {showMyWeight ? 'You have been' : 'The same juror can be'}{' '}
+                    drafted multiple times to arbitrate the same dispute for the
+                    same round. When that happens,{' '}
+                    {showMyWeight ? 'your' : 'their'} voting weight{' '}
+                    {showMyWeight ? 'is' : 'will be'} proportional to the number
+                    of times {showMyWeight ? 'you are' : 'is'} drafted.
```
sohkai:
```suggestion
                    of times {showMyWeight ? 'you are' : 'they are'} drafted.
```

src/components/Disputes/DisputeCurrentRuling.js
```diff
@@ -116,14 +117,15 @@ function DisputeCurrentRuling({ dispute }) {
...
-                    Your voting weight
+                    {showMyWeight ? 'Your v' : 'V'}oting weight
                   </span>
-                  <Help>
-                    The same juror can be drafted multiple times to arbitrate
-                    the same dispute for the same round. When that happens,
-                    their voting weight will be proportional to the number of
-                    times is drafted, as well as the % of ANJ locked in the
-                    Active balance.
+                  <Help hint="What's the voting weight?">
```
sohkai:
```suggestion
                  <Help hint="What's the voting weight?">
```

Maybe we don't need the `hint`; it doesn't really summarize what we explain below. If we wanted it, it'd be something like "Why do some jurors have more voting weight?", but this is not something a natural question from the information we currently show (we don't breakdown the voting weights at all, I think?).

src/components/Disputes/DisputeCurrentRuling.js
fabriziovigevani: Haha yes sure!

src/components/Disputes/DisputeCurrentRuling.js
fabriziovigevani: Good point. We don't break them down no, at least for now.

Will leave it `''`  since it's a required prop by the `<Help />`

PR #252, opened 2020-03-25
Refactor: Court subscriptions and treasury balances query

fabriziovigevani:
Now that we have available the `subscriptionModule` at the subgraph, we can now drop most of the contract calls (we must still use `getJurorShare`)  and improve performance.

This PR also includes changes on the way we request an account treasury balances and contemplates latest changes made by @facuspagnuolo here https://github.com/aragon/court-subgraph/pull/53

### DIFF
--- a/src/hooks/query-hooks.js
+++ b/src/hooks/query-hooks.js
@@ -21,7 +21,8 @@ export function useJurorDraftQuery(jurorId) {
 }
 
 /**
- * Queries if juror has ever claimed any rewards
+ * Queries if the juror  by id `jurorId` has already claimed rewards
+ * Rewards can be claimed from two places: Subscriptions fees or Dispute fees (the later includes appeal and juror fees)
  *
  * @param {String} jurorId Address of the juror
  * @returns {Boolean} True if juror has already claimed rewards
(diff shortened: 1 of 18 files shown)


src/hooks/query-hooks.js
```diff
@@ -21,27 +21,23 @@ export function useJurorDraftQuery(jurorId) {
...
     return false
   }
 
-  const {
-    drafts: rewardedDrafts,
-    feeMovements: claimedSubscriptionFeeMovements,
-  } = data.juror
-
-  return rewardedDrafts.length > 0 || claimedSubscriptionFeeMovements.length > 0
+  return data.feeMovements.length > 0
```
sohkai: Just wondering: this seem _too easy_ of an check. It seems to suggest that it returns true if the juror has _ever_ claimed a reward. Is that the intent (e.g. to change the copy of the rewards module in the dashboard to something else if they've previously claimed rewards)?

fabriziovigevani: Yes, that's exactly the intent. What do you mean by too easy ?

sohkai: In that case, it may be nice to rename the hook slightly, e.g. `useJurorRewardsEverClaimedQuery()`.

The current name suggests that they've claimed all their current rewards, which I would assume would involve much more logic :).

fabriziovigevani: Yes totally true!

PR #258, opened 2020-03-26
Dispute: detail component changes for precedence campaign

### LINKED ISSUE #260: Dispute: Precedence campaign changes needed
https://www.figma.com/file/BB1vIpBzf40ORO2O2XsnIG/Dispute-detail-changes?node-id=0%3A1

rperez89:
This PR includes all the change needed for the precedence campaign restart for the dispute detail component and the timeline.

Screen Shot 2020-03-26 at 1 33 52 AM

This closes #260

### DIFF
--- a/src/components/Disputes/DisputeInfo.js
+++ b/src/components/Disputes/DisputeInfo.js
@@ -4,17 +4,15 @@ import resolvePathname from 'resolve-pathname'
 import {
   Box,
   GU,
-  IconCheck,
-  IconCross,
   Link,
   TransactionBadge,
-  // isAddress,
   textStyle,
   useTheme,
 } from '@aragon/ui'
 import styled from 'styled-components'
 import DisputeActions from './DisputeActions'
 import DisputeCurrentRuling from './DisputeCurrentRuling'
+import DisputeOutcomeText from './DisputeOutcomeText'
 import DisputeStatus from './DisputeStatus'
 import DisputeVoided from './DisputeVoided'
 import ErrorLoading from '../Errors/ErrorLoading'
@@ -60,10 +58,6 @@ const DisputeInfo = React.memo(function({
 
   const isFinalRulingEnsured =
     phase === DisputePhase.ExecuteRuling || status === Status.Closed
-
-  // const lastRound = dispute?.rounds?.[dispute.lastRoundId]
-  // const appealedRuling = lastRound?.appeal?.appealedRuling
-  // const voteWinningOutcome = lastRound?.vote?.winningOutcome
 
   const isDisputeVoided = dispute?.status === Status.Voided
 
@@ -170,8 +164,7 @@ const DisputeInfo = React.memo(function({
 
 function DisputeHeader({ dispute, error }) {
   const theme = useTheme()
-  const { id, description } = dispute || {}
-  const transaction = dispute && dispute.txHash
+  const { id, description, transaction } = dispute || {}
 
   return (
     <div
@@ -207,7 +200,7 @@ function DisputeHeader({ dispute, error }) {
             width: 100%;
           `}
         >
-          <div
+          <h1
             css={`
               display: flex;
               align-items: center;
@@ -228,7 +221,7 @@ function DisputeHeader({ dispute, error }) {
                 `}
               />
             )}
-          </div>
+          </h1>
           {Boolean(dispute?.status !== Status.Voided && transaction) && (
             <TransactionBadge
               transaction={transaction}
@@ -263,7 +256,6 @@ function Field({ label, value }) {
             {transformAddresses(line, (part, isAddress, index) =>
               isAddress ? (
                 <span title={part} key={index}>
-                  {' '}
                   <LocalIdentityBadge
                     connectedAccount={addressesEqual(part, wallet.account)}
                     entity={part}
@@ -279,7 +271,7 @@ function Field({ label, value }) {
                       )
                       return (
                         <Link href={ipfsUrl} key={i}>
-                          {word}{' '}
+                          {word}
                         </Link>
                       )
                     }
@@ -306,10 +298,10 @@ function Field({ label, value }) {
 }
 
 function DisputeActionText({ dispute, isFinalRulingEnsured }) {
-  const theme = useTheme()
   const { disputedActionText, disputedActionURL } = dispute
   const lastRound = dispute?.rounds?.[dispute.lastRoundId]
   const voteWinningOutcome = lastRound?.vote?.winningOutcome
+  const appealedRuling = lastRound?.appeal?.appealedRuling
 
   if (!isFinalRulingEnsured) {
     return (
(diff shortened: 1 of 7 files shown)


src/components/Disputes/DisputeInfo.js
```diff
@@ -175,8 +168,9 @@ const DisputeInfo = React.memo(function({
   )
 })
 
-function DisputeHeader({ id, dispute, error }) {
+function DisputeHeader({ dispute, error }) {
   const theme = useTheme()
+  const { id, description } = dispute || {}
   const transaction = dispute && dispute.txHash
```
fabriziovigevani: Maybe we could combine this two lines

src/components/Disputes/DisputeInfo.js
```diff
@@ -209,17 +204,22 @@ function DisputeHeader({ id, dispute, error }) {
         <div
           css={`
             margin-left: ${3 * GU}px;
+            width: 100%;
           `}
         >
-          <h1
+          <div
```
fabriziovigevani: Was this changed for a reason? I tihnk we should keep using the `h1` here

src/components/Disputes/DisputeInfo.js
```diff
@@ -295,3 +348,28 @@ DisputeInfo.propTypes = {
...
+      ) : (
+        <div
+          css={`
+            ${textStyle('body2')};
+          `}
+        >
+          {value}
+        </div>
+      )}
+ */
```
fabriziovigevani: Reminder to remove here as well

src/components/Disputes/DisputeInfo.js
```diff
@@ -256,18 +256,42 @@ function Field({ label, value }) {
...
-          />
-        </div>
+
+      {typeof value === 'string' ? (
+        value.split('\n').map((line, i) => (
+          <React.Fragment key={i}>
+            {transformAddresses(line, (part, isAddress, index) =>
+              isAddress ? (
+                <span title={part} key={index}>
+                  {' '}
```
fabriziovigevani: Should we remove this?

src/components/Disputes/DisputeInfo.js
```diff
@@ -256,18 +256,42 @@ function Field({ label, value }) {
...
+                <React.Fragment key={index}>
+                  {transformIPFSHash(part, (word, isIpfsHash, i) => {
+                    if (isIpfsHash) {
+                      const ipfsUrl = resolvePathname(
+                        word,
+                        `${IPFS_ENDPOINT}/${word}`
+                      )
+                      return (
+                        <Link href={ipfsUrl} key={i}>
+                          {word}{' '}
```
fabriziovigevani: And also here ?

```suggestion
                          {word}
```

src/components/Disputes/DisputeInfo.js
```diff
@@ -281,9 +305,38 @@ function Field({ label, value }) {
...
+      css={`
+        display: flex;
+        color: ${voteWinningOutcome ? theme.positive : theme.negative};
+      `}
+    >
+      <Icon />
+      {prefix + disputedActionText}
+    </div>
+  )
+}
```
fabriziovigevani: What do you think of using the `DisputeOutcomeText` for this part? Since the logic and the rener structure is basically the same?

What we could do is add a to new props to the component `action` and `verbose`. If verbose is present we render the action (and we set it as `false` by default). 

And in this component we can do :

```
return (
    <DisputeOutcomeText
      action={disputedActionText}
      outcome={appealedRuling || voteWinningOutcome}
      phase={
        appealedRuling ? DisputePhase.AppealRuling : DisputePhase.RevealVote
      }
      verbose
    />
  )
```

Note that we also removed the `appealedRuling` outcome and we should keep this logic because it can happen that the dispute was appealed and nobody confirmed the appeal. In this case the final ruling is what the appealer's outcome

src/components/Disputes/DisputeInfo.js
rperez89: oh yes those are getting added when i save

src/components/Disputes/DisputeInfo.js
rperez89: good catch

src/components/Disputes/DisputeInfo.js
```diff
@@ -170,8 +164,7 @@ const DisputeInfo = React.memo(function({
 
 function DisputeHeader({ dispute, error }) {
   const theme = useTheme()
-  const { id, description } = dispute || {}
-  const transaction = dispute && dispute.txHash
+  const { id, description, transaction } = dispute || {}
```
fabriziovigevani: Ah i think in this case should be `txHash` now right?

src/hooks/useDisputes.js
```diff
@@ -157,8 +157,14 @@ function useProcessedDispute(dispute) {
           return setDisputeProcessed({
             ...dispute,
             description: parsedDisputeData.description || disputeDescription,
-            agreementText: agreementText || '',
+            disputedActionText: parsedDisputeData.disputedActionText || '',
+            disputedActionURL: parsedDisputeData.disputedActionURL || '',
+            agreementText:
+              parsedDisputeData.agreementTitle || agreementText || '',
```
sohkai: Just found this a bit weird; should we place the agreement title in a different field? Would this always be replacing the agreement text now?

src/components/Disputes/DisputeInfo.js
```diff
@@ -256,18 +249,41 @@ function Field({ label, value }) {
...
-          `}
-        >
-          <LocalIdentityBadge
-            connectedAccount={addressesEqual(value, wallet.account)}
-            entity={value}
-          />
-        </div>
+
+      {typeof value === 'string' ? (
+        value.split('\n').map((line, i) => (
```
sohkai: Perhaps we can move this bit all into its own component, something like `DisputeTextField`

src/lib/ipfs-utils.js
```diff
@@ -24,3 +27,9 @@ export const getIpfsCidFromUri = string => {
   }
   return ''
 }
+
+export function transformIPFSHash(str, callback) {
```
sohkai: Just wondering, are we using this?

src/hooks/useDisputes.js
fabriziovigevani: Yeah, in this case the `agreementText` that comes from ipfs metadata is the relative path to the ipfs content (and `agreementUrl ` is the resolved link) and the agreementTitle is the description. 

Maybe we could ask for `agreementText` to be renamed to `agreementUrl` at the metadata level

src/lib/ipfs-utils.js
fabriziovigevani: Yes we use it where we also use `transformAddresses `

src/hooks/useDisputes.js
sohkai: > Maybe we could ask for agreementText to be renamed to agreementUrl at the metadata level

Not sure on this—don't want to change it too much, but it just seems weird to set agreement**Title** for agreement**Text**.

PR #264, opened 2020-03-27
Precedence campaign banner: General fixes

fabriziovigevani:
From review at https://github.com/aragon/court-dashboard/pull/244


src/components/Disputes/PrecedenceCampaign/PrecedenceCampaignBanner.js
```diff
@@ -1,26 +1,40 @@
...
 import { Button, GU, useTheme } from '@aragon/ui'
+
 import Modal from './PrecedenceCampaignModal'
+import StoredList from '../../../StoredList'
+import { getNetworkType } from '../../../lib/web3-utils'
+
+const KEY_PREFIX = 'PCDisputeBannerRead'
 
-const KEY_PREFIX = 'PC_DISPUTE_BANNER_READ'
+function getStoredList() {
```
sohkai: This is not a big issue, but I was wondering if we could just instantiate this once.

We currently use this same pattern of creating a new one on every call in the client, but I'm not sure if there was a reason we needed to do so @bpierre?

bpierre: No I don’t think it is needed, we could probably do it once :+1: 

```jsx
const storedList = new StoredList(`${KEY_PREFIX}:${getNetworkType()}`)
```

fabriziovigevani: Ah yeah true, since this is account agnostic we can use the same one 👍

PR #266, opened 2020-03-27
Prevent the selection of illustrations

fabriziovigevani:


public/index.html
```diff
@@ -22,6 +22,9 @@
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
     <title>Aragon Court Dashboard</title>
+    <style>
+      body img {user-select: none;}
```
sohkai: I think we probably want to use styled-component's global stylesheet functionality instead—not sure if we already use it in this app.

fabriziovigevani: Ah wasn't aware of that feature 👍

PR #267, opened 2020-03-31
Dispute Detail: Show dispute action buttons if error from graph

fabriziovigevani:


src/components/Disputes/DisputeInfo.js
```diff
@@ -138,7 +142,7 @@ const DisputeInfo = React.memo(function({
             </>
           )
         })()}
-        {!isDisputeVoided && !error && (
+        {!isDisputeVoided && !error?.fromGraph && (
```
sohkai, line 145: Just to double check—this allows us to still show the actions when there are errors _except_ when it's the graphQL query errored?

fabriziovigevani: Yes exactly, the other possible error aside from graphQL is the error from IPFS and i don't think it makes sense to hide the actions when is only the metadata that couldn't be fetched.

PR #269, opened 2020-04-01
Mock data

fabriziovigevani:
This is an initial implementation to test the court without the burdens to setup everything locally or have a way to test different dispute phases as well as final rounds. 

TODO: 

- [ ] Add claimed rewards data
- [ ] Add total jurors stat
- [ ] General Clean up code
- [ ] Add comments

This implementation attempts to keep a consistent data across the dashboard so some contract logic had to be implemented :D 

### Considerations

- Currently there are about 11 disputes created, each one of them with a different state or different result. 

- There are 5 harcoded accounts that have participation in the court and these are the first 5 accounts from the default mnemonic that we use locally
 ```
myth like bonus scare over problem client lizard pioneer submit female collect
```

The first 3 accounts are jurors, the 4th is the appealer and the 5th the confirm appealer.

- Draft: For each round created we created drafts randomly between these 3 jurors, so on every page refresh you could see different results (jurors drafted for different disputes, dispute current ruling distribution changed, etc). What remains always unchanged are the dispute final results and  previous round results (this will always be outcome: Blocked Action, appeal: Allow Action, Confirm appeal: Blocked action).

- Since we get some data from the contracts itself,  this data is not available (e.g Subscription rewards, Total ANT staked, appeal deposits amount, etc)

- Actions are not possible here, it's read only for now.

### How to test

You can run it locally by `npm run start:local:mock`

I recommend using Rinkeby or Mainnet since running locally requires to run Ganache and possibly also ipfs to see the disputes content (we are using the recent App mining dispute metadata for all disputes)

So you can do 

```
npm run start:rinkeby:mock
#or
npm run start:mainnet:mock
```

### CODE REVIEW

File: src/queries/jurorDrafts.js [hunk 1]
```diff
@@ -40,9 +38,9 @@ export const JurorDraftRewarded = gql`
   }
 `
 
-// Jurors drafts by `id` for current term
-export const CurrentTermJurorDrafts = gql`
-  subscription JurorDrafts($id: ID!, $from: BigInt!) {
+// Jurors drafts for juror with id `$id` created from `$from`
```
sohkai, line 41:
```suggestion
// Jurors drafts for juror with id `$id` created since `$from`
```

src/utils/juror-draft-utils.js
```diff
@@ -33,7 +33,7 @@ export function transformJurorDataAttributes(jurorDraft) {
 
   return {
     ...jurorDraft,
-    weight: bigNum(weight),
```
sohkai, line 36: Just sanity checking—are we sure in these cases where we're not using a BN anymore?

src/exchanges.js
```diff
@@ -0,0 +1,30 @@
+import { fetchExchange, subscriptionExchange } from 'urql'
```
sohkai, line 1: Perhaps we can rename this file to be `graphql-exchanges`, to be clearer

src/index.js
```diff
@@ -43,6 +41,10 @@ subscriptionClient.onConnected(() => (connectionAttempts = 0))
 
 // Check for connection errors and if reaches max attempts send error log to Sentry
 subscriptionClient.onError(err => {
+  if (env('MOCK_DATA')) {
```
sohkai, line 44: Could we totally ignore subscriptions if we're in the mocked env?

src/components/Disputes/DisputeActions.js
```diff
@@ -187,6 +187,7 @@ function InformationSection({
 }
 
 // Helper function that returns main attributes for the YourVoteInfo component
+// TODO: Contemplate final round cases (when a juror has voted, the ANJ amount is pre-slashed)
```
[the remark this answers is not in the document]
fabriziovigevani: Ah no, here i meant to change the message that we display for jurors when for example they have voted in consensus, we currently display `Your ANJ balance will remain unlocked until...` so we should change this messages for final rounds.

src/utils/juror-draft-utils.js
fabriziovigevani: For regulars rounds the juror weight will be at most 81 (and that is if a juror get's all available slots of a round ,highly unlikely). For final rounds it will be the number of times the juror has the min active balance which it's also unlikely to be too big for it to not be supported.

src/index.js
fabriziovigevani: I'm not totally sure if we could ignore them completely because we are using a custom subscription exchange which fetches the data from the `mock` folder when we are in the mocked env. See https://github.com/aragon/court-dashboard/blob/b560acc7298656904a5853953eb3aa471dfe26af/src/mock/exchanges/index.js#L13 
If we were to discard using subscriptions then i think we would have to change all the `subscription-hooks` logic.

fabriziovigevani:
> Could we pull the src/mock files out into another PR that builds after this? There are a lot of uncontroversial changes here that are going to cause merge conflicts otherwise, and we should merge them in as soon as possible

Yep sure

> Do we require any network information when in the mocked state? Just the IPFS gateway and information from ETH nodes, but not the subgraph?

We don't require the subgraph information because on the mocked env we use our custom subscription exchange which will fetch our own mocked data (in `src/mock`) querying by the operation name.

src/utils/juror-draft-utils.js
sohkai: So now I'm wondering... I think we did this in a few other places to make the code simpler, but maybe we should align all the internal numbers to be BNs?

Otherwise it might get confusing later—`weight` is a normal number, but `round.number` is a string and `round.collectedTokens` is a BN. The string / BN difference can make sense (some are identifiers, some are meant for calculations), but juggling two number representations can be frustrating unless they're clearly delinated (e.g. suffixing `Bn` in variable names or [moving all the normal numbers to a separate field](https://github.com/aragon/aragon-apps/blob/master/apps/voting/app/src/app-state-reducer.js#L54)).

src/index.js
sohkai: Gotcha, that makes sense!

PR #270, opened 2020-04-03
Signer panel: auto-reveal flow

fabriziovigevani:
Includes general signer panel implementation and auto-reveal service request.

Once we have designs for the modal multi tx experience all we should have to do is change how we display the content. The logic shouldn't change much.

Seized the opportunity and made all court actions to use the signer panel as well.

Left TODO: 

- [x] Add court-server Mainnet and Rinkeby endpoints
- [x] Mitigate the issue with @facuspagnuolo of the auto-reveal service failing after certain dispute number. 
- [x] Add progress bar for txs waiting for confirmation

**Preview**

#### Single Tx (Same as the client):

Screen Shot 2020-04-22 at 1 23 56 AM

#### Multiple transactions / requests

Screen Shot 2020-04-22 at 1 22 39 AM


src/components/SignerPanel/SigningStatus.js
```diff
@@ -0,0 +1,290 @@
...
+  const theme = useTheme()
+
+  const isMultiTx = transactions.length > 1
+
+  return (
+    <div>
+      <div
+        css={`
+          background: ${theme.feedbackSurface};
+          height: 350px;
```
sohkai, line 40: Let's use GUs for heights and widths :)

src/components/SignerPanel/SigningStatus.js
```diff
@@ -0,0 +1,290 @@
...
+        flex-direction: column;
+        align-items: center;
+      `}
+    >
+      <div
+        css={`
+          display: flex;
+          align-items: center;
+          justify-content: center;
+          width: 60px;
```
sohkai, line 149: Let's use GUs :)

src/components/SignerPanel/SigningStatus.js
```diff
@@ -0,0 +1,290 @@
...
+  if (isMultiTx) {
+    return (
+      <Info>
+        {requiresMultipleSignatures ? (
+          <span>
+            This action requires multiple transactions. Please sign them one
+            after another and <b>do not close this window</b> until the process
+            is finished. Open your Ethereum provider {provider.name} to sign the
+            transactions.{' '}
+            <Link href="#">I don’t see the Ethereum provider.</Link>
```
sohkai, line 249: I don't think we have a great place to link this to, and I'm not really even sure what we would say in it :(.

Maybe let's remove it for now?

src/components/SignerPanel/SigningStatus.js
```diff
@@ -0,0 +1,290 @@
...
+  onClosePanel,
+  onNextAttempt,
+}) => {
+  return (
+    <>
+      <Info>
+        {maxAttemptsReached
+          ? `Seems that the transaction won't go through`
+          : `An error has occurred during the signature process. Don't worry, you can
+        try to send the transaction again.`}
```
sohkai, line 275: Nitpick: could we line up the indents here?

src/components/SignerPanel/TransactionStepItem.js
```diff
@@ -0,0 +1,174 @@
...
+        labelText: 'Transaction confirmed!',
+      }
+    }
+
+    if (status === TRANSACTION_STATUS_PENDING) {
+      return {
+        background: theme.info,
+        iconColor: theme.accentContent,
+        labelColor: theme.contentSecondary.alpha(0.6),
+        labelText: 'Transaction being mined...',
```
sohkai, line 57: Let's use the unicode `…` character instead of `...` where possible!

src/components/SignerPanel/TransactionStepItem.js
```diff
@@ -0,0 +1,174 @@
...
+    status === TRANSACTION_STATUS_CONFIRMED ||
+    (status === TRANSACTION_STATUS_SIGNED && !transaction.waitForConfirmation)
+
+  return (
+    <div
+      css={`
+        display: grid;
+        grid-template-columns: 20% 80%;
+        grid-gap: 16px;
+        width: 280px;
```
sohkai, line 89: Let's use GU for these values :)

src/components/SignerPanel/SigningStatus.js
```diff
@@ -0,0 +1,290 @@
+import React, { useMemo } from 'react'
```
sohkai, line 1: I see a lot of language here that's intended only for signing transactions, but what about the case where we have to make an API call, like in committing votes?

src/components/Disputes/panels/AppealPanel.js
```diff
@@ -77,34 +80,44 @@ function AppealPanel({
...
-          // TODO: some ERC20s don't let to set a new allowance if the current allowance is positive (handle this cases)
-          if (feeAllowance.eq(0)) {
-            console.warn('Allowance must be zero')
-          }
-          // Approve fee deposit for appealing
-          const approveTx = await onApproveFeeDeposit(requiredDeposit)
-          await approveTx.wait()
+      const transactionQueue = []
+      if (feeAllowance.lt(requiredDeposit)) {
+        // TODO: some ERC20s don't let to set a new allowance if the current allowance is positive (handle this cases)
```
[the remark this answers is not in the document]
fabriziovigevani: Good catch :)

src/services/requestAutoReveal.js
```diff
@@ -0,0 +1,42 @@
...
+      //    }
+      //    {
+      //      outcome: 'An outcome must be given'
+      //    },
+      //  ]
+      const errors = revealData.errors
+        ? revealData.errors.map(err => Object.values(err).join(', ')).join(', ')
+        : null
+
+      const errorMessage = revealData.error || errors
```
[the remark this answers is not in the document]
fabriziovigevani: Yes, initially i was thinking on checking on `res.ok` and throw if we got any error but we couldn't get the specifics errors in that way. Or perhaps we could do it this way and monitor possible errors on the server itself ?

fabriziovigevani:
> I'd also like to think a bit about the transaction queue's design. The comment below about encapsulation is in the direction of where I'd like to move—right now, the API feels too "loosely wired" and we could do a better job of hiding some of the details from the components that invoke the transactions.

Your concerns are valid! I think we can make the transaction creators to handle the logic of adding them to the queue as you  well suggested.

> The fact that we have to handle normal network requests rather than just transactions also makes the current design feel constrained compared to a more general design (e.g. an async request queue where we don't assume the async requests have anything to do with signing transactions, but could as well be).

Yes this is also true, we currently treat normal network requests also as transactions and we differentiate them from web3 txs by having a `skipSignature` prop. I agree is not the most clean an abstracted implementation. Will rethink how we can restructure this part :)

src/services/requestAutoReveal.js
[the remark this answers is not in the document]
sohkai: This sounds like we're not returning correct status codes on the backend. Let's ask Facu/Martynas about this—it looks like we're not documenting the error responses in [the court-backend](https://github.com/aragonone/court-backend/blob/development/packages/server/README.md#2-reveals) yet.

src/components/RequestPanel/TransactionProgress.js
```diff
@@ -0,0 +1,47 @@
...
+import { norm } from '../../lib/math-utils'
+import { getRelativeTime } from '../../utils/date-utils'
+
+const MINUTE = 60 * 1000
+const TX_DURATION_AVERAGE = 2 * MINUTE
+const TX_DURATION_THRESHOLD = TX_DURATION_AVERAGE - MINUTE / 2
+
+function TransactionProgress({ createdAt }) {
+  const theme = useTheme()
+  const now = useNow().valueOf()
```
sohkai, line 13: I believe we use `getTime()` in the client—let's keep doing that. Semantically, it's a bit more correct in intent (https://stackoverflow.com/a/9710267).

src/components/RequestPanel/ConfirmRequest.js
```diff
@@ -0,0 +1,63 @@
...
+          `}
+        >
+          Action{descriptions.length > 0 ? 's' : ''} to be triggered
+        </h4>
+        <div
+          css={`
+            color: ${theme.content};
+          `}
+        >
+          {descriptions.length > 1 ? (
```
sohkai, line 23: Would it be possible to see screenshots of each case? Or maybe we can jump on a call to demo this in a few scenarios quickly!

src/components/RequestPanel/RequestStatus.js
```diff
@@ -182,34 +168,34 @@ function SingleTx({ status }) {
   )
 }
 
-function SigningStatusInfo({
+function RequestStatusInfo({
```
sohkai, line 171: I find these changes afterwards to be more confusing—most of the time, we will be sending transactions so using solely the "request" language is odd.

Perhaps we can either:
- Use the `isTx` flag to help us build messages
- Generalize the language so we don't have to choose one or the other (e.g. "Everything was successful" vs. "Transaction/Request was successful")

src/components/RequestPanel/RequestStepItem.js
```diff
@@ -146,7 +145,7 @@ function TransactionStepItem({ hash, status, stepNumber, transaction }) {
             ${textStyle('title4')}
           `}
         >
-          {numberToOrdinal(stepNumber)} transaction
+          {capitalize(numberToOrdinal(stepNumber))} request
```
sohkai, line 148: Similar to above, I think this is a bit weird and we might want to re-think the copy.

src/services/requestAutoReveal.js
```diff
@@ -7,36 +7,30 @@ export default async (juror, disputeId, roundId, outcome, password) => {
...
+    }
 
-      return revealData
-    })
+    const response = await rawResponse.json()
+    const errors = response.errors
+      .map(err => Object.values(err).join(', '))
+      .join(', ')
+
+    throw new Error(`Failed to request auto-reveal service ${errors}`)
```
sohkai, line 31:
```suggestion
    throw new Error(`Failed to request auto-reveal service due to errors: ${errors}`)
```

src/components/RequestPanel/RequestPanel.js
```diff
@@ -0,0 +1,291 @@
...
+      }
+
+      createRequests()
+
+      return () => {
+        cancelled = true
+      }
+    },
+    // We'll remove `progress.processed` from the dependencies to prevent the effect from running every time we process a request
+    // Note that we'll retry processing requests form the last failed one only if the user requests to do so
```
sohkai, line 223:
```suggestion
    // Note that we'll retry processing requests from the last failed one only if the user requests to do so
```

src/components/RequestPanel/RequestPanel.js
```diff
@@ -0,0 +1,291 @@
...
+
+          try {
+            const requestProcess = await processRequest(request)
+
+            if (!cancelled) {
+              if (request.isTx && request.ensureConfirmation) {
+                await ensureRequestConfirmed(requestProcess)
+
+                // Execute callback function if required
+                if (request.callback) {
```
sohkai, line 199: Might as well do a quick `typeof === 'function` check

src/providers/RequestQueue.js
```diff
@@ -0,0 +1,51 @@
...
+    request => {
+      const newRequests = [...requestQueue, request]
+      return setRequestQueue(newRequests)
+    },
+    [requestQueue]
+  )
+
+  const addRequests = useCallback(
+    requests => {
+      const newRequests = [...requestQueue, ...requests]
```
sohkai, line 19: Note that with `array.concat()` you effectively can do both using the same API, so we could just have `addRequests()`.

src/components/Activity/activity-types.js
```diff
@@ -13,156 +14,112 @@ const ACTIVITY_TYPES = new Map(
...
+    claimRewards({ amount }) {
       return {
         title: 'Claim rewards',
         icon: iconClaimRewards,
-        description: `
-          Claim reward for round #${roundId} of dispute #${disputeId} for juror
-          ${jurorAddress}
-        `,
+        description: radspec.claimRewards(amount),
       }
```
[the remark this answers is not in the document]
fabriziovigevani: It doesn't assume the same arguments no, what name do you think best suits this functionality if not `radpsec` ? Maybe `describe` or something around that ?

src/components/Dashboard/RewardsModule.js
```diff
@@ -358,16 +309,6 @@ function TotalFees({ totalFees, requiresMultipleTxs }) {
...
-        {requiresMultipleTxs && (
-          <Info
-            css={`
-              margin-top: ${2 * GU}px;
-            `}
-          >
-            This action requires multiple transactions to be signed in{' '}
-            {provider.name}. Please confirm them one after another.
-          </Info>
-        )}
```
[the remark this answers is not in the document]
fabriziovigevani: Yes, this is going to be shown at the signer panel now.

src/components/Dashboard/panels/ANJForm.js
```diff
@@ -102,14 +102,8 @@ const ANJForm = React.memo(function ANJForm({
...
-    try {
-      const tx = await onSubmit(amount.valueBN)
-
-      onDone()
-      await tx.wait()
-    } catch (err) {
-      console.error('Error submitting tx: ', err) // TODO: How should we handle errors ?
-    }
+    onDone()
+    onSubmit(amount.valueBN)
```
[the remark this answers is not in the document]
fabriziovigevani: True! I think i might forgot to remove it from every submit handler.

src/components/Disputes/actions/DisputeDraft.js
```diff
@@ -9,12 +9,7 @@ function DisputeDraft({ disputeId, onDraft }) {
     async event => {
       event.preventDefault()
 
-      try {
-        const tx = await onDraft(disputeId)
-        await tx.wait()
-      } catch (err) {
-        console.error('Error submitting tx: ', err)
-      }
+      return onDraft(disputeId)
```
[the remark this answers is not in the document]
fabriziovigevani: It's definitely not needed. Removed them along with the `async`

src/radspec.js
```diff
@@ -0,0 +1,97 @@
...
+      `
+  },
+  settleAppealDeposit: (roundId, disputeId) => {
+    return `
+        Settle appeal deposit for round ${numberToWord(
+          roundId
+        )} of dispute #${disputeId}
+      `
+  },
+  transaction: transactionHash => {
```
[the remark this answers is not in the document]
fabriziovigevani: Ah no we don't, i should have removed it when copying all activity types.

src/radspec.js
```diff
@@ -0,0 +1,97 @@
...
+    return `
+        Compute the final ruling for dispute #${disputeId}
+      `
+  },
+  heartbeat: transitions => {
+    return `
+        Transition ${transitions} court term${transitions > 1 ? 's' : ''}
+      `
+  },
+  leakVote: (voteId, voter) => {
```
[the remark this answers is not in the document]
fabriziovigevani: No, not yet. This is just a placeholder for when we actually build it.

src/components/RequestPanel/ConfirmRequest.js
fabriziovigevani: Ah Yes! I will upload some screenshots and can do a quick demo in the standup tomorrow

src/components/RequestPanel/TransactionProgress.js
fabriziovigevani: Ah actually this is a `dayjs` instance  and it [uses `getTime` under the hood](https://github.com/iamkun/dayjs/blob/bcea06730da2d19ddbad844de3557ee077af821c/src/index.js#L158)

src/components/Activity/activity-types.js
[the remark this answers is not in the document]
sohkai: Wondering if we could compromise on the description here so that they can eventually be made radspec descriptions?

src/components/RequestPanel/RequestPanel.js
```diff
@@ -196,7 +196,7 @@ function RequestPanel() {
                 await ensureRequestConfirmed(requestProcess)
 
                 // Execute callback function if required
-                if (request.callback) {
+                if (typeof request?.callback === 'function') {
```
sohkai, line 199: Oh this is interesting, what does this do if `request` is null? Would it do a `typeof null`?

fabriziovigevani: Ah actually i think we don't need the `?` since we are iterating on the requests. a `typeof request.callback === 'function'` should do. If there's no callback it will be `typeof undefined === 'function'`

src/components/RequestPanel/RequestStatus.js
```diff
@@ -0,0 +1,282 @@
...
+  onClosePanel,
+  onNextAttempt,
+}) => {
+  return (
+    <>
+      <Info>
+        {maxAttemptsReached
+          ? `Seems that the request can't be processed`
+          : `An error has occurred during the request process. Don't worry, you can
+             try to process the request again.`}
```
sohkai, line 267:
```suggestion
          : `An error occurred at this step. Don't worry, you can try to process
             this step again.`}
```
We could probably improve this... but let's not worry about it for now.

src/components/RequestPanel/RequestStatus.js
```diff
@@ -0,0 +1,282 @@
...
+        lastProcessedAt={lastProcessedAt}
+        status={status}
+        stepNumber={index + 1}
+        request={request}
+      />
+    )
+  })
+}
+
+function SingleRequest({ request, status }) {
```
sohkai, line 84: More of a nitpick, but what do you think about reorganizing the code here a bit:

- Moving `MultiStepRequest` as its own component
- Moving `SingleRequest` to a `SingleStepRequest` component
- Putting both of these components into a `RequestSteps` wrapper component + file
- Migrating `RequestStepItem` to be an internal component inside this file

We might able be able to take the opportunity to make a pure function that gives us the label text based on the request given, to dedupe some of the strings.

src/components/RequestPanel/RequestPanel.js
```diff
@@ -0,0 +1,291 @@
...
+          try {
+            const requestProcess = await processRequest(request)
+
+            if (!cancelled) {
+              if (request.isTx && request.ensureConfirmation) {
+                await ensureRequestConfirmed(requestProcess)
+
+                // Execute callback function if required
+                if (typeof request?.callback === 'function') {
+                  request.callback()
```
sohkai, line 200: Will these callback functions ever change a component's state?

If not, I think we should still run them even if we've unmounted (but obviously not run any new ones).

src/components/RequestPanel/RequestPanel.js
```diff
@@ -0,0 +1,291 @@
...
+                }
+              }
+            }
+          } catch (err) {
+            console.error(`Error when executing: ${request.description}:`, err)
```
sohkai, line 205:
```suggestion
            console.error(`Error when executing '${request.description}':`, err)
```

src/components/RequestPanel/RequestPanel.js
```diff
@@ -0,0 +1,291 @@
...
+            captureException(err)
+
+            if (requests.length > 1) {
```
sohkai, line 208: Perhaps we can always just re-throw here to force us to exit.

Using the exception seems natural as it's a native "escape hatch" to normal control flow.

src/components/RequestPanel/ConfirmRequest.js
```diff
@@ -0,0 +1,63 @@
...
+                </li>
+              ))}
+            </ul>
+          ) : (
+            <span>{descriptions[0]}</span>
+          )}
+        </div>
+      </Info>
+      <Button
+        label="Create request"
```
sohkai, line 51:
```suggestion
        label="Confirm"
```

src/components/RequestPanel/RequestStatus.js
```diff
@@ -0,0 +1,282 @@
...
+        {requiresMultipleSignatures ? (
+          <span>
+            This action requires multiple transactions. Please sign them one
+            after another and <b>do not close this window</b> until the process
+            is finished. Open your Ethereum provider {provider.name} to sign the
+            transactions.{' '}
+          </span>
+        ) : (
+          <span>
+            This action requires multiple steps. Please,{' '}
```
sohkai, line 238:
```suggestion
            This action involves multiple steps. Please,{' '}
```

src/components/RequestPanel/RequestPanel.js
fabriziovigevani: They are not supposed to change component state no. 
They are meant as functions to execute after we ensure that the tx has been confirmed e.g saving one time code to local storage.

From convo offline we decided to remove the cancel condition here

src/hooks/useCourtContracts.js
```diff
@@ -329,14 +473,86 @@ export function useRewardActions() {
...
+      // Claim subscription fees
+      for (const subscriptionFee of subscriptionFees) {
+        requestQueue.push({
+          intent: () => claimFees(subscriptionFee.periodId),
+          description: radspec.claimSubscriptionFees(subscriptionFee.periodId),
+          isTx: true,
+        })
+      }
+
+      return addRequests(requestQueue)
```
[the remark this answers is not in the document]
fabriziovigevani: Good question haha, it was overflowing in these cases. Updated the div component  accordingly.

src/components/RequestPanel/RequestStatus.js
fabriziovigevani: Much cleaner 🥇

PR #277, opened 2020-04-14
Settings: Subgraph endpoints configuration

### LINKED ISSUE #273: Network Preferences: Add subgraph endpoint into the network preferences

rperez89:
This PR closes #273 

We also need to handle the subscriptions ws endpoint change since we handle two types of connections with the graph `getFetchExchange` and  `getSubscriptionExchange`.
If the WS endpoint change we need to close the previous connection and open a new one.

Screen Shot 2020-04-13 at 10 55 51 PM


src/providers/Subgraph.js
```diff
@@ -0,0 +1,43 @@
...
+  return (
+    <SubgraphContext.Provider
+      value={{
+        resetSubgraphClient: () => setClient(newClient(true)),
+      }}
+    >
+      <UrqlProvider value={client}>{children}</UrqlProvider>
+    </SubgraphContext.Provider>
+  )
+}
```
fabriziovigevani, line 39: Since we are reloading the window on every network change, is this provider really needed ?

rperez89: ahh i though that would be better to have an specific provider for that with an specific action instead of just relying on the web page reload, is there any specific reason why you want to remove it?

fabriziovigevani: Yes i agree that having it encapsulated in a component looks cleaner, just was thinking that maybe the logic of resetting it and saving the client as a state wasn't needed because we are reloading the page any ways and i don't think we'll be resetting it from anywhere else except from the Network settings

src/lib/web3-utils.js
```diff
@@ -10,6 +10,7 @@ export const ETH_FAKE_ADDRESS = `0x${''.padEnd(40, '0')}`
 const ETH_ADDRESS_SPLIT_REGEX = /(0x[a-fA-F0-9]{40}(?:\b|\.|,|\?|!|;))/g
 const ETH_ADDRESS_TEST_REGEX = /(0x[a-fA-F0-9]{40}(?:\b|\.|,|\?|!|;))/g
 const httpRegex = /^(https?):\/\//
```
sohkai, line 12:
```suggestion
const httpRegex = /^https?:\/\//
```

Wondering if there was a reason we had the regex grouping previously? It's not a very big deal, but I find this easier to read (simpler and we don't care about any matching groups) and it's more performant (no need to match and create groups).

src/environment.js
```diff
@@ -34,6 +34,12 @@ const ENV_VARS = {
   IPFS_GATEWAY() {
     return process.env.REACT_APP_IPFS_GATEWAY || ''
   },
+  CUSTOM_HTTP_SUBGRAPH_ENDPOINT() {
+    return process.env.REACT_APP_CUSTOM_HTTP_SUBGRAPH_ENDPOINT || ''
+  },
+  CUSTOM_WS_SUBGRAPH_ENDPOINT() {
+    return process.env.REACT_APP_CUSTOM_WS_SUBGRAPH_ENDPOINT || ''
+  },
```
sohkai, line 42: Nitpick: I might alphabetize this file; right now it's a bit hard to find the exact key I'm looking for.

Also curious—should we prefer providing the `endpoint()` defaults here instead of the `local-settings`? That might also allow us to simplify / align the env variable names to just `SUBGRAPH_HTTP_ENDPOINT` and `SUBGRAPH_WS_ENDPOINT`.

src/components/GlobalPreferences/Network/Network.js
```diff
@@ -9,35 +9,50 @@ import {
...
+    subgraphHttpEndpoint,
+    subgraphWsEndpoint,
     handleEthNodeChange,
     handleIpfsGatewayChange,
-    networkError,
+    handleSubgraphHttpEndpointChange,
+    handleSubgraphWsEndpointChange,
+    settingsErrors,
     handleNetworkChange,
     handleClearNetworkSettings,
```
sohkai, line 49: Nitpick: we might as well alphabetize this destructor :)

src/components/GlobalPreferences/Network/Network.js
```diff
@@ -89,6 +104,50 @@ function Network() {
...
+              css={`
+                ${textStyle('body4')};
+                color: ${theme.negative};
+              `}
+            >
+              {httpSubgraphError}
+            </span>
+          )}
+        </Label>
+        <Label theme={theme}>
```
sohkai, line 129: We're duplicating these same sets of components multiple times... so might be a good opportunity to create a new component :).

src/components/GlobalPreferences/Network/Network.js
```diff
@@ -89,6 +104,50 @@ function Network() {
             `}
           />
         </Label>
+        <Label theme={theme}>
+          Subgraph HTTP Endpoint
```
sohkai, line 108:
```suggestion
          Subgraph HTTP endpoint
```

src/components/GlobalPreferences/Network/Network.js
```diff
@@ -89,6 +104,50 @@ function Network() {
...
+                ${textStyle('body4')};
+                color: ${theme.negative};
+              `}
+            >
+              {httpSubgraphError}
+            </span>
+          )}
+        </Label>
+        <Label theme={theme}>
+          Subgraph WS Endpoint
```
sohkai, line 130:
```suggestion
          Subgraph WS endpoint
```

src/local-settings.js
```diff
@@ -1,8 +1,11 @@
 import env from './environment'
-import { IPFS_ENDPOINT } from './endpoints'
+import endpoints, { IPFS_ENDPOINT } from './endpoints'
 
 const DEFAULT_ETH_NODE = 'DEFAULT_ETH_NODE'
 const IPFS_GATEWAY = 'IPFS_GATEWAY'
+const CUSTOM_HTTP_SUBGRAPH_ENDPOINT = 'CUSTOM_HTTP_SUBGRAPH_ENDPOINT'
+const CUSTOM_WS_SUBGRAPH_ENDPOINT = 'CUSTOM_WS_SUBGRAPH_ENDPOINT'
+const [GRAPH_API_HTTP_ENDPOINT, GRAPH_API_WS_ENDPOINT] = endpoints()
```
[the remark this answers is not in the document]
rperez89: not sure if i understood this one

src/lib/web3-utils.js
rperez89: ahh before before i was doing (http || https)

src/local-settings.js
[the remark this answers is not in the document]
sohkai: It felt odd that we were using the structure `CUSTOM_<X>_SUBGRAPH_ENDPOINT` for the localstorage keys but `GRAPH_API_<X>_ENDPOINT` for the values.

src/components/GlobalPreferences/Network/Network.js
```diff
@@ -265,13 +222,26 @@ const useNetwork = () => {
...
     await checkValidEthNode(ethNode)
   } catch (err) {
-    settingsErrors.ethError = err
+    if (err instanceof InvalidNetworkType) {
+      settingsErrors.ethError = `Node must be connected to ${sanitizeNetworkType(
+        networkType
+      )}`
+    }
+    if (err instanceof InvalidURI) {
+      settingsErrors.ethError = 'Must provide Http endpoint to node'
```
sohkai, line 238:
```suggestion
      settingsErrors.ethError = 'Must provide HTTP endpoint to node'
```

src/components/GlobalPreferences/Network/Network.js
```diff
@@ -239,9 +239,6 @@ async function validateNetworkSettings(
     }
     if (err instanceof NoConnection) {
       settingsErrors.ethError = 'Could not connect to node'
-    } else {
-      settingsErrors.ethError = 'URI does not seem to be a ETH node'
```
sohkai: To make sure—was this intended to be removed?

rperez89: ah yes i realized that we are capturing the error with the NoConnection one

src/providers/Subgraph.js
sohkai: Coming back to this, these lines definitely seem like they aren't necessary since we're just refreshing the page later:

https://github.com/aragon/court-dashboard/pull/277/files#diff-9eff867218c8af415bc22e0835ba9dd7R181-R183

PR #279, opened 2020-04-14
Settings: FAQ button

rperez89:
Screen Shot 2020-04-14 at 5 05 16 PM

### DIFF
--- a/src/components/GlobalPreferences/GlobalPreferencesButton.js
+++ b/src/components/GlobalPreferences/GlobalPreferencesButton.js
@@ -85,34 +85,36 @@ function GlobalPreferencesButton({ onOpen }) {
           </li>
           <Item
             onClick={handleItemClick('network')}
-            iconFirst
-            icon={iconNetwork}
+            Icon={() => <img src={iconNetwork} alt="" />}
             label="Network"
           />
-          <Item
-            iconFirst={false}
-            icon="IconExternal"
-            label="Any questions? Visit our FAQ"
-            href="https://help.aragon.org/article/48-aragon-court-faq"
-          />
+          <Item href="https://help.aragon.org/article/48-aragon-court-faq">
+            <div
+              css={`
+                flex-grow: 1;
+                display: flex;
+                align-items: center;
+                margin-left: ${1 * GU}px;
+              `}
+            >
+              Any questions? Visit our FAQ
+            </div>
+            <div
+              css={`
+                color: ${theme.surfaceContentSecondary};
+              `}
+            >
+              <IconExternal />
+            </div>
+          </Item>
         </ul>
       </Popover>
     </React.Fragment>
   )
 }
 
-function Item({ icon, iconFirst, label, onClick, href }) {
+function Item({ children, Icon, label, onClick, href }) {
   const theme = useTheme()
-  let Icon
-
-  switch (icon) {
-    case 'IconExternal':
-      Icon = IconExternal
-      break
-    default:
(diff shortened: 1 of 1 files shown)


src/components/GlobalPreferences/GlobalPreferencesButton.js
```diff
@@ -84,17 +85,34 @@ function GlobalPreferencesButton({ onOpen }) {
           </li>
           <Item
             onClick={handleItemClick('network')}
+            iconFirst
             icon={iconNetwork}
             label="Network"
           />
+          <Item
+            onClick={handleItemClick('network')}
```
fabriziovigevani, line 93: I think this one should be removed ?

rperez89: ahh yes good catch!

rperez89:
> I was thinking that maybe we can throw the `iconFirst`prop and differentiate whether we need to add the external icon or not by the `href`. And still tell if we should add the icon first just by `icon`
> 
> What do you think?

mmm not sure about this one, we might  have another preferences button that the icon goes to the right and is not an external one

src/components/GlobalPreferences/GlobalPreferencesButton.js
```diff
@@ -84,17 +85,33 @@ function GlobalPreferencesButton({ onOpen }) {
...
+            href="https://help.aragon.org/article/48-aragon-court-faq"
+          />
         </ul>
       </Popover>
     </React.Fragment>
   )
 }
 
-function Item({ icon, label, onClick }) {
+function Item({ icon, iconFirst, label, onClick, href }) {
```
sohkai, line 104: Rather than exposing an API with `iconFirst`, what do you think about exposing `children` instead?

We've used this pattern a few times; with that you can opt for the "easy" pre-defined layout, or customize the internals completely by using children.

src/components/GlobalPreferences/GlobalPreferencesButton.js
```diff
@@ -84,17 +85,33 @@ function GlobalPreferencesButton({ onOpen }) {
...
   const theme = useTheme()
+  let Icon
+
+  switch (icon) {
```
sohkai, line 108: Let's update the old API as we find new edge cases :).

Rather than doing this, let's force users to pass down the node they'd like to render in the icon slot (e.g. users would pass down `IconExternal` or ``).

sohkai:
@rperez89 I've added a few commits to fix the spacing and revert the `Item`'s API to be the old one.

If you look at the ["removed whitespace" diff](https://github.com/aragon/court-dashboard/pull/279/files?diff=unified&w=1) you'll see that we didn't have to change very much at all to add this "exception"!

rperez89:
> @rperez89 I've added a few commits to fix the spacing and revert the `Item`'s API to be the old one.
> 
> If you look at the ["removed whitespace" diff](https://github.com/aragon/court-dashboard/pull/279/files?diff=unified&w=1) you'll see that we didn't have to change very much at all to add this "exception"!

Hi! not sure why we need that margin? and not sure why we reverted the Icon to the old one?

rperez89:
> @rperez89 I've added a few commits to fix the spacing and revert the `Item`'s API to be the old one.
> 
> If you look at the ["removed whitespace" diff](https://github.com/aragon/court-dashboard/pull/279/files?diff=unified&w=1) you'll see that we didn't have to change very much at all to add this "exception"!

yeah! agree. Was a simple but nice exercise to do

PR #295, opened 2020-04-27
Notifications: Signature request 

rperez89:
This PR can be merged since this component is not being referenced from anywhere

78e5e3c536f797eb47c15825d28c4880ddddb762

Screen Shot 2020-04-27 at 3 05 34 PM

Screen Shot 2020-04-27 at 3 09 16 PM

Screen Shot 2020-04-27 at 3 06 53 PM

82b8597182b66a9661a53b124a8e3d37e004be3e

Screen Shot 2020-04-27 at 3 06 19 PM


src/components/SignatureRequest.js
```diff
@@ -0,0 +1,182 @@
...
+        return
+      }
+      const now = dayjs().toString()
+      const { signHash, error } = await signMessage(wallet, now)
+
+      if (error) {
+        setSigningError(true)
+        return
+      }
+      const { error: actionError } = await onSignSuccessAction()
```
fabriziovigevani, line 48: An example of this action would be verifying the email right?

src/components/SignatureRequest.js
```diff
@@ -0,0 +1,182 @@
...
+            `}
+          >
+            <ActionButtons onClick={() => {}} width={buttonWidth}>
+              Troubleshooting
+            </ActionButtons>
+            <ActionButtons
+              mode="strong"
+              onClick={() => {
+                setSigningError(false)
+              }}
```
fabriziovigevani, line 132: Maybe we could have a `useCallback` above (`handleRepeatSignature` or similar) to prevent re creating the function on every render?

src/lib/web3-utils.js
```diff
@@ -196,3 +196,18 @@ export async function checkValidEthNode(uri) {
...
+export async function signMessage(wallet, message) {
+  let signHash
+  let error = false
+  let errorMsg
+
+  try {
+    signHash = await wallet.ethers.getSigner().signMessage(message)
+  } catch (err) {
+    error = true
+    errorMsg = err
```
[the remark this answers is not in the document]
rperez89: no because we are not showing on the designs the error but given that we have severals possible errros at some point we would want to show the error

src/components/SignatureRequest.js
```diff
@@ -0,0 +1,182 @@
...
+            css={`
+              display: flex;
+              justify-content: space-between;
+              width: 100%;
+              flex-direction: ${compactMode ? 'column' : 'row'};
+              margin-bottom: ${1.5 * GU}px;
+              margin-top: ${3 * GU}px;
+            `}
+          >
+            <ActionButtons onClick={() => {}} width={buttonWidth}>
```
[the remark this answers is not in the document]
rperez89: sure

src/components/SignatureRequest.js
```diff
@@ -0,0 +1,182 @@
...
+}
+
+function getInfoText(error, signHash, successText, actionText) {
+  if (error) {
+    return 'An error has occurred when signing the message.'
+  }
+  if (signHash) {
+    return successText
+  }
+  return `Open your Ethereum provider (Metamask or similar) to complete the
```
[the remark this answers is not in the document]
rperez89: that is exactly how they are on the designs, not sure if we should change it cc @dizzypaty

src/components/SignatureRequest.js
rperez89: yes, is the action that should be executed after the sign

src/components/SignatureRequest.js
rperez89: could you provide an example? not sure if i follow the idea

fabriziovigevani: Havinig like

```
const handleRepeatSignature = useCallback(() => setSigningError(false))
```

And passing this function to the `onClick`

src/lib/web3-utils.js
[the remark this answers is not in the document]
fabriziovigevani: In that case, does it make sense to modify `error` to be the error message and not a boolean, and also drop `errorMsg`? I'm thinking this because looks like we can indicate whether there was en error or not directly from the error returned value. Returning both `error` and `errorMsg` feels like having duplicated logic.

src/components/SignatureRequest.js
```diff
@@ -0,0 +1,186 @@
...
+  signature request. Signing this message will prove ownership of your
+  account and ${actionText}`
```
sohkai, line 179:
```suggestion
  account and ${actionText}.`
```

src/components/SignatureRequest.js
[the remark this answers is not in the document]
sohkai: Yes, I think we should change it, as that's what we do elsewhere when we need a provider name.

src/components/SignatureRequest.js
```diff
@@ -0,0 +1,186 @@
+import React, { useEffect, useState } from 'react'
+import styled from 'styled-components'
+import { Button, GU, textStyle, useTheme } from '@aragon/ui'
+import { useWallet } from '../providers/Wallet'
+import { signMessage } from '../lib/web3-utils'
```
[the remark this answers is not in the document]
rperez89: ah, the lib one i think that is more mapped with real libs like web-, ipfs but we can remove that folder if you want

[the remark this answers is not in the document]
sohkai: Gotcha, it did seem like the utils was more directly related to the court? But yes, might be good to just use one for now, since we have other utils like `date-utils` in `/utils`

src/components/SignatureRequest.js
```diff
@@ -0,0 +1,186 @@
...
+import signRequestIllustration from '../../src/assets/signRequest.svg'
+import signRequestSuccessIllustration from '../../src/assets/signRequestSuccess.svg'
+import signRequestFailIllustration from '../../src/assets/signRequestFail.svg'
+
+const SignerRequest = React.memo(function SignerRequest({
+  actionText,
+  compactMode,
+  onActionErrorScreen,
+  onActionSuccessScreen,
+  onSignSuccessAction,
```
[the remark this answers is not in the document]
rperez89: Yes, that is something that i have been thinking to change because i am repeating this pattern in another components where might worth just to handle everything on the parent

src/components/SignatureRequest.js
```diff
@@ -0,0 +1,186 @@
...
+      <div
+        css={`
+          display: flex;
+          flex-direction: column;
+          text-align: center;
+          align-items: center;
+        `}
+      >
+        <img src={illustration} height={140} width={140} alt="" />
+        <span
```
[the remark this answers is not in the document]
rperez89: anyways we need to apply the ` ${textStyle('title2')};` right?

src/components/SignatureRequest.js
```diff
@@ -0,0 +1,180 @@
...
+import styled from 'styled-components'
+import { Button, GU, textStyle, useTheme } from '@aragon/ui'
+import { useWallet } from '../providers/Wallet'
+import { getProviderFromUseWalletId } from '../ethereum-providers'
+import { signMessage } from '../lib/web3-utils'
+import { dayjs } from '../utils/date-utils'
+
+import signRequestSuccessIllustration from '../../src/assets/signRequestSuccess.svg'
+import signRequestFailIllustration from '../../src/assets/signRequestFail.svg'
+import signProcessing from '../../src/assets/loader.gif'
```
sohkai, line 11: Total nitpick, but let's use a better file name!

src/components/SignatureRequest.js
```diff
@@ -0,0 +1,180 @@
...
+    }
+    return {
+      statusText: 'Waiting for signature…',
+      statusTextColor: theme.surfaceContentSecondary,
+    }
+  }, [signingError, signHash, theme])
+
+  const infoText = useMemo(() => {
+    if (signingError) {
+      return 'An error has occurred when signing the message.'
```
sohkai, line 58:
```suggestion
      return 'A error occurred when signing the message.'
```

src/components/SignatureRequest.js
```diff
@@ -0,0 +1,180 @@
...
+    }
+    if (signHash) {
+      return successText
+    }
+    return `Open ${provider.name} to complete the
+    signature request. Signing this message will prove ownership of your
+    account and ${actionText}`
```
sohkai, line 65:
```suggestion
    return `Open ${provider.name} to complete the
      signature request. Signing this message will prove ownership of your
      account and ${actionText}.`
```

src/components/SignatureRequest.js
```diff
@@ -0,0 +1,180 @@
...
+  }, [actionText, provider, signingError, signHash, successText])
+
+  const requestSignature = useCallback(async () => {
+    if (!wallet) {
+      return
+    }
+    const now = dayjs().toString()
```
sohkai, line 72: Hmm, we may want to be careful here. Which timezone is this formatted as? We should be either using UTC time or unix time when signing and sending to the backend.

(Something we can edit afterwards, since we're not using this yet)

src/components/SignatureRequest.js
```diff
@@ -0,0 +1,180 @@
...
+    if (typeof onSignSuccess === 'function') {
+      await onSignSuccess(signHash, now)
+
+      setSignHash(signHash)
+    }
+  }, [onSignSuccess, wallet])
+
+  useEffect(() => {
+    requestSignature()
+  }, []) // eslint-disable-line react-hooks/exhaustive-deps
```
sohkai, line 89: Let's leave the comment about why we're running this once on render

src/components/SignatureRequest.js
```diff
@@ -0,0 +1,180 @@
...
+
+    if (error) {
+      setSigningError(true)
+      return
+    }
+
+    if (typeof onSignSuccess === 'function') {
+      await onSignSuccess(signHash, now)
+
+      setSignHash(signHash)
```
sohkai, line 83: Shouldn't we do this irrespective of `onSignSuccess()`?

src/components/SignatureRequest.js
```diff
@@ -0,0 +1,180 @@
...
+    }
+  }, [onSignSuccess, wallet])
+
+  useEffect(() => {
+    requestSignature()
+  }, []) // eslint-disable-line react-hooks/exhaustive-deps
+
+  useEffect(() => {
```
sohkai, line 91: I still get the sense this should be done in the parent component; it can decide in its `onSignSuccess()` how to handle the `onActionSuccess()`.

src/components/SignatureRequest.js
```diff
@@ -0,0 +1,180 @@
...
+const SignerRequest = React.memo(function SignerRequest({
+  actionText,
+  compactMode,
+  onActionSuccess,
+  onSignSuccess,
+  successText,
+  title,
+}) {
+  const [signingError, setSigningError] = useState(false)
+  const [signHash, setSignHash] = useState('')
```
sohkai, line 22: Perhaps we can use a boolean for this state instead, and use a name like `signedHash`, since we never need the actual signature in this component.

src/components/SignatureRequest.js
rperez89: signLoadingRing.gif? processingSignRing.gif?

src/components/SignatureRequest.js
```diff
@@ -0,0 +1,180 @@
...
+    }
+    const now = dayjs().toString()
+    const { signHash, error } = await signMessage(wallet, now)
+
+    if (error) {
+      setSigningError(true)
+      return
+    }
+
+    if (typeof onSignSuccess === 'function') {
```
sohkai, line 80: `onSignSuccess()` is really meant to be a required prop; this component doesn't work at all if it's not provided (why sign a message if there's nothing to give it back to?).

Although we're not using prop types, it still seems reasonable for us to drop the typecheck here to make it more explicit that we expect this to be passed as a prop.

src/components/SignatureRequest.js
rperez89: ahh i can do it now, good catch

src/components/SignatureRequest.js
rperez89: yes we are using it https://github.com/aragon/court-dashboard/blob/90a4d954960d1a47bee4ea0dc656ac70b470721d/src/components/SignatureRequest.js#L81

src/components/SignatureRequest.js
rperez89: no, because if something failes on the signSucces  the parent will change to another screen, if we are there is because everything went ok and we can show the signing success with the configs provided

src/components/SignatureRequest.js
sohkai: What do you think about `signRequestLoading.gif`, so it's similar to the other illustration names?

src/components/SignatureRequest.js
sohkai: But there we're not using the state variable... we're using the local variable in the function closure.

rperez89: ahhh hah true

src/lib/web3-utils.js
```diff
@@ -196,3 +196,18 @@ export async function checkValidEthNode(uri) {
...
   return true
 }
+
+export async function signMessage(wallet, message) {
+  let signHash
+  let error = false
+  let errorMsg
+
+  try {
+    signHash = await wallet.ethers.getSigner().signMessage(message)
```
sohkai, line 206: I know we discussed this earlier, but on second thought, what was the motivation behind catching the error here and returning it rather than letting it bubble up to the user?

src/components/SignatureRequest.js
```diff
@@ -51,56 +48,38 @@ const SignerRequest = React.memo(function SignerRequest({
       statusText: 'Waiting for signature…',
       statusTextColor: theme.surfaceContentSecondary,
     }
-  }, [signingError, signHash, theme])
+  }, [signingError, successMode, theme])
 
   const infoText = useMemo(() => {
     if (signingError) {
-      return 'An error has occurred when signing the message.'
+      return 'A error occurred when signing the message.'
```
sohkai, line 55: This is an English nitpick, but `An error` would be more correct :).

PR #297, opened 2020-04-27
Onboarding modal: prevent error when trying to close

bpierre:
This leaves `onClose` as `undefined` rather than `null`, which makes `Modal` use its default rather than trying to call `null`.

Also removes the close button since `onClose` is never set.


bpierre:
> I remember we also had some fixes for this in aragonUI—do we still need to release to get those updates?

Oh yes, that would also have fixed the issue: https://github.com/aragon/aragon-ui/pull/724

But in any case, we better not set the prop rather than setting it to `null`.

PR #300, opened 2020-04-28
Notifications: Verify email address screen

rperez89:
This PR can be merged since is not being referenced from anywhere 

Existing email Address:

Screen Shot 2020-04-28 at 4 38 26 PM

Screen Shot 2020-04-28 at 4 39 01 PM

New Email address or the last step of the existing email address:

Screen Shot 2020-04-28 at 4 40 00 PM


src/components/EmailNotifications/VerifyEmailAddress.js
```diff
@@ -0,0 +1,251 @@
...
+            updateMode
+              ? 'Alternatively, you can update this email or delete it, if you wish to unsubscribe'
+              : ''
+          }`}
+        </span>
+        {(() => {
+          if (updateMode) {
+            const buttonWidth = compactMode
+              ? '100% '
+              : `calc((100% - ${2 * GU}px) /  2)`
```
bpierre, line 108: Nitpicking: should we move this to be inside of `ActionsButtons`, as a normal React component using `css`? That way we could also avoid the inline function here.

src/components/EmailNotifications/VerifyEmailAddress.js
```diff
@@ -0,0 +1,251 @@
...
+                              opacity: 0;
+                            `}
+                          />
+                        )
+                      }
+                      adornmentPosition="end"
+                      type="email"
+                      wide
+                      onChange={handleEmailAddressChange}
+                      placeholder="email@address.com"
```
bpierre, line 144: Nitpicking: I generally use [example.org](http://example.org/) for examples as it is reserved for this case. Also should we use a name? Like `rodrigo` :laughing: 

```suggestion
                      placeholder="rodrigo@example.org"
```

src/components/EmailNotifications/VerifyEmailAddress.js
```diff
@@ -0,0 +1,251 @@
...
+                    {emailInvalid && (
+                      <p
+                        css={`
+                          color: ${theme.negative};
+                          ${textStyle('body4')};
+                          float: left;
+                          margin-top: ${0.5 * GU}px;
+                        `}
+                      >
+                        Please enter a valid email address
```
bpierre, line 156:
```suggestion
                        Please enter a valid email address.
```

src/components/EmailNotifications/VerifyEmailAddress.js
```diff
@@ -0,0 +1,251 @@
...
+                      onChange={handleEmailAddressChange}
+                      placeholder="email@address.com"
+                      onBlur={handleEmailAddressBlur}
+                    />
+                    {emailInvalid && (
+                      <p
+                        css={`
+                          color: ${theme.negative};
+                          ${textStyle('body4')};
+                          float: left;
```
bpierre, line 152: Is the `float` needed?

src/components/EmailNotifications/VerifyEmailAddress.js
```diff
@@ -0,0 +1,251 @@
...
+          css={`
+            ${textStyle('body2')};
+            color: ${theme.surfaceContentSecondary};
+            margin-left: ${1.5 * GU}px;
+            text-align: left;
+          `}
+        >
+          {`By continuing with your email, you agree `}
+          <Link href="#">legal terms </Link> and{' '}
+          <Link href="#"> email collection policy.</Link>
```
bpierre, line 240: Do we have the URLs for these? Also I think the sentence should say “you agree with our legal terms and […]”.

src/components/EmailNotifications/VerifyEmailAddress.js
```diff
@@ -0,0 +1,251 @@
...
+        </span>
+
+        <span
+          css={`
+            ${textStyle('body2')};
+            color: ${theme.surfaceContentSecondary};
+            margin-top: ${1.5 * GU}px;
+          `}
+        >
+          {`Almost there! We’ve sent a verification email to ${email}.
```
bpierre, line 97: I think this should be `emailAddress`?

```suggestion
          {`Almost there! We’ve sent a verification email to ${emailAddress}.
```

src/components/EmailNotifications/VerifyEmailAddress.js
```diff
@@ -0,0 +1,251 @@
...
+  textStyle,
+  useTheme,
+} from '@aragon/ui'
+import { validateEmail } from '../../utils/notifications-utils'
+import emailIllustration from '../../assets/emailIllustration.svg'
+
+const VerifyEmailAddress = React.memo(function VerifyEmailAddress({
+  updateMode,
+  compactMode,
+  email,
```
bpierre, line 21: This doesn’t seem to be used.

src/components/EmailNotifications/VerifyEmailAddress.js
```diff
@@ -0,0 +1,251 @@
...
+      // Set only as valid while user typing. Use blur to set invalid
+      setEmailInvalid(false)
+    }
+  }, [])
+
+  const handleSubscribe = useCallback(async () => {
+    const { subscribedEmail, error } = await onSubscribe(emailAddress)
+
+    if (error) {
+      onError()
```
bpierre, line 51: Not sure what `error` could be here, but should we pass it to `onError()`?

rperez89: ah i raised a concern about the error screens, i left everything prepared just in case we want to show the exact error that happened but right now we only have this general error screen 
https://www.figma.com/file/IdVA6iMn0cWLV5imylwEUP/%E2%9A%96%EF%B8%8F-Court-Dashboard-Master?node-id=4980%3A7830

src/components/EmailNotifications/VerifyEmailAddress.js
rperez89: ahh haha just using the same that is on the designs.
https://www.figma.com/file/IdVA6iMn0cWLV5imylwEUP/%E2%9A%96%EF%B8%8F-Court-Dashboard-Master?node-id=4980%3A7830

src/components/EmailNotifications/VerifyEmailAddress.js
rperez89: no, is ok like this , this is because this screen has two modes update and the normal one that is the third image. if you take a look to the first image says ` We’ve sent a verification email to ` but says no email (that is because that api endpoint is not working haha) and there is using the juror email detected, and  [here](https://github.com/aragon/court-dashboard/pull/300/files#diff-9485ef7603f80508c2cd1eaf16d62436R48)  i am passing to the parent the new juror email ( i am handling the juror email on the parent since is used across several pages from this flow)

src/components/EmailNotifications/VerifyEmailAddress.js
rperez89: replied on the previous one

src/utils/notifications-utils.js
```diff
@@ -0,0 +1,5 @@
+export function validateEmail(email) {
+  // eslint-disable-next-line no-useless-escape
+  var re = /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/
+  return re.test(String(email).toLowerCase())
```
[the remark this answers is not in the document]
rperez89: yes! agree!

src/components/EmailNotifications/VerifyEmailAddress.js
rperez89: ahh yes
Screen Shot 2020-04-28 at 8 18 24 PM

src/components/EmailNotifications/VerifyEmailAddress.js
rperez89: ahh good catch, will add a TODO to add the links once we have them

src/components/EmailNotifications/VerifyEmailAddress.js
bpierre: Should we use `text-align: left` or flexbox? I try to avoid `float` as it impacts the following elements and the parent.

rperez89: ah yes!  just to understand by impact what do you mean? what kind of error or weird thing generates?

rperez89: having a weird behavior using text-align: left, moves the content above a bit down

src/utils/notifications-utils.js
```diff
@@ -0,0 +1,4 @@
+export function validateEmail(email) {
+  // eslint-disable-next-line no-useless-escape
+  return /^.+\@.+\..+$/.test(email)
```
sohkai, line 3: There's no real way to win with email validation... but we should at least be consistent :)

Let's use the same as in aragon/aragon: https://github.com/aragon/aragon/blob/master/src/utils.js#L148

src/components/EmailNotifications/VerifyEmailAddress.js
```diff
@@ -0,0 +1,258 @@
...
+          css={`
+            ${textStyle('body2')};
+            color: ${theme.surfaceContentSecondary};
+            margin-left: ${1.5 * GU}px;
+            text-align: left;
+          `}
+        >
+          By continuing with your email, you agree to Aragon court{' '}
+          <Link href="#">legal terms </Link> and{' '}
+          <Link href="#"> email collection policy.</Link>
```
sohkai, line 246:
```suggestion
          By continuing with your email, you agree to Aragon Court's{' '}
          <Link href="https://anj.aragon.org/legal/terms-general.pdf">terms </Link> and{' '}
          <Link href="https://aragon.one/email-collection.md"> email collection policy</Link>.
```

src/components/EmailNotifications/VerifyEmailAddress.js
```diff
@@ -0,0 +1,258 @@
...
+      `}
+    >
+      <div
+        css={`
+          display: flex;
+        `}
+      >
+        <Checkbox
+          checked={termsAccepted}
+          onChange={checked => setTermsAccepted(checked)}
```
sohkai, line 232:
```suggestion
          onChange={onTermsAccepted}
```

Rather than directly exposing the `setState()` function, we should be generic and accept a handler for when the `onChange()` is triggered.

src/components/EmailNotifications/VerifyEmailAddress.js
```diff
@@ -0,0 +1,258 @@
...
+                  >
+                    Send verification email
+                  </ActionButtons>
+                </div>
+              </>
+            )
+          }
+          return (
+            <Button
+              css={`
```
sohkai, line 203: Nitpick: we usually leave the `css` prop last, as it's usually noisy and doesn't impact the component's primary behaviour (through props) :)

src/components/EmailNotifications/VerifyEmailAddress.js
```diff
@@ -0,0 +1,258 @@
...
+  email,
+  onSubscribe,
+  onSubscribeSuccess,
+  onReSendVerificationEmail,
+  onDeleteEmail,
+  onError,
+}) {
+  const theme = useTheme()
+  const [emailAddress, setEmailAddress] = useState('')
+  const [emailInvalid, setEmailInvalid] = useState(null)
```
sohkai, line 30:
```suggestion
  const [emailInvalid, setEmailInvalid] = useState(false)
```

src/components/EmailNotifications/VerifyEmailAddress.js
```diff
@@ -0,0 +1,258 @@
...
+import { validateEmail } from '../../utils/notifications-utils'
+import emailIllustration from '../../assets/emailIllustration.svg'
+
+const VerifyEmailAddress = React.memo(function VerifyEmailAddress({
+  updateMode,
+  compactMode,
+  email,
+  onSubscribe,
+  onSubscribeSuccess,
+  onReSendVerificationEmail,
```
sohkai, line 24:
```suggestion
  onResendVerificationEmail,
```

src/components/EmailNotifications/VerifyEmailAddress.js
```diff
@@ -0,0 +1,258 @@
...
+    onSubscribeSuccess(subscribedEmail)
+  }, [emailAddress, onError, onSubscribe, onSubscribeSuccess])
+
+  const handleReSendEmail = useCallback(async () => {
+    const { error } = await onReSendVerificationEmail()
+
+    if (error) {
+      onError()
+    }
+  }, [onError, onReSendVerificationEmail])
```
sohkai, line 62:
```suggestion
  const handleResendEmail = useCallback(async () => {
    const { error } = await onResendVerificationEmail()

    if (error) {
      onError()
    }
  }, [onError, onResendVerificationEmail])
```

src/components/EmailNotifications/VerifyEmailAddress.js
```diff
@@ -0,0 +1,258 @@
...
+        `}
+      >
+        <div
+          css={`
+            align-items: center;
+          `}
+        >
+          <img src={emailIllustration} width={141} height={141} alt="" />
+        </div>
+        <span
```
sohkai, line 87: Similar nitpick to before, but perhaps we can use a header element :)

src/components/EmailNotifications/VerifyEmailAddress.js
```diff
@@ -0,0 +1,258 @@
...
+            ${textStyle('body2')};
+            color: ${theme.surfaceContentSecondary};
+            margin-top: ${1.5 * GU}px;
+          `}
+        >
+          Almost there! We’ve sent a verification email to {email}. Kindly check
+          your inbox and click the link to verify your account.
+          {updateMode
+            ? 'Alternatively, you can update this email or delete it, if you wish to unsubscribe'
+            : ''}
```
sohkai, line 107:
```suggestion
          {updateMode && 'Alternatively, you can update this email or delete it, if you wish to unsubscribe.'}
```

Assuming `updateMode` is always a boolean (otherwise wrap in `Boolean()`).

src/components/EmailNotifications/VerifyEmailAddress.js
```diff
@@ -0,0 +1,258 @@
...
+        </span>
+        {(() => {
+          if (updateMode) {
+            const buttonWidth = compactMode
+              ? '100% '
+              : `calc((100% - ${2 * GU}px) /  2)`
```
sohkai, line 113: Looks like we've added this in ActionButton but have not removed this yet :).

src/components/EmailNotifications/VerifyEmailAddress.js
```diff
@@ -0,0 +1,258 @@
...
+                <div
+                  css={`
+                    margin-top: ${5 * GU}px;
+                  `}
+                >
+                  <Field label="Update email address">
+                    <TextInput
+                      value={emailAddress}
+                      adornment={
+                        emailInvalid === false ? (
```
sohkai, line 125: Ah, so this is why `emailInvalid` started out as `null`.

I wonder if we could simplify to:

```js
emailInvalid ? <IconCross /> : emailAddress.trim() ? <IconCheck /> : <IconCheck /> (opacity 0)
```

src/components/EmailNotifications/VerifyEmailAddress.js
```diff
@@ -0,0 +1,258 @@
...
+          `}
+        >
+          Almost there! We’ve sent a verification email to {email}. Kindly check
+          your inbox and click the link to verify your account.
+          {updateMode
+            ? 'Alternatively, you can update this email or delete it, if you wish to unsubscribe'
+            : ''}
+        </span>
+        {(() => {
+          if (updateMode) {
```
sohkai, line 110: To avoid some indentation, we could probably remove the iife and do something like

```js
{!updateMode ? (<ResendButton />) : (
   ...
)}

File: src/components/EmailNotifications/VerifyEmailAddress.js [hunk 22]
```diff
@@ -0,0 +1,258 @@
...
+              Resend verification email
+            </Button>
+          )
+        })()}
+      </div>
+    </div>
+  )
+})
+
+function LegalTermsAndPolicy({ termsAccepted, setTermsAccepted, theme }) {
```
sohkai, line 218: Let's just use `useTheme()` in the component :).

src/components/EmailNotifications/VerifyEmailAddress.js
bpierre: Would it work with an explicit `height: 0`? By impact I mean that its height doesn’t count to calculate the parent height anymore, and the following elements have a specific layout algorithm being applied to them.

src/components/EmailNotifications/VerifyEmailAddress.js
rperez89: why do a call to a hook if we are using it on the parent? don't get the idea around this comment

src/components/EmailNotifications/VerifyEmailAddress.js
```diff
@@ -216,6 +195,13 @@ const VerifyEmailAddress = React.memo(function VerifyEmailAddress({
 })
 
 function LegalTermsAndPolicy({ termsAccepted, setTermsAccepted, theme }) {
+  const handleOnChange = useCallback(
```
sohkai, line 198: I think the last comment might've been misunderstood.

What do you think changing this component's API such that an `onChange()` is passed into the component via its props? That way the responsibility of handling the state change is placed in the parent, and this component is completely agnostic to what happens when the checkbox's state is changed.

src/components/EmailNotifications/VerifyEmailAddress.js
sohkai: Primarily for API reasons—it feels very unnatural to pass down `theme` as it's not something the component should expose (and be controlled by) in its public API.

Using `useTheme()` (which is very, very cheap—we almost never have to worry about its performance) allows us to abstract this completely away from a user. They don't need to provide it, simplifying how the component is used as well as which dependencies are needed to render the component. Imagine a parent component using this, which doesn't need `useTheme()`. We'd force that component to both know, and use `useTheme()`, despite it not being relevant.

Moreover, exposing it as a prop acts as an open door for mistakes to happen in the future. If we passed a hardcoded object down, this component's theme wouldn't change with the rest of the components if the user changed the theme (e.g. changed the site to be in dark mode).

src/components/EmailNotifications/VerifyEmailAddress.js
```diff
@@ -0,0 +1,246 @@
...
+                            opacity: 0;
+                          `}
+                        />
+                      )
+                    }
+                    adornmentPosition="end"
+                    type="email"
+                    wide
+                    onChange={handleEmailAddressChange}
+                    placeholder="email@address.com"
```
sohkai, line 142: This was mentioned elsewhere, but should we use `you@example.org`? Would like to make these consistent across our email fields (in the client, etc) as much as possible.

src/components/EmailNotifications/VerifyEmailAddress.js
```diff
@@ -42,165 +42,155 @@ const VerifyEmailAddress = React.memo(function VerifyEmailAddress({
     }
   }, [])
 
+  const handleOnTermsChange = useCallback(
+    checked => {
+      setTermsAccepted(checked)
+    },
+    [setTermsAccepted]
```
sohkai, line 49: Small nitpick, we don't need to memoize this. In fact, we could pass `setTermsAccepted()` directly into the child component's handlers, but agree that it's better to have a wrapper as it's clearer.

PR #305, opened 2020-04-29
Notifications: Delete email

rperez89:
Modal:

Screen Shot 2020-04-29 at 7 44 58 PM

Global preferences:

Screen Shot 2020-05-06 at 9 34 57 PM


src/components/EmailNotifications/DeleteEmail.js
```diff
@@ -0,0 +1,95 @@
...
+    }
+    onDeleteSuccess()
+  }, [onDelete, onDeleteError, onDeleteSuccess, onNeedsSignature])
+
+  return (
+    <div
+      css={`
+        display: flex;
+        flex-direction: column;
+        justify-content: center;
```
fabriziovigevani, line 34: I think this is not needed ?

src/components/EmailNotifications/DeleteEmail.js
```diff
@@ -0,0 +1,95 @@
...
+            `}
+            onClick={onCancel}
+            size="medium"
+          >
+            Cancel
+          </Button>
+          <Button
+            css={`
+              background: linear-gradient(263.5deg, #ff6969 0%, #ff6969 100%);
+              color: #ffffff;
```
fabriziovigevani, line 82: I think here we could just use the `mode="negative"` prop of the `Button`

src/components/EmailNotifications/DeleteEmail.js
```diff
@@ -0,0 +1,101 @@
...
+}) {
+  const theme = useTheme()
+
+  const handleOnDelete = useCallback(() => {
+    onDelete()
+  }, [onDelete])
+
+  const handleOnCancel = useCallback(async () => {
+    onCancel()
+  }, [onCancel])
```
sohkai, line 20: We don't need these—the responsibility of memoizing them should be at the parent.

src/components/EmailNotifications/DeleteEmail.js
```diff
@@ -0,0 +1,101 @@
+import React, { useCallback } from 'react'
+import { Button, GU, textStyle, useTheme } from '@aragon/ui'
+
+import emailNotifcationIllustration from '../../assets/emailNotifications.svg'
+
+const DeleteEmail = React.memo(function DeleteEmail({
+  isModal,
```
sohkai, line 7: Given the amount of things changing, and that this is not a really complex component, what do you think about just making two different components?

src/components/EmailNotifications/DeleteEmail.js
```diff
@@ -0,0 +1,101 @@
...
+        <span
+          css={`
+            ${textStyle('body2')};
+            color: ${theme.surfaceContentSecondary};
+            margin-top: ${1.5 * GU}px;
+          `}
+        >
+          This action will irreversibly unsubscribe you from all email
+          notifications and remove this email address from our databases. You
+          can always re-subscribe from Notifications preferences later.
```
sohkai, line 68:
```suggestion
          can always re-subscribe from the notifications preferences later.
```

src/components/EmailNotifications/DeleteEmail.js
rperez89: Not sure about this we are setting the same header, same text,  not sure why we should repeat that.
And about the mode could be, i just prefer the boolean but can change it if needed

rperez89: for the isModal same answer than here https://github.com/aragon/court-dashboard/pull/306#discussion_r423143611

src/components/EmailNotifications/DeleteEmail.js
```diff
@@ -0,0 +1,93 @@
...
+          text-align: ${isModal ? 'left' : 'center'};
+          margin-top: ${isModal ? 0 : 5 * GU}px;
+        `}
+      >
+        <span
+          css={`
+            ${textStyle('title2')};
+          `}
+        >
+          {`Delete "${email}"`}
```
sohkai, line 49:
```suggestion
          Delete {email}"
```

src/components/EmailNotifications/DeleteEmail.js
```diff
@@ -0,0 +1,93 @@
...
+        <div
+          css={`
+            margin-top: ${4 * GU}px;
+            display: flex;
+            justify-content: ${isModal ? 'flex-end' : 'center'};
+          `}
+        >
+          <Button
+            css={`
+            margin-right: ${GU}px;}
```
sohkai, line 71:
```suggestion
              margin-right: ${GU}px;}
```

src/components/EmailNotifications/DeleteEmail.js
bpierre: Should we use `Inside` and `useInside()`? `Modal` in aragonUI doesn’t declare an `Inside` scope, but we could do it ourselves anyway:

```jsx
<Modal>
  <Inside name="Modal">
    // …
  </Inside>
</Modal>
```

```jsx
function DeleteEmail() {
  const [insideModal] = useInside('Modal')
  // …
}
```

Both `Inside` and `useInside` are exported by aragonUI: https://github.com/aragon/use-inside

sohkai: Oh interesting, yes, I hadn't considered this!

rperez89: And i never saw this comment!!!

Issue #159, opened 2020-02-14
AccountModule: update "provider" copy to move towards wallets and accounts


sohkai:
See https://github.com/aragon/court-dashboard/pull/126#discussion_r377702525.

Changes we could do:

- [X]  Replace “Enable account” by “Connect account”.

- [x] Replace “Ethereum providers” by “Use account from”.
- [x] Replace “What is an Ethereum provider” by “Don’t have an Ethereum account?”
- [x] Replace “Active wallet” by “Active account” on the connected state
- [x] Replace "Ethereum providers" by "Connection error" on the error state

PR #308, opened 2020-04-30
Notifications: Global preferences - unlock notifications

rperez89:
This PR can be merged since this component is not being referenced from anywhere yet

Screen Shot 2020-04-30 at 12 11 54 PM


src/components/GlobalPreferences/Notifications/UnlockNotifications.js
```diff
@@ -0,0 +1,113 @@
+import React from 'react'
+import { Button, GU, IconConnect, textStyle, useTheme } from '@aragon/ui'
+import emailNotifcationIllustration from '../../../../src/assets/emailNotifications.svg'
```
fabriziovigevani, line 3: Ah we also need to include this file

src/components/GlobalPreferences/Notifications/UnlockNotifications.js
```diff
@@ -0,0 +1,113 @@
...
+                css={`
+                  margin-top: ${GU * 3}px;
+                `}
+                mode="strong"
+                onClick={onUnlock}
+              >
+                Unlock notifications settings
+              </Button>
+            </>
+          ) : (
```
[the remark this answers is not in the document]
rperez89: the first part is for when the user has the eth account disabled, the second one is for when the user has no cookie and needs to unlock to be able to see the settings screen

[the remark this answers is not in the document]
fabriziovigevani: Ah ok, so if the user already has a Cookie this entire component won't be rendered at all right?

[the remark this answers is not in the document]
fabriziovigevani: And what happens when the user already has a Cookie and has the account enabled? I assume this component won't be rendered?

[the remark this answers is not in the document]
rperez89: exactly goes directly to the preferences screen

src/components/GlobalPreferences/Notifications/UnlockNotifications.js
```diff
@@ -0,0 +1,106 @@
...
+          `}
+        >
+          Unlock notifications settings
+        </span>
+        <div
+          css={`
+            display: flex;
+            flex-direction: column;
+            background: ${needsUnlockSettings ? '' : theme.background};
+            padding: ${3 * GU}px ${15 * GU}px;
```
sohkai, line 45: This looks like something we may want to adjust if it's in a small viewport

src/components/GlobalPreferences/Notifications/UnlockNotifications.js
```diff
@@ -0,0 +1,106 @@
...
+      >
+        <div>
+          <img
+            src={emailNotifcationIllustration}
+            width={141}
+            height={141}
+            alt=""
+          />
+        </div>
+        <span
```
sohkai, line 32: Might be worth making this one of the header elements instead

src/components/GlobalPreferences/Notifications/UnlockNotifications.js
```diff
@@ -0,0 +1,106 @@
...
+          ) : (
+            <>
+              <span
+                css={`
+                  ${textStyle('body1')};
+                  color: ${theme.surfaceContentSecondary};
+                  font-weight: 600;
+                `}
+              >
+                You must enable your account to access notifications settings
```
sohkai, line 84:
```suggestion
                You must connect your account to access notifications settings
```

Note that we should be [updating the strings in the account module to also say "connect"](https://github.com/aragon/court-dashboard/issues/159) (can be done separately).

src/components/GlobalPreferences/Notifications/UnlockNotifications.js
```diff
@@ -0,0 +1,106 @@
...
+              <span
+                css={`
+                  ${textStyle('body2')};
+                  color: ${theme.surfaceContentSecondary};
+                  margin-top: ${1.5 * GU}px;
+                `}
+              >
+                Connect to your Ethereum provider by clicking on the{' '}
+                <IconConnect size="small" />
+                Enable account button on the header. You may be temporarily
```
sohkai, line 95:
```suggestion
                Connect account button on the header. You may be temporarily
```

Note that we should also increase the font weight for `Connect account` (I think just wrapping it around a `<strong>` might be enough, but let's see).

src/components/GlobalPreferences/Notifications/UnlockNotifications.js
rperez89: ahh nice, will do it

src/components/GlobalPreferences/Notifications/UnlockNotifications.js
```diff
@@ -0,0 +1,107 @@
...
+          `}
+        >
+          Unlock notifications settings
+        </h3>
+        <div
+          css={`
+            display: flex;
+            flex-direction: column;
+            background: ${needsUnlockSettings ? '' : theme.background};
+            padding: ${3 * GU}px ${compactMode ? 3 : 13 * GU}px;
```
sohkai, line 46:
```suggestion
            padding: ${3 * GU}px ${(compactMode ? 3 : 13) * GU}px;
```

The previous version was not working as you might have expected in `compactMode` :).

PR #318, opened 2020-05-12
Notifications: wiring and integration with API

rperez89:
Working branch for better review and deployment


src/components/EmailNotificationsLoader.js
```diff
@@ -0,0 +1,50 @@
...
+
+  const handleOnClose = useCallback(() => {
+    setModalVisible(false)
+  }, [])
+
+  return (
+    <React.Fragment>
+      {!fetching && (
+        <Modal
+          width={Math.min(76 * GU, width - 40)}
```
sohkai, line 28:
```suggestion
          width={Math.min(76 * GU, width - 5 * GU)}
```

src/services/servicesRequests.js
```diff
@@ -0,0 +1,333 @@
...
+        error: null,
+      }
+    }
+
+    const errors = jsonResponse.errors
+      .map(err => Object.values(err).join(', '))
+      .join(', ')
+
+    return {
+      needsSignature:
```
sohkai, line 77: It looks like these error and authentication validation bits could be taken out and abstracted into a function, e.g.

```js
function notificationsApi(req) {
  const rawResponse = await req()

  if (!rawResponse.ok) {
    const errors = jsonResponse.errors
      .map(err => Object.values(err).join(', '))
      .join(', ')

    return {
      needsSignature:
        rawResponse.status === UNAUTHORIZED_RESPONSE ||
        rawResponse.status === USER_NOT_FOUND ||
        rawResponse.status === FORBIDDEN_RESPONSE,
      error: errors,
    }
  }

  return rawResponse
}

File: src/hooks/useEmailNotifications.js [hunk 3]
```diff
@@ -0,0 +1,84 @@
...
+      error: true,
+      errorMsg: errors,
+    }
+  } catch (err) {
+    console.error(err)
+    return { error: true, errorMsg: err }
+  }
+}
+
+export function useSubscriptionDetails(address) {
```
sohkai, line 58: We should probably run this as a `useEffect()` in the shared notification context, and be dependent on `useWallet()`'s account param to re-run.

src/Routes.js
```diff
@@ -18,3 +21,7 @@ export default function Routes() {
     </Switch>
   )
 }
+
+export function getPreferencesSearch(screen) {
+  return `${GLOBAL_PREFERENCES_QUERY_PARAM}${screen}`
+}
```
sohkai, line 27: It looks like `usePreferences()` also exports this constant.

src/lib/web3-utils.js
```diff
@@ -201,14 +201,12 @@ export async function checkValidEthNode(uri) {
...
   try {
     signHash = await wallet.ethers.getSigner().signMessage(message)
   } catch (err) {
-    error = true
-    errorMsg = err
+    error = err
   }
 
-  return { signHash, error, errorMsg }
+  return { signHash, error }
```
sohkai, line 211: Mentioned in a closed PR, but I'm still wondering if we really need to do this wrapping; can we catch in the `SignatureRequest` / other places where we use this?

src/endpoints.js
```diff
@@ -13,14 +13,15 @@ export const IPFS_ENDPOINT = isLocalOrUnknownNetwork(CHAIN_ID)
...
-  }.eth.aragon.network`
+  // if (isLocalOrUnknownNetwork(CHAIN_ID)) {
+  //   return 'http://127.0.0.1:8050'
+  // }
+
+  // const networkType = getNetworkType(CHAIN_ID)
+  // return `https://court-backend${
+  //   networkType === 'main' ? '' : `-${COURT_SERVER_NAME || networkType}`
+  // }.eth.aragon.network`
+  return 'https://court-backend-usability.eth.aragon.network'
```
sohkai, line 24: Reminder to update this (although maybe this is also a hint to provide an env var to control this)

src/components/GlobalPreferences/GlobalPreferences.js
```diff
@@ -13,12 +13,17 @@ import {
 import { Transition, animated } from 'react-spring/renderprops'
 import { useEsc } from '../../hooks/useKeyboardArrows'
 import Network from './Network/Network'
+import NotificationsManager from './Notifications/NotificationsManager'
 
-const SECTIONS = new Map([['network', 'Network']])
+const SECTIONS = new Map([
+  ['network', 'Network'],
+  ['notifications', 'Email Notifications'],
```
sohkai, line 20:
```suggestion
  ['notifications', 'Notifications'],
```

Not sure, but should we update this name to be consistent with the client?

src/components/GlobalPreferences/Notifications/LoadingPreferences.js
```diff
@@ -0,0 +1,44 @@
...
+      </div>
+      <div
+        css={`
+          ${textStyle('title2')};
+          margin-top: ${4 * GU}px;
+          display: flex;
+          align-items: center;
+        `}
+      >
+        <LoadingRing />{' '}
```
sohkai, line 31: We probably don't need the extra space here, if we're already applying a `margin-left`

src/components/GlobalPreferences/Notifications/LoadingPreferences.js
```diff
@@ -0,0 +1,44 @@
...
+        <span
+          css={`
+            margin-left: ${1.5 * GU}px;
+          `}
+        >
+          Loading...
```
sohkai, line 37:
```suggestion
          Loading…
```

src/components/EmailNotificationsLoader.js
```diff
@@ -0,0 +1,46 @@
...
+          visible={modalVisible}
+          onClose={handleOnClose}
+          css="z-index: 4"
```
sohkai, line 31: I might've asked this for the onboarding modal too, but do we need this `z-index`? We should be rendering this modal in a portal outside of the main view so it should automatically get put on top.

src/components/GlobalPreferences/Notifications/NotificationsManager.js
```diff
@@ -0,0 +1,116 @@
...
+        }
+      }
+    }
+    getEmail()
+    return () => {
+      cancelled = true
+    }
+  }, [address, account, emailExists, emailVerified, notificationsDisabled, fetchingSubscriptionData, token])
+
+  console.log('startingScreenId ', startingScreenId)
```
sohkai, line 102: Reminder to remove this :)

src/components/GlobalPreferences/Notifications/NotificationsManager.js
```diff
@@ -58,46 +59,73 @@ const NotificationsManager = React.memo(function NotificationsManager() {
...
+      return setStartingScreenId(VERIFY_EMAIL_ADDRESS_PREFERENCES)
+    }
+    if (!emailVerified) {
+      return setStartingScreenId(EMAIL_NOTIFICATIONS_FORM_SCREEN)
+    }
+
+    if (emailVerified && !jurorNeedsSignature) {
+      return setStartingScreenId(NOTIFICATIONS_PREFERENCES_SCREEN)
+    }
+  }, [account, emailExists, emailVerified, fetchingEmail, fetchingSubscriptionData, jurorNeedsSignature])
```
sohkai, line 106: I'm a little bit confused with the above `useEffect()`s. Should they all live under a single effect? Otherwise it's going to be weird if some conditions trigger multiple of the `useEffect()`s and call `setStartingScreenId()` multiple times.

src/services/notificationServiceApi.js
```diff
@@ -143,6 +143,7 @@ export async function getJurorEmail(address) {
       return {
         email: jsonResponse?.email,
         error: null,
+        needsSignature: false,
```
sohkai, line 146: Do we need to return this? It seems weird that only this `ok` response would be returning `needsSignature`.

rperez89: nop, forgot to remove it after testing

rperez89:
> 🙌🙌🙌 I think we've squashed the bugs and we're good to go!

would love a bit more of testing from our folks? did you guys have a chance to try it?

PR #320, opened 2020-05-13
Clock Module: Use new prop for Timer

fabriziovigevani:
Uses new `showIcon` prop for the Timer component

### DIFF
--- a/src/components/ClockModule.js
+++ b/src/components/ClockModule.js
@@ -76,7 +76,7 @@ function ClockModule() {
             css={`
               background-color: ${theme.background};
               display: flex;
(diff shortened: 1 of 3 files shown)


src/components/ClockModule.js
```diff
@@ -96,19 +96,12 @@ function ClockModule() {
               & span {
                 line-height: 1;
               }
-
-              // TODO: Use showIcon new prop when available, see: https://github.com/aragon/aragon-ui/pull/734
-              & > time {
-                & > span:first-child {
-                  display: none;
-                }
-              }
```
bpierre, line 105: :massage_man:

PR #323, opened 2020-05-14
Emails: migrated to court-backend

promaty:
Moved to https://github.com/aragonone/court-backend/blob/development/emails


promaty:
Signed!

PR #325, opened 2020-05-15
Metadata docs: Update expected structure

fabriziovigevani:
Considers new expected structure for most recent disputes.

### DIFF
--- a/docs/metadata.md
+++ b/docs/metadata.md
@@ -20,11 +20,7 @@ type Arbitrable @entity {
 }
 ```
 
-In the frontend we try to parse the `metadata` as a JSON object.
-
-### metadata - JSON object
-
-We are going to try to find the following keys:
+The dispute `metadata` should be a JSON Object and should have the following keys:
 
 1. `description`
 2. `metadata`
@@ -46,9 +42,7 @@ The content is expected to have the following structure:
 }
 ```
 
-#### metadata - String
-
-In case `metadata` is a plain string, we assume it to be the dispute's description and set the dispute creator (`subject.id`) as the plaintiff. No agreement link will be available.
+> Note that even though the description is present in the ipfs content, we require to also be present at the `metdata` field. The reason for this is so we don't need to fetch all disputes ipfs content when loading all disputes cards.
 
 ---
 
--- a/docs/metadata.md
+++ b/docs/metadata.md
@@ -29,7 +29,7 @@ We are going to try to find the following keys:
 1. `description`
 2. `metadata`
 
-If `metadata` is a valid IPFS hash/cid we'll fetch its content.
+If `metadata` is a valid [native IPFS URL](https://docs.ipfs.io/guides/guides/addressing/#native-urls) we'll fetch its content through IPFS.
 The content is expected to have the following structure:
 
 ```javascript
@@ -84,4 +84,3 @@ The dashboard checks if the `data` field is a valid IPFS hash/cid
 ### data - Plain String
 
   Show the data as the evidence text
-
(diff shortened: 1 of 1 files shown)


docs/metadata.md
```diff
@@ -20,19 +20,37 @@ type Arbitrable @entity {
...
 
-### metadata - Json object
+### metadata - JSON object
 
 We are going to try to find the following keys:
 
 1. `description`
 2. `metadata`
 
+If `metadata` is a valid IPFS hash/cid we'll fetch its content.
```
sohkai, line 32:
```suggestion
If `metadata` is a valid [native IPFS path](https://docs.ipfs.io/guides/guides/addressing/#native-urls) we'll fetch its content through IPFS.
```

Wanted to double check, we are assuming the `ipfs://` protocol prefix, right (although we still attempt to fallback to detecting if it's a CID/IPLD path if it doesn't exist)?

docs/metadata.md
```diff
@@ -46,25 +64,24 @@ type Evidence @entity {
 }
 ```
 
-The dashboard check if the `data` field is a valid IPFS hash/cid
+The dashboard checks if the `data` field is a valid IPFS hash/cid
```
sohkai, line 67: Should this also have the same correction above, as a native IPFS URL?

docs/metadata.md
```diff
@@ -46,25 +64,24 @@ type Evidence @entity {
...
 
 ### data - IPFS hash
+  Try to fetch the content from IPFS
```
sohkai, line 70: It looks like we may have added some additional indentation here.

docs/metadata.md
fabriziovigevani: Oh yeah good catch.

fabriziovigevani: We expect to be in the form of `ipfs:{cid | hash}`. Not sure if it's the same?

sohkai: Ah shit, no it isn't :( (it needs the `://`, not just the `:`).

Nevermind, I've changed the suggestion to something else!

docs/metadata.md
```diff
@@ -46,25 +64,24 @@ type Evidence @entity {
 }
 ```
 
-The dashboard check if the `data` field is a valid IPFS hash/cid
+The dashboard checks if the `data` field is a string in the form of `ipfs:{CID | IPLD path}`.
 
-### data - IPFS hash
+### data - IPFS string
```
sohkai, line 69:
```suggestion
### data - IPFS path
```

docs/metadata.md
```diff
@@ -46,9 +42,7 @@ The content is expected to have the following structure:
 }
 ```
 
-#### metadata - String
-
-In case `metadata` is a plain string, we assume it to be the dispute's description and set the dispute creator (`subject.id`) as the plaintiff. No agreement link will be available.
```
fabriziovigevani, line 34: Note that we only require the `metadata` field to be A JSON object. If we can't parse it, we are displaying the `Could not fetch content from IPFS` error.

docs/metadata.md
```diff
@@ -46,9 +42,7 @@ The content is expected to have the following structure:
...
+> Note that even though the description is present in the ipfs content, we require to also be present at the `metdata` field. The reason for this is so we don't need to fetch all disputes ipfs content when loading all disputes cards.
```
sohkai, line 45:
```suggestion
> Note that even though `description` should be present in the IPFS content, we also assume it to be present in the `metadata` field. We do this to avoid needing to fetch every disputes' IPFS content when loading their dispute cards.
```

fabriziovigevani:
Thanks!

PR #327, opened 2020-05-15
Notifications: contextualize email notifications

sohkai:
Stub/spike of moving the email notifications information up to a context.


src/components/EmailNotifications/EmailNotificationsProvider.js
```diff
@@ -0,0 +1,142 @@
...
+      const response = await subscribeToNotifications(account, email)
+
+      if (response.error && !response.needsSignature) {
+        return response.error
+      }
+
+      if (!asyncCancelled.current) {
+        setEmail(email)
+        if (needsSignature !== Boolean(response.needsSignature)) {
+          setNeedsSignature(!needsSignature)
```
rperez89, line 41: Trying to figure out how are you thinking to handle the action after the signature? i mean after the signature is successful we need to call again the `subscribeToNotifications` again.
And the other question that i have is : given that here we don't have an useEffect that listen to the `needsSignature` state,  i guess that the plan is to have that useEffect in the manager?

src/components/EmailNotifications/EmailNotificationsProvider.js
```diff
@@ -0,0 +1,142 @@
...
+  }, [account])
+
+  // Once we know the account has an associated email, fetch its email
+  useEffect(() => {
+    if (!subscriptionDetails.emailExists) {
+      return
+    }
+
+    const getEmail = async () => {
+      const { needsSignature, email } = await getJurorEmail(account)
```
rperez89, line 104: Giving a second thought to this, i think that here we are going to be fetching the juror email every time an account is connected, and when the cookie has expired we are going to ask the juror to sign a message just for connecting his account and not sure if that is what we want.
The way i see this is that we need to do that only when we are subscribing to notifications through the modal or entering the global preferences not every time the juror connect his account to interact with the dashboard.
What do you think?

src/components/EmailNotifications/EmailNotificationsProvider.js
sohkai: > given that here we don't have an useEffect that listen to the needsSignature state, i guess that the plan is to have that useEffect in the manager?

Yes, exactly, that was something I wanted to leave to the state machine manager, given that we could have different machines based on the area (preferences or modal).

> Trying to figure out how are you thinking to handle the action after the signature?

Similar to above, I was hoping the manager would handle this. These functions just attempt to "optimistically" call the API, and if it fails, it'll let us know either with the error or a state change.

src/components/EmailNotifications/EmailNotificationsProvider.js
```diff
@@ -0,0 +1,142 @@
...
+  // When account connects, fetch their subscription details
+  useEffect(() => {
+    if (!account) {
+      return
+    }
+
+    const fetchSubscriptionDetails = async () => {
+      const response = await getSubscriptionDetails(account)
+
+      if (!asyncCancelled.current) {
```
sohkai, line 81: I used the same `asyncCancelled` ref here and after, but thinking about it, `useEffects()` should contain their own local cancellation.

src/components/EmailNotifications/EmailNotificationsProvider.js
sohkai: Good point. Yes, it sounds like we should do this in a callback instead.

src/components/EmailNotifications/EmailNotificationsProvider.js
```diff
@@ -0,0 +1,142 @@
+import React, {
```
sohkai, line 1: Nitpick: we may want to rename this file to be `EmailNotificationsApiProvider` for clarity.

src/components/EmailNotifications/EmailNotificationsProvider.js
```diff
@@ -0,0 +1,142 @@
...
+    error: null,
+    fetching: false,
+  })
+
+  const handleOnSubscribe = useCallback(
+    async email => {
+      const response = await subscribeToNotifications(account, email)
+
+      if (response.error && !response.needsSignature) {
+        return response.error
```
sohkai, line 35: Thinking about this, we may want to throw the error instead (and perhaps all the requests should be throwing those errors, similar to https://github.com/aragon/aragon/blob/master/src/components/GlobalPreferences/Notifications/notification-service-api.js#L60).

PR #329, opened 2020-05-16
DisputeTimeline: Fix drafted date

fabriziovigevani:
Noticed that the drafted date was off in the dispute timeline.

- Sets the `endTime` on the draft phase only after it has passed.

**Before:**
Screen Shot 2020-05-16 at 8 16 49 PM

**After:** 
Screen Shot 2020-05-16 at 8 20 44 PM

### DIFF
--- a/src/utils/dispute-utils.js
+++ b/src/utils/dispute-utils.js
@@ -381,8 +381,9 @@ function getRoundPhasesAndTime(courtConfig, round, currentPhase) {
     ]
   }
 
-  const votingEndTime =
-    disputeDraftStartTime + termDuration * (delayedTerms + commitTerms)
+  const disputeDraftEndTime =
+    disputeDraftStartTime + termDuration * delayedTerms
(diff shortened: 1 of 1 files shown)


src/utils/dispute-utils.js
```diff
@@ -393,8 +394,13 @@ function getRoundPhasesAndTime(courtConfig, round, currentPhase) {
 
   const roundPhasesAndTime = [
     {
-      // Jurors can be drafted at any time
+      // Jurors can be drafted at any time, so we'll only set the
+      // `endTime` when the drafting phase has already passed
       phase: DisputesTypes.Phase.JuryDrafting,
+      endTime:
+        DisputesTypes.Phase.JuryDrafting !== currentPhase.phase
```
sohkai, line 401: Just wondering, how do we know the later steps' end times if we haven't got past the draft phase?

Should we do an early return here in that case?

PR #335, opened 2020-05-19
Advance court terms when necessary

fabriziovigevani:
This PR queues the heartbeat action request  to the `Request Queue` when attempting to perform an action which requires the court to be synced or at most 1 term behind.

E.g Court clock is behind by 2 terms and a user tries to activate some ANJ.

For a reference of up to how many terms behind can the court clock be for a given action see https://github.com/aragon/court-dashboard/commit/cdd5fe8bf29827344ce932d0dfc33c5025b7a4ae

In the resulting implementation, the request processing logic was centralized into a `useRequestProcessor` hook which is the one in charge of adding the activities and constructing the request that will be handled by the request queue.

Also since we had to implement a mapping of `action => allowedTermsBehind` decided also to create a mapping for court action types that will be use by `radspec`, `activity-types` and the new `allowedTermsBehind` mapping.

For easier review: 

- [Court Clock now tracks the expected current term as opposed to the actual current term](https://github.com/aragon/court-dashboard/commit/bc025cb4508c365c731ac71e6b0d8782abf827ed)

- [Use new action types for radspec and activity-types](https://github.com/aragon/court-dashboard/commit/adde4408ef74d244e4ea264fd899952858b7f982)

- [Implement `useRequestProcessor` hook which handles all request logic and adds the heartbeat request at the front of the request queue if necessary](https://github.com/aragon/court-dashboard/commit/44ff1b0e0d52f3567b3834d7f3f15e84f8015cc9)

## **Left TODO:**

- [x] Update package version env after we merge https://github.com/aragon/court-dashboard/pull/341 to clear activities


src/providers/CourtClock.js
```diff
@@ -17,16 +17,16 @@ function CourtClockProvider({ children }) {
...
+    expectedCurrentTerm,
     terms,
     termDuration
   )
-  const expectedCurrentTerm = getExpectedCurrentTermId(now, terms, termDuration)
 
   const courtClock = useMemo(
     () => ({
-      currentTermId: actualCurrentTerm,
+      currentTermId: expectedCurrentTerm,
```
sohkai, line 29: Not sure what this change does—how does this affect the clock module?

src/hooks/useRequestProcessor.js
```diff
@@ -0,0 +1,66 @@
+import { useCallback } from 'react'
+import { useHeartbeat } from './useCourtContracts'
+import { useCourtClock } from '../providers/CourtClock'
+import { useRequestQueue } from '../providers/RequestQueue'
+import { useActivity } from '../components/Activity/ActivityProvider'
```
sohkai, line 5: This is separate from this, but perhaps we should move this component to the `providers`? Maybe you've already had this conversation before haha.

src/allowedTermsBehind.js
```diff
@@ -0,0 +1,22 @@
+import actions from './types/court-action-types'
```
sohkai, line 1: Maybe this might be better placed in an `actions/` or other directory. For a new person to this repo, this file does not seem immediately obvious what its purpose is for :).

A couple comments explaining it might also be useful!

src/hooks/useRequestProcessor.js
```diff
@@ -0,0 +1,66 @@
...
+export function useRequestProcessor() {
+  const { request } = useHeartbeat()
+  const { addActivity } = useActivity()
+  const { addRequests } = useRequestQueue()
+  const { neededTransitions } = useCourtClock()
+
+  const processRequests = useCallback(
+    requests => {
+      const processedRequests = []
+        .concat(requests)
```
sohkai, line 19:
```suggestion
      const processedRequests = requests
```

Since we're just going to map, we don't need to create an extra array and concat

src/hooks/useRequestProcessor.js
```diff
@@ -0,0 +1,66 @@
...
+              ensureConfirmation,
+              isTx,
+              ...request,
+            }
+          }
+        )
+
+      const maxAllowedTermsBehind = processedRequests
+        .filter(req => req.isTx)
+        .reduce((acc, req) => Math.max(acc, allowedTermsBehind[req.name]), -1)
```
sohkai, line 48: The `allowedTermsBehind[req.name]` just feels a _little_ bit sketchy, since we're using the name (which may be mistyped or disconnected from the actual action).

Not sure if there's a good solution for this though.

src/hooks/useRequestProcessor.js
```diff
@@ -0,0 +1,66 @@
...
+
+      // Some court actions require the court clock to be at most x terms behind
+      // Check if a term transition is needed
+      if (
+        maxAllowedTermsBehind >= 0 &&
+        neededTransitions > maxAllowedTermsBehind
+      ) {
+        processedRequests.unshift(request(neededTransitions, true))
```
sohkai, line 56: Perhaps we can rename `request()` to `requestHeartbeat()`; this usage makes it look like an arbitrary request

src/hooks/useCourtContracts.js
```diff
@@ -1,21 +1,30 @@
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react'
 import { captureException } from '@sentry/browser'
-import { CourtModuleType } from '../types/court-module-types'
-import { useActivity } from '../components/Activity/ActivityProvider'
-import { useContract, useContractReadOnly } from '../web3-contracts'
+
+// hoooks
```
sohkai, line 4:
```suggestion
// hooks
```

src/hooks/useRequestProcessor.js
```diff
@@ -0,0 +1,66 @@
...
+  const { addRequests } = useRequestQueue()
+  const { neededTransitions } = useCourtClock()
+
+  const processRequests = useCallback(
+    requests => {
+      const processedRequests = []
+        .concat(requests)
+        .map(
+          ({
+            action,
```
sohkai, line 22: Maybe we should rename this key to `request`? Since the higher-level concept we're using is "requests".

src/hooks/useCourtContracts.js
```diff
@@ -410,24 +366,31 @@ export function useHeartbeat() {
...
     transitions => {
-      return addRequests({
-        intent: () =>
-          addActivity(aragonCourtContract.heartbeat(transitions), 'heartbeat', {
-            transitions,
-          }),
-        description: radspec.heartbeat(transitions),
-        isTx: true,
-      })
+      return addRequests(request(transitions))
```
sohkai, line 384: Similar here, it might be nice to rename this to `requestHeartbeat()`

src/hooks/useRequestProcessor.js
```diff
@@ -0,0 +1,66 @@
...
+            ...request
+          }) => {
+            const intent = isTx
+              ? () => addActivity(action(), name, params)
+              : action
+
+            return {
+              name,
+              intent,
+              description:
```
sohkai, line 37: It might just be me, but I actually like being more explicit about the radspec description and passing it down from the request creation site.

I think that would allow us to avoid passing down the params (which seems like an odd thing to do if we can avoid it—does `addActivity()` really need it or can it also just use the passed down description?)

src/providers/CourtClock.js
fabriziovigevani: This change is to keep track of the supposed current term id assuming the court transitions every time a heartbeat is needed. 

In the context of disputes, we need to be able to transition phases even though the actual current term is behind. 
E.g we are in the committing phase, and the heartbeat service is down (and no one has heartbeat it manually). If we would keep track of the `actualCurrentTerm` as opposed to the `expectedCurrentTerm` then the dispute would not transition to the reveal phase when is supposed to.

> how does this affect the clock module?

The clock module will still work as expected since the UI is reactive relative to the wether the clock is synced or not.

src/hooks/useRequestProcessor.js
fabriziovigevani: Ah yes! I had some doubts on where to place this hook at the beginning haha. Agree that placing it in the `providers` folder makes sense

src/hooks/useRequestProcessor.js
fabriziovigevani: Ah actually this is necessary because the way we pass requests (can be an `Array` or`Object`) See https://github.com/aragon/court-dashboard/blob/0ec4b75a73f01a5c544c51cdbe6902f5a829d6d3/src/hooks/useCourtContracts.js#L90

Maybe we should just pass always an array?

src/hooks/useRequestProcessor.js
fabriziovigevani: hmm maybe we could change `name` to be `type`? Because now that you mention it, the name is the actually the `court-action-type` that now we are changing it to be a `Symbol` as you suggested.

src/hooks/useRequestProcessor.js
fabriziovigevani: Makes sense

fabriziovigevani: The reason i used `action` is because the entire object is actually the request and we would have the problem where we would have to rename this as well https://github.com/aragon/court-dashboard/blob/0ec4b75a73f01a5c544c51cdbe6902f5a829d6d3/src/hooks/useRequestProcessor.js#L28

src/hooks/useRequestProcessor.js
fabriziovigevani: Good point, i think the way you suggest results in a more abstracted functionality.

> does addActivity() really need it or can it also just use the passed down description?)

It does need it yes, but we can change it to pass down the full description. Besides the params passed are actually used to compute the description using `radpsec` again when getting the activity data, si we definitely need to update that.

src/types/court-action-types.js
```diff
@@ -0,0 +1,18 @@
+export default {
+  ActivateAnj: 'activateAnj',
```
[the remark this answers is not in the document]
fabriziovigevani: Ah realized that since we now updated the `activity-types` and `radspec` mappings to use the `action-types` (see https://github.com/aragon/court-dashboard/commit/adde4408ef74d244e4ea264fd899952858b7f982), if we change them to symbols then older activities that use the older convention would error.

src/hooks/useRequestProcessor.js
sohkai: Yes, let's do that!

src/hooks/useRequestProcessor.js
sohkai: Ah, I missed that.

Yes, in this case, since we don't have proptypes/typescript, being a bit more obvious with the cast (either avoiding it entirely via the API, or ensuring that the reader realizes we may need to wrap into an Array) would be useful.

src/types/court-action-types.js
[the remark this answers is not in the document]
fabriziovigevani: We decided to wipe older activities so we will update in favor of using symbols (see https://github.com/aragon/court-dashboard/pull/341)

src/hooks/useRequestProcessor.js
fabriziovigevani: Since we will be  wiping older activities, a backwards compat implementation is not needed (see https://github.com/aragon/court-dashboard/pull/341)

src/components/Activity/activity-types.js
```diff
@@ -10,134 +9,77 @@ import iconDraftJury from './assets/activity-icon-draft-jury.svg'
...
+  },
+  [actions.SETTLE_REWARD]: {
+    icon: iconAnj,
+    title: 'Settle reward',
+  },
+  [actions.SETTLE_APPEAL_DEPOSIT]: {
+    icon: iconAnj,
+    title: 'Settle appeal deposit',
+  },
+  transaction: {
```
sohkai, line 73: Just wondering if we should move this to be the last one or etc.

It's kind of weird just sticking out here (and in fact, do we really need this?)

src/providers/useRequestProcessor.js
```diff
@@ -0,0 +1,61 @@
+import { useCallback } from 'react'
```
sohkai, line 1: Not sure, but this looks like it shouldn't be in the `providers` folder?

fabriziovigevani: Ah shit, i miss understood one comment above, thought that initially you were suggesting to move it to the providers folder since it uses a lot of the providers hooks https://github.com/aragon/court-dashboard/pull/335#discussion_r428560449 but i can see you were referring to the activity provider haha.

src/components/Activity/activity-types.js
fabriziovigevani: Hmm yeah probably we can just remove it, there's no really an action that will use this default activity.

src/components/Activity/activity-types.js
```diff
@@ -70,10 +70,6 @@ const ACTIVITY_TYPES = {
     icon: iconAnj,
     title: 'Settle appeal deposit',
   },
-  transaction: {
-    title: 'Transaction',
-    icon: iconAnj,
-  },
```
sohkai: I think we also need to remove the default in https://github.com/aragon/court-dashboard/pull/335/commits/119a2d7b9ae4cc7391888dc92cb63399a91bdf15#diff-cabff7987bb1e5e4650eb3786dac7c19R80

PR #337, opened 2020-05-20
Notifications: Animation transition fix and prevent same email update

rperez89:


src/components/EmailNotifications/VerifyEmailAddress.js
```diff
@@ -25,11 +26,18 @@ const VerifyEmailAddress = React.memo(function VerifyEmailAddress({
   const theme = useTheme()
   const [emailAddress, setEmailAddress] = useState('')
   const [emailInvalid, setEmailInvalid] = useState(false)
+  const [errorMessage, setErrorMessage] = useState('')
```
sohkai, line 29: Perhaps we could just use a single state for the email error? E.g. if there's an error, it's obviously invalid :).

src/components/EmailNotifications/VerifyEmailAddress.js
```diff
@@ -25,11 +26,18 @@ const VerifyEmailAddress = React.memo(function VerifyEmailAddress({
...
     const email = e.target.value
-    setEmailInvalid(!validateEmail(email))
+    const emailInvalid = !validateEmail(email)
+    setEmailInvalid(emailInvalid)
+    if (emailInvalid) {
+      setErrorMessage('Please enter a valid email address.')
+    }
   }, [])
 
   const handleEmailAddressChange = useCallback(e => {
```
sohkai, line 43: If we don't consolidate the state, we should probably also remove the error when changing the email input?

src/components/EmailNotifications/EmailNotificationsManager.js
```diff
@@ -651,35 +651,60 @@ const EmailNotificationsManager = React.memo(
...
+      : AnimatedBox
+
+  return <Container {...props} />
+})
+
+const AnimatedBox = ({ ...props }) => {
+  const ref = useRef(null)
+  const [height, setHeight] = useState(0)
+
+  const setRef = useCallback(node => {
```
sohkai, line 670: Not sure if I understand this; do we even need to set the ref? Should we also listen to the `node`'s height changing?

src/components/EmailNotifications/EmailNotificationsManager.js
```diff
@@ -651,35 +651,60 @@ const EmailNotificationsManager = React.memo(
...
       </div>
     </Box>
   )
-})
+}
 
 const AnimatedContainer = React.memo(function AnimatedModal({
   screenId,
   children,
+  refCallback,
```
sohkai, line 707: We may want to use [`forwardRef()`](https://reactjs.org/docs/forwarding-refs.html) instead.

src/components/EmailNotifications/EmailNotificationsManager.js
fabriziovigevani: Ah yeah i think @rperez89 already did this but we can remove `useRef` here, as it's not necessary and we can change `setRef` to be `ref`.

Also, now that you mentioned to listen to the node height we can probably create a custom hook to encapsulate the logic.

```
function useNodeHeight() {
  const [height, setHeight] = useState(0)
  const nodeObserver = useRef(null)

  const customRef = useCallback(node => {
    if (nodeObserver.current) {
      nodeObserver.current.disconnect()
    }

    if (node) {
      setHeight(node.clientHeight)
      nodeObserver.current = new ResizeObserver(entries => {
        const { height } = entries[0].contentRect
        setHeight(height)
      })

      nodeObserver.current.observe(node)
    }
  }, [])

  return [height, customRef]
}

```

sohkai: Might be something useful to consider adding to aragonUI cc @bpierre!

bpierre: Nice, yes!

PR #338, opened 2020-05-20
Disputes: Transition phases using Court Clock

fabriziovigevani:
Builds from #335 

Disputes now transition phases relative to the Clock term as opposed to checking the current phase every second.


src/utils/dispute-utils.js
```diff
@@ -473,7 +484,7 @@ function getRoundPhasesAndTime(courtConfig, round, currentPhase) {
     }
 
     // If round not appealed
-    if (!currentPhase.appealed) {
```
fabriziovigevani, line 476: Note that is actually not necessary to tell if the `currentPhase` was appealed or not since we have that information directly from the `round`.  For this reason we removed the `appealed` attribute at `getAdjudicationPhase`

E.g See https://github.com/aragon/court-dashboard/pull/338/files#diff-0311c4b36f556c7a49619bd3afa3d955L341

src/utils/court-utils.js
```diff
@@ -54,16 +42,30 @@ export function getExpectedCurrentTermId(now, terms, termDuration) {
...
+  return termStartTime
+}
+
+export function getTermEndTime(termId, { terms, termDuration }) {
+  let termEndTime = 0
+
+  const termStartTime = getTermStartTime(termId, { terms, termDuration })
+  termEndTime = termStartTime + (termDuration - 1)
+
+  return termEndTime
```
sohkai, line 62: Only comment for the above is whether we should use a default of `0`; that doesn't seem to be very useful and we should perhaps error in cases where we can reach them (only in `getTermStartTime()` it seems?)

src/utils/dispute-utils.js
```diff
@@ -84,21 +83,26 @@ function overrideVoidedDispute(dispute, voidedDispute) {
  * Construct dispute timeline for the given dispute
  * @param {Object} dispute The dispute to get the timeline from
  * @param {Object} courtConfig The court configuration
+ * @param {Number} currentTerm The court current term
  * @returns {Array} The timeline of the given dispute
  */
-export function getDisputeTimeLine(dispute, courtConfig) {
+export function getDisputeTimeLine(dispute, courtConfig, currentTerm) {
```
sohkai, line 89: I'm not sure we want to make this change. It's not as if a dispute is still open if the term hasn't adjusted—it definitely has closed / moved on to the next round, but the internal state is just not updated yet.

src/utils/dispute-utils.js
```diff
@@ -216,30 +225,36 @@ export function getPhaseAndTransition(dispute, courtConfig, date) {
...
+    const { phase, phaseEndTerm } = currentAdjudicationPhase
+    if (phase === DisputesTypes.Phase.Ended) {
       currentAdjudicationPhase = {
         ...currentAdjudicationPhase,
         phase: DisputesTypes.Phase.ExecuteRuling,
       }
     }
-    return { ...currentAdjudicationPhase, roundId: number }
+
+    const nextTransition = phaseEndTerm
```
sohkai, line 240: Since we weren't passing this back before, I'm wondering what this will do

src/utils/dispute-utils.js
```diff
@@ -271,23 +284,24 @@ export function getAdjudicationPhase(dispute, round, date, courtConfig) {
     }
   }
 
-  // If given term is before the reveal start term of the last round, then jurors are still allowed to commit votes for the last round
-  const revealTermStartTime = draftTermEndTime + commitTerms * termDuration
-  if (date.isBefore(revealTermStartTime)) {
+  // Jurors can commit their votes between when the commit phase term starts and ends
+  // Note that the commit start term is the same as the draft end term
```
sohkai, line 288: Wondering: if the time has passed, but the term hasn't updated, can a juror still commit/reveal their vote?

src/utils/dispute-utils.js
```diff
@@ -350,11 +360,11 @@ export function getAdjudicationPhase(dispute, round, date, courtConfig) {
  * @param {Object} courtConfig The court configuration
  * @param {Object} round The round to get the phases from
  * @param {Object} currentPhase The dispute's current phase
+ * @param {Number} currentTerm The court current term
  * @returns {Array} Array of all `round` phases.
  */
-function getRoundPhasesAndTime(courtConfig, round, currentPhase) {
+function getRoundPhasesAndTime(courtConfig, round, currentPhase, currentTerm) {
```
sohkai, line 366: Total bikeshedding, but if it's not too much trouble, maybe we can standardize on `courtConfig` always being the last parameter?

src/utils/dispute-utils.js
```diff
@@ -271,23 +284,24 @@ export function getAdjudicationPhase(dispute, round, date, courtConfig) {
     }
   }
 
-  // If given term is before the reveal start term of the last round, then jurors are still allowed to commit votes for the last round
-  const revealTermStartTime = draftTermEndTime + commitTerms * termDuration
-  if (date.isBefore(revealTermStartTime)) {
+  // Jurors can commit their votes between when the commit phase term starts and ends
+  // Note that the commit start term is the same as the draft end term
+  const commitEndTerm = draftEndTerm + commitTerms - 1
```
sohkai, line 289: Saw the note above, but wanted to sanity and double check the `- 1`.

If the `draftEndTerm` is `3`, and we allow for `2` `commitTerms`, then we would be allowed to start committing votes at `3` until `4`?

It just seems a bit weird _this_ particular phase is _inclusive_, whereas all the other seem _exclusive_.

src/utils/dispute-utils.js
```diff
@@ -498,12 +509,12 @@ function getRoundPhasesAndTime(courtConfig, round, currentPhase) {
  * @param {Object} courtConfig The court configuration
  * @returns {Number} The end time of the evidence submission phase in ms
  */
-function getEvidenceSubmissionEndTime(dispute, courtConfig) {
+function getEvidenceSubmissionEndTerm(dispute, courtConfig) {
   const firstRound = dispute.rounds[0]
 
   // If the evidence period is closed before the full `evidenceTerms` period,
   // the drafTermId for the first round is updated to the term this happened.
```
sohkai, line 516:
```suggestion
  // the draftTermId for the first round is updated to the term this happened.
```

src/utils/dispute-utils.js
fabriziovigevani: Yes, see https://github.com/aragon/court-dashboard/pull/335#discussion_r428656118

src/utils/court-utils.js
fabriziovigevani: The default `0` is because when loading the court, the court config may haven't loaded yet. 

> we should perhaps error in cases where we can reach them

What do you mean exactly by this?

src/utils/dispute-utils.js
fabriziovigevani: Ah actually we were passing this, just that we changed how we compute it. 

Before we were returning it along the phase at `getAdjudicationPhase`, (see for example https://github.com/aragon/court-dashboard/pull/338/files#diff-0311c4b36f556c7a49619bd3afa3d955L263) 

Now instead, we return the term at which the phase ends and compute the `endTime` of said term. This change was mainly to prevent computing the `nextTransition` on every return in `getAdjudicationPhase` which required calling `getTermEndTime` every time

src/utils/dispute-utils.js
fabriziovigevani: > If the draftEndTerm is 3, and we allow for 2 commitTerms, then we would be allowed to start committing votes at 3 until 4?

Yes, this is correct. I think that the reason works this way is because the drafting phase can be executed any time as opposed to other phases in which there's a fixed period.

See https://github.com/aragon/aragon-court/blob/492abfd3fb887ab5dd9a71dc480ef7e33c45b7d1/contracts/disputes/DisputeManager.sol#L1054

More interestingly, in your example, if the drafting is executed say at the middle of the term, then the total commit period would last 1 and a half terms.

src/utils/dispute-utils.js
fabriziovigevani: Could you extend on this? I'm not totally sure i follow. 

Note that we are tracking the expected current term now as opposed to the internal court contract term as explained here https://github.com/aragon/court-dashboard/pull/335#discussion_r428656118

fabriziovigevani:
> With this change, the dispute timeline WILL NOT be updated until the term itself is, right?

Yes exactly. And also note that the current term that we track now is the **expected** current term and not the **actual** internal current term.

See https://github.com/aragon/court-dashboard/blob/0ec4b75a73f01a5c544c51cdbe6902f5a829d6d3/src/utils/court-utils.js#L43

src/utils/court-utils.js
sohkai: > we should perhaps error in cases where we can reach them

I'm wondering if we should not instead expect the user to make sure that they're providing valid inputs to this function

If the court's config or current term has not been loaded yet, I think avoiding a call to this function is better than calling and having to handle the `0` default.

src/utils/dispute-utils.js
sohkai: Perhaps we can add this clarification as a comment and as an explanation of why we'd use the current court term in the contracts to calculate whether they can still commit/reveal their vote.

src/utils/dispute-utils.js
sohkai: Ah yes, I remember this being discussed and I may have confused the evidence phase with the draft phase from this change: https://github.com/aragon/aragon-court/commit/68587d72337fdd56ce955a65d6231bc17dff5952.

But yes, this makes sense; selected jurors should be able to immediately commit votes!

src/utils/dispute-utils.js
sohkai: Update to this: we've clarified that the above ("if the time has passed, but the term hasn't updated, can a juror still commit/reveal their vote?") will **NOT** work. Commiting/revealing will advance the court term.

We will use the expected court term in the dispute detail view.

src/utils/court-utils.js
fabriziovigevani: Good point, updated accordingly.

src/utils/court-utils.js
```diff
@@ -28,29 +28,22 @@ export function transformCourtConfigDataAttributes(courtConfig) {
...
+  return Math.floor((now.valueOf() - firstTermStartTime) / termDuration)
+}
 
-  return termStartTime
+export function getTermStartTime(termId, { terms, termDuration }) {
+  const firstTermStartTime = getFirstTermStartTime(terms)
+  return termId * termDuration + firstTermStartTime
 }
 
 export function getTermEndTime(termId, { terms, termDuration }) {
```
sohkai, line 49: A minor nitpick on the below is that we can just return the `termEndTime` directly.

PR #341, opened 2020-05-21
Force clear activities

fabriziovigevani:
This PR is intended to automatically wipe users activities when setting and changing the court package version.

This is manly due to the breaking change we would be doing here https://github.com/aragon/court-dashboard/pull/335


src/local-settings.js
```diff
@@ -50,6 +51,20 @@ export function setIpfsGateway(gateway) {
   return setLocalSetting(IPFS_GATEWAY, gateway)
 }
 
+// The previous package version is stored in localStorage,
```
sohkai, line 54: Are we already injecting the package version into the build environment elsewhere?

src/components/Activity/ActivityProvider.js
```diff
@@ -43,6 +48,25 @@ function getStoredList(account) {
       status: SymbolsByName.get(`ACTIVITY_STATUS_${activity.status}`),
     }),
   })
+
+  // Temporary function to reset activities if package version changed
```
sohkai, line 52: This is a bit coupled to this component; I wonder if we should use a more explicit mechanism where we can detail "migrations" like this since it's not just wiping the entire cache every time.

For example, on starting up the app, we check between last and current version and do any associated "migrations" within those ranges.

src/local-settings.js
fabriziovigevani: Added the package version in all build configs now

now-mainnet.json
```diff
@@ -6,6 +6,7 @@
   "build": {
     "env": {
       "REACT_APP_CHAIN_ID": "1",
+      "REACT_APP_PACKAGE_VERSION": "1.0.0",
```
sohkai, line 9: Ah, this we usually inject this in the build process via the package.json, like so: https://github.com/aragon/aragon/blob/master/scripts/build#L5

We might have to update the build script to be a js file or use a CLI json processor (e.g. jq, if it exists in all the different build environments).

src/migrations/index.js
```diff
@@ -0,0 +1,28 @@
...
+
+export function checkMigrations() {
+  const packageVersion = getPackageVersion()
+  const lastPackageVersion = getLastPackageVersion()
+
+  if (isNewVersion(lastPackageVersion, packageVersion)) {
+    // Save the current package version
+    setPackageVersion(packageVersion)
+
+    // Run each migration which falls in the range `(lastPackageVersion, packageVersion]
```
sohkai, line 21:
```suggestion
    // Run each migration which falls in the range (lastPackageVersion, packageVersion]
```

src/migrations/index.js
```diff
@@ -0,0 +1,28 @@
...
+  getLastPackageVersion,
+  getPackageVersion,
+  setPackageVersion,
+} from '../local-settings'
+import { clearActivities } from './migrations'
+import { isNewVersion, satisifesRange } from './helper'
+
+const MIGRATIONS = {
+  '1.0.1': () => clearActivities(),
+}
```
sohkai, line 11: Not opposed to this format, but maybe an array of objects is a better fit?

fabriziovigevani: What's the clear advantage you see with the suggested format? 

Won't there be only one key value pair for each object anyways?

now-mainnet.json
fabriziovigevani: Ended up updating the `build.sh` into a js file

src/migrations/index.js
sohkai: > Won't there be only one key value pair for each object anyways?

Yep. I'm just always a bit hesitant to use something like `1.0.1` as a key—it's not wrong, but just feels a little odd.

It feels a bit more natural to do use one of these two:

```js
const MIGRATIONS = [
  ['1.0.1', () => {}],
]

const MIGRATIONS = [
  {
    version: '1.0.1',
    migration: () => {}
  },
]
```

The first is more similar to what we currently have, and reduces the `Object.keys().forEach()` into a `for of` loop, and the latter is more verbose and probably what would be used in a JSON configuration file.

scripts/build
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env node
```
sohkai, line 1: Just to double check, did we make this file executable (`chmod u+x`)?

scripts/build
```diff
@@ -0,0 +1,55 @@
...
+let enableSentry = '0'
+// Current branch
+if (process.env.NOW_GITHUB_COMMIT_REF) {
+  // ZEIT Now (no .git but an env var)
+  branch = process.env.NOW_GITHUB_COMMIT_REF
+} else if (process.env.GITHUB_REF) {
+  // Github Action
+  branch = process.env.GITHUB_REF
+} else {
+  // // Other environments
```
sohkai, line 18:
```suggestion
  // Other environments
```

scripts/build
```diff
@@ -0,0 +1,55 @@
...
+  branch = execute(`git symbolic-ref --short -q HEAD`).toString()
```
sohkai, line 19: When we do this, we may want to double check and trim any leading or trailing whitespace.

Trying this out on a repo locally, I got `'next\n'`, which will not compare well to `development` or `master` later.

src/migrations/index.js
fabriziovigevani: Thanks for the explanation!

scripts/build
fabriziovigevani: Ah good catch!

fabriziovigevani:
@sohkai Moved the `start` script to a node file as the `build` one as well so we can also set the `PACKAGE_VERSION` when running locally

scripts/start
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env node
+
+require('dotenv').config()
+
+const version = require('../package.json').version
+const execute = require('child_process').execSync
+
+let build
+// Build number from the short hash
+if (process.env.NOW_GITHUB_COMMIT_SHA) {
```
sohkai: I think we can remove the now and Github Action env vars here, as we should never be using the `start` command in those environments.

scripts/start
```diff
@@ -0,0 +1,29 @@
...
+console.log('')
+
+console.log('Syncing assets…')
+console.log('')
+execute('npm run sync-assets', {
+  stdio: 'inherit',
+})
+
+console.log('Building app…')
+console.log('')
```
sohkai, line 23: Ah, small nitpick, you can just use `console.log()` to write a newline :).

PR #342, opened 2020-05-21
Update dashboard wording

john-light:
- Fixes typos:  
  - mayority -> majority  
  - ANJ -> DAI  
- Updates `leak` action radspec description to be more accurate.

### DIFF
--- a/src/radspec.js
+++ b/src/radspec.js
@@ -6,7 +6,7 @@ import { numberToWord } from './lib/math-utils'
 
 export default {
   approveFeeDeposit: amount => {
-    return `Approve fee deposit: ${amount} ANJ`
+    return `Approve fee deposit: ${amount} DAI`
   },
   activateAnj: amount => {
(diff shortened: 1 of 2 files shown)


src/radspec.js
```diff
@@ -6,7 +6,7 @@ import { numberToWord } from './lib/math-utils'
 
 export default {
   approveFeeDeposit: amount => {
-    return `Approve fee deposit: ${amount} ANJ`
+    return `Approve fee deposit: ${amount} DAI`
```
sohkai, line 9: Would like @fabriziovigevani to triple check, but I think this is correct :).

john-light: @sohkai to clarify you think my correction is correct, or the original is correct? I created this PR because I was testing and it said to approve ANJ even though the next tx took DAI from my account. And all fees are paid in DAI.

fabriziovigevani: Yes, this is correct! It should be `DAI` as you well suggest.

PR #343, opened 2020-05-21
Notifications: Use Input hook + EmailInput component

rperez89:


src/hooks/useInput.js
```diff
@@ -0,0 +1,41 @@
...
+import { useFocusEnter, useFocusLeave } from '@aragon/ui'
+
+export function useInput(validationFn) {
+  const [value, setValue] = useState('')
+  const [status, setStatus] = useState('empty')
+
+  const inputRef = useRef()
+
+  const handleBlur = useCallback(() => {
+    const invalidValue = !validationFn(value)
```
sohkai, line 11: Doing this is kind of weird, what do you think about something like

```
setStatus(validationFn(value) ? 'valid' : 'invalid')
```

Instead? The main part I don't like about the above is `!validationFn(value)`—this is easy to miss and the variable separation makes it harder to think about what state the validation function will drive.

src/hooks/useInput.js
```diff
@@ -0,0 +1,41 @@
...
+    if (invalidValue) {
+      return setStatus('invalid')
+    }
+    return setStatus('valid')
+  }, [validationFn, value])
+
+  const handleInputFocus = useCallback(event => event.target.select(), [])
```
sohkai, line 18: I'm not sure if this is consistent with how we use other inputs; might want to double check with @dizzypaty if we should select the text when focusing.

src/hooks/useInput.js
```diff
@@ -0,0 +1,41 @@
...
+  const handleValueChange = useCallback(e => {
+    const val = e.target.value
+    setValue(val)
+    // Set only as valid while user typing. Use blur to set invalid
+    setStatus(val.trim() ? 'valid' : 'empty')
+  }, [])
+
+  return {
+    inputProps: {
+      inputRef,
```
sohkai, line 33: I'm not sure this will work as intended—if we place this into an input field, we'll need to use it as the `ref` prop

src/hooks/useInput.js
```diff
@@ -0,0 +1,41 @@
...
+      value,
```
sohkai, line 34: I think we should lift `value` out to be returned directly

src/hooks/useInput.js
```diff
@@ -0,0 +1,41 @@
...
+      onBlur: handleFocusLeave,
+      onFocus: handleFocusEnter,
+      onChange: handleValueChange,
+    },
+    status,
```
sohkai, line 39: With the above note on `value`, I think we can replace `status` with `valid`; you can tell the input's empty via a check on `value`.

src/components/EmailNotifications/EmailInput.js
```diff
@@ -0,0 +1,107 @@
...
+  existingEmail,
+  status,
+  ...inputProps
+}) {
+  const theme = useTheme()
+  const [error, setError] = useState('')
+  const [insideModal] = useInside('NotificationsModal')
+
+  useEffect(() => {
+    if (status === 'invalid') {
```
sohkai, line 23: We most likely want to use `getDerivedStateFromProps` instead of an `useEffect()` for this.

However, rather than that, I've been thinking that this component is currently at an awkward position, abstraction-wise. I see two directions we could go in:

1. Remove `status` (see comments above, also couples this component deeply with `useInput()`) and `existingEmail`, and use `valid` and `value` as props instead. This removes as much "magic" from this component as possible.
2. Encapsulate `useInput()` and the validation functions into this component, so users don't have to worry about those, and they can just consume the `onChange` value. The adds as much "magic" into the component as possible.

I was originally thinking about 2, but after some thought, 1 may be the simpler and better option, as pushing validation into input components is not something we've standardized on yet.

cc @bpierre

src/hooks/useInput.js
```diff
@@ -0,0 +1,41 @@
+import { useCallback, useRef, useState } from 'react'
+import { useFocusEnter, useFocusLeave } from '@aragon/ui'
+
+export function useInput(validationFn) {
+  const [value, setValue] = useState('')
+  const [status, setStatus] = useState('empty')
+
+  const inputRef = useRef()
```
bpierre, line 8: Should we allow users to pass the ref? I generally use this pattern: https://github.com/aragon/aragon-ui/blob/master/src/components/SidePanel/SidePanel.js#L286-L288

src/hooks/useInput.js
bpierre: I wonder if it should be here, even if we want this behavior? What would you guys think of renaming it `useInputValidation()` and keeping it specialized?

src/hooks/useInput.js
```diff
@@ -0,0 +1,41 @@
...
+
+  const { handleFocusLeave } = useFocusLeave(handleBlur, inputRef)
+
+  const { handleFocusEnter } = useFocusEnter(handleInputFocus, inputRef)
+
+  const handleValueChange = useCallback(e => {
+    const val = e.target.value
+    setValue(val)
+    // Set only as valid while user typing. Use blur to set invalid
+    setStatus(val.trim() ? 'valid' : 'empty')
```
bpierre, line 28: Maybe not now, but an improvement could be to also pass a `filter` function, defaulting to `v => v.trim()`. We would also export this filtered value, alongside the unaltered one.

src/hooks/useInput.js
bpierre, line 18: Should we also do `setStatus(val.trim() ? 'valid' : 'empty')` here?

src/components/EmailNotifications/EmailInput.js
bpierre: > We most likely want to use `getDerivedStateFromProps` instead of an `useEffect()` for this.

I think a good equivalent for function components would be `useMemo()`?

```jsx
const error = useMemo(() => {
  return status === 'invalid'? 'Please enter…' : ''
}, [status])
```

Or an expression directly:

```jsx
const error = status === 'invalid'? 'Please enter…' : ''
```

> I was originally thinking about 2, but after some thought, 1 may be the simpler and better option, as pushing validation into input components is not something we've standardized on yet.

Agreed, 1 seems reasonable as a first step. Having 2 would be nice, but there are many ways we can do it, and we might want to do it at the `TextInput` level first.

src/hooks/useInput.js
rperez89: should i remove the input focus function maybe?

src/components/EmailNotifications/EmailInput.js
rperez89: so to sumarize i shouldn't return any status from here?

src/hooks/useInput.js
bpierre: I think we’ll still need it for https://github.com/aragon/court-dashboard/pull/343/files#r430469262

PR #352, opened 2020-05-28
Notifications: Modifications requested by Jhon

### LINKED ISSUE #351: Notifications disable option resets after refresh
1) Open Notification settings
2) Click disable:
image
3) Refresh the page:
image

rperez89:
Screen Shot 2020-05-28 at 4 47 07 PM

Screen Shot 2020-05-28 at 11 46 27 AM

Screen Shot 2020-05-28 at 11 47 14 AM

Screen Shot 2020-05-28 at 11 48 14 AM

Screen Shot 2020-05-28 at 12 33 40 PM

also closes #351

### DIFF
--- a/src/components/EmailNotifications/actions.js
+++ b/src/components/EmailNotifications/actions.js
@@ -4,7 +4,6 @@ import {
   resendVerificationEmail,
   subscribeExistingEmail,
   subscribeToNotifications,
-  switchNotificationsStatus,
 } from '../../services/notificationServiceApi'
 
 import {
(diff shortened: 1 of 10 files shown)


src/components/EmailNotifications/actions.js
```diff
@@ -45,8 +44,6 @@ export const actions = {
     requiresEmail: false,
   },
   [OPTOUT_ACTION]: {
-    request: switchNotificationsStatus,
-    params: [true], // disabled = true
     requiresEmail: false,
```
sohkai, line 47: Double checking here, what happens when the user presses "no, thanks"? Don't we still need to send an API request?

sohkai: Oh, I see; what we're doing is just verifying their address and not setting anything else?

PR #355, opened 2020-06-03
Balance Module: Update latest movement message when not effective

### LINKED ISSUE #340: Balance is shown as active even though it isn't counted as active in the court
In the following screenshot, my active balance is shown as 10,000,000 ANJ (on the usability test network) and my drafting odds are shown as <1 in 100+ cases even though with an active balance of 10 million ANJ my odds should really be like 1 in 2 cases.

Screenshot_2020-05-20 Aragon Court Dashboard(1)

I was told that this is probably because the court will not recognize my balance as actually being active until the next term after I activated my tokens begins. And sure enough, once the next term began my odds shown changed.

Screenshot_2020-05-20 Aragon Court Dashboard(16)

I think to prevent confusion about this we should indicate some "pending activation" state that will change to "activated" once the next term begins.

fabriziovigevani:
fixes #340 

Show a `Effective next term` message when the movement is immediate and not yet effective.

By immediate we mean that the movement had an immediate impact on the total balance.

E.g for activating tokens, the active balance is already increased, the increased amount is not yet effective in that is not going to be taken into account for drafting. 

**Preview**

Activating

**Not effective**
Screen Shot 2020-06-02 at 9 56 07 PM

**Effective**

Screen Shot 2020-06-02 at 9 59 31 PM

### DIFF
--- a/src/components/Dashboard/Balance.js
+++ b/src/components/Dashboard/Balance.js
@@ -17,13 +17,13 @@ import ANJIcon from '../../assets/IconANJ.svg'
 import lockIcon from '../../assets/IconLock.svg'
 
 const Balance = React.memo(function Balance({
-  label,
   amount,
-  loading,
   actions,
-  mainIcon,
   activity,
   distribution,
+  label,
+  loading,
+  mainIcon,
   mainIconBackground,
 }) {
   const theme = useTheme()
@@ -76,11 +76,11 @@ const Balance = React.memo(function Balance({
             </div>
             <div>
               <span
-                css={`      
-                ${textStyle('body2')}
-                color: ${theme.surfaceContentSecondary};
-                display:block;
-              `}
+                css={`
+                  ${textStyle('body2')};
+                  color: ${theme.surfaceContentSecondary};
+                  display: block;
+                `}
               >
                 {label}
               </span>
@@ -99,10 +99,10 @@ const Balance = React.memo(function Balance({
               </div>
               <span
                 css={`
-                ${textStyle('body4')}
-                color: ${theme.surfaceContentSecondary};
-                display:block;
-              `}
+                  ${textStyle('body4')};
+                  color: ${theme.surfaceContentSecondary};
+                  display: block;
+                `}
               >
                 $ {convertedAmount}
               </span>
@@ -218,7 +218,9 @@ const LatestActivity = ({ activity, distribution }) => {
             color: ${theme.content};
           `}
         >
-          {convertToString(activity.type, activity.direction)}
+          {activity.isImmediate && !activity.isEffective
+            ? 'Effective next term'
+            : convertToString(activity.type, activity.direction)}
         </span>
       </div>
       {distribution && <ANJLockedHelp distribution={distribution} />}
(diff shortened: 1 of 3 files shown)


src/components/Dashboard/Balance.js
```diff
@@ -218,7 +218,9 @@ const LatestActivity = ({ activity, distribution }) => {
             color: ${theme.content};
           `}
         >
-          {convertToString(activity.type, activity.direction)}
+          {activity.isImmediate && !activity.isEffective
+            ? 'Effective next term'
+            : convertToString(activity.type, activity.direction)}
```
sohkai, line 223: I'm wondering if we should have a tooltip for the `Active` part of the balance module when it's `Effective next term`.

I can still see a user being confused about their "immediate" active balance vs. their "effective" active balance at the moment, as we only show one value in the module.

src/components/Dashboard/Balance.js
```diff
@@ -218,7 +218,9 @@ const LatestActivity = ({ activity, distribution }) => {
             color: ${theme.content};
           `}
         >
-          {convertToString(activity.type, activity.direction)}
+          {activity.isImmediate && !activity.isEffective
+            ? 'Effective next term'
```
sohkai, line 222: Wondering if we should have different text for the `Inactive` part of the balance module to clarify where the inactive portion has moved (to be withdrawal or activated).

fabriziovigevani: hmm yeah perhaps doesn't make sense to show the `effective next term` message there since is not something that is relevant for that balance and we can just return the `Activated`.

fabriziovigevani:
> In the "Not effective" screenshot, where it says +9490 ANJ Effective next term is this 9490 ANJ included in the balance of Active: 9990 ANJ shown directly above? Or will the 9490 ANJ only be added to the Active balance in the next term? (and then the Active balance will be 19480 ANJ)

It is included already in the balance yes.

> In the "Effective" screenshot, under the "Inactive" column it says -9490 ANJ Activated. Should it say "Activated" here? Or Deactivated? And should it show a - sign or a + sign, if the token balance is getting added to this column? (Similar question for the -10,000 ANJ Withdrawal shown on the "My Wallet" column of this screenshot.)

The `-` a `+` signs reflects how the latest movement affected that balance. In the case of the inactive balance it shows the `-` because the `9490 ANJ` went from the inactive to the active balance.

Regarding the message, the intention of it is to show the reason of why the balance increased/decreased and in this the inactive balance decreased by `9490 ANJ` because it was activated.

Do you see that a little confusing?

Regarding the wallet balance movement, in our local environment we run a script which does two txs to initiate balances, first moves the amount from the wallet balance to the inactive balance(and that's why it's showing the withdrawal) before moving it to the active balance. (In production this would not occur) since the dashboard only allows moving from wallet balance to active balance in one tx.
But in the rare case where someone would create txs outside the dashboard and does a staking tx, then the amount would move from the wallet balance to the inactive balance and that's the movement you would see.

john-light:
> It is included already in the balance yes.

I think the ANJ should only be included in the active balance once it is "effective".

> The `-` and `+` signs reflects how the latest movement affected that balance. In the case of the inactive balance it shows the `-` because the `9490 ANJ` went from the inactive to the active balance.
> 
> Regarding the message, the intention of it is to show the reason of why the balance increased/decreased and in this the inactive balance decreased by `9490 ANJ` because it was activated.
>
> Do you see that a little confusing?

It is a bit confusing but perhaps only because I lacked context for why the ANJ was being deducted from the Inactive column. We can leave this part as-is for now.

Issue #349, opened 2020-05-26
Unable to prematurely dismiss onboarding modal


andy-hook:
![image](https://user-images.githubusercontent.com/11708259/82878554-8c0c8280-9f33-11ea-958c-d5dbbc85d435.png)

Hi folks, I spotted a minor defect with this modal. I'm unable to dismiss the flow by clicking the background scrim or the close button in the top right. The modal also sits vertically off-center in Safari.

fabriziovigevani:
Hmm i think it's not supposed to be closable so we should definitely remove the cross icon @rperez89 ?

> The modal also sits vertically off-center in Safari.

Will look into it 👍

sohkai:
Yes, it's not supposed to be closable.

I have the feeling we've tackled this issue a few times, to avoid throwing react errors, but perhaps we've forgotten to remove the cross over time 😅

Evalir:
Done in #358 .

PR #358, opened 2020-06-03
AccountModule: update "provider" copy to move towards wallets and accounts

### LINKED ISSUE #159: AccountModule: update "provider" copy to move towards wallets and accounts

Evalir:
Closes #159, #349.

Also solves a bug in the Onboarding Modal: clicking the Close button (The X) doesn't close the modal.

Screenshots:
- Providers before:
Screen Shot 2020-06-03 at 10 23 13 AM
- Providers after (only two providers as I'm running it locally):
Screen Shot 2020-06-03 at 10 22 38 AM

- Connection error before:
Screen Shot 2020-06-03 at 10 22 54 AM
- Connection error after:
Screen Shot 2020-06-03 at 10 22 47 AM

- Connected before:
Screen Shot 2020-06-03 at 10 23 52 AM
- Connected after:
Screen Shot 2020-06-03 at 10 23 29 AM

(Note that it kinda bugs me that we're using "account now" but we have "Disconnect wallet". It's accurate, but I think we could also change this to "Disconnect account")

### DIFF
--- a/src/components/Onboarding/Navigation.js
+++ b/src/components/Onboarding/Navigation.js
@@ -8,6 +8,7 @@ const buttonTransitionStyles = show => ({
   opacity: Number(show),
   transform: `translate3d(0, ${show ? 0 : 2}px, 0)`,
   config: springs.swift,
+  pointerEvents: show ? 'auto' : 'none',
 })
 
 const Navigation = ({ step, steps, onPrev, onNext }) => {
--- a/src/components/Onboarding/OnboardingModal.js
+++ b/src/components/Onboarding/OnboardingModal.js
@@ -34,7 +34,7 @@ const OnboardingModal = React.memo(function OnboardingModal({
         const compactMode = width < 500 || height < 400
         return (
           <Modal
-            onClose={onComplete}
+            closeButton={false}
             padding={0}
             width={Math.min(1055, width - 40)}
             visible={visible}
--- a/src/components/Onboarding/OnboardingModal.js
+++ b/src/components/Onboarding/OnboardingModal.js
@@ -34,6 +34,7 @@ const OnboardingModal = React.memo(function OnboardingModal({
         const compactMode = width < 500 || height < 400
         return (
           <Modal
+            onClose={onComplete}
             padding={0}
             width={Math.min(1055, width - 40)}
             visible={visible}
--- a/src/components/Account/AccountModule.js
+++ b/src/components/Account/AccountModule.js
@@ -13,7 +13,7 @@ import ScreenProviders from './ScreenProviders'
 const SCREENS = [
   {
     id: 'providers',
-    title: 'Ethereum providers',
+    title: 'Use account from',
     height:
       4 * GU + // header
       (12 + 1.5) * GU * Math.ceil(getUseWalletProviders().length / 2) + // buttons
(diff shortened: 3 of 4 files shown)


src/components/Onboarding/OnboardingModal.js
```diff
@@ -34,6 +34,7 @@ const OnboardingModal = React.memo(function OnboardingModal({
         const compactMode = width < 500 || height < 400
         return (
           <Modal
+            onClose={onComplete}
```
fabriziovigevani, line 37: Ah note that the intention here is for the Modal to be not closable see https://github.com/aragon/court-dashboard/issues/349#issuecomment-638025200. We should actually remove the cross icon :D

fabriziovigevani:
> (Note that it kinda bugs me that we're using "account now" but we have "Disconnect wallet". It's accurate, but I think we could also change this to "Disconnect account")

Yeah i agree, i would advocate to for consistency and change every `wallet` to `account`.

sohkai:
> I think we could also change this to "Disconnect account"

Agreed as well; let's change this here and in the client!

Evalir:
I think "Disconnect" can work. I still think that "Disconnect account" looks better, due to the consistency factor.

> I can understand why we used "wallet"—technically the account comes from your wallet, and you can change your account outside of the app via the wallet.

Yes, that's why I say it's accurate wording. Still, if we're shifting away from using "wallet", I don't think it makes sense to keep it for this very particular case.

src/components/Onboarding/OnboardingModal.js
Evalir: Ah, cool! Will remove it then, makes sense due to how the `onComplete` function behaves. 👍

src/components/Onboarding/Navigation.js
```diff
@@ -8,6 +8,7 @@ const buttonTransitionStyles = show => ({
   opacity: Number(show),
   transform: `translate3d(0, ${show ? 0 : 2}px, 0)`,
   config: springs.swift,
+  pointerEvents: show ? 'auto' : 'none',
```
Evalir, line 11: Another little fix here: The modal buttons have pointer-events even if they are not being shown as only the opacity changes, and this prevents this behavior. The upgrade organization modal from the client also had this, and during spring cleaning [we applied the same fix](https://github.com/aragon/aragon/blob/master/src/components/Upgrade/Navigation.js#L10).

sohkai:
> Still, if we're shifting away from using "wallet"

Ah, the intention was actually to shift towards using "wallet" instead of "provider".

A "wallet" holds a number of "accounts".

sohkai:
> I still think that "Disconnect account" looks better, due to the consistency factor.

Let's punt on this for later, as we work on our nomenclature for specifying ethereum providers / wallets / accounts.

Can we merge this now?

rperez89:
> (Note that it kinda bugs me that we're using "account now" but we have "Disconnect wallet". It's accurate, but I think we could also change this to "Disconnect account")

ahh yes that was my bad, the intention was to start using account everywhere but i forgot about the disconnect

Evalir:
> Can we merge this now?

There is still changes to do left (the height measuring, mainly) which I'll get to very soon as we're finishing up the local connection fixes, but as this PR is mainly concerned about copy aligning, I believe we can merge this and continue on another one.

sohkai:
> as this PR is mainly concerned about copy aligning, I believe we can merge this and continue on another one

Let's merge (since this has fixed the bug with the onboarding modal, and we should've merged that a _long_ time ago 😄), and please raise another issue (and/or the PR) for finishing the updates to the account modal!

PR #360, opened 2020-06-03
Balance module: Resize amounts fractional part font

fabriziovigevani:


fabriziovigevani:
In the balance module designs we don't have amounts with fractional part. But following the ones of the [Court stats](https://www.figma.com/file/IdVA6iMn0cWLV5imylwEUP/%E2%9A%96%EF%B8%8F-Court-Dashboard-Master?node-id=3628%3A23834) there isn't a difference in size between whole and fraction.

sohkai:
Gotcha. Actually in some of those designs, I noticed that the inactive panel has a smaller font size (`24px`) than the wallet and active panels' (`28px`).

Not sure if that was intentional—it's only in some of the screens—but perhaps something to double check!

fabriziovigevani:
Good eye!

Already ping @dizzypaty about it, will merge this one in the meantime and we can do another PR in case we need to update it.

PR #362, opened 2020-06-04
Fix typo and clarify wording

john-light:
- "successfuly" -> "successfully"
- Added "and revealed" so that jurors have an explicit acknowledgement that the reveal step worked properly
- Change "majority" to "plurality", because Aragon Court rulings are based on plurality rule, not majority rule.

### DIFF
--- a/src/components/Disputes/DisputeActions.js
+++ b/src/components/Disputes/DisputeActions.js
@@ -282,7 +282,7 @@ const useInfoAttributes = ({
 
     // Juror has voted and reveal period hasn't ended
     return {
-      title: 'Your vote was cast successfuly.',
+      title: 'Your vote was cast and revealed successfully.',
       paragraph: (
         <VoteInfo
           commitmentDate={jurorDraft.commitmentDate}
--- a/src/components/Disputes/DisputeActions.js
+++ b/src/components/Disputes/DisputeActions.js
@@ -282,7 +282,9 @@ const useInfoAttributes = ({
 
     // Juror has voted and reveal period hasn't ended
     return {
-      title: 'Your vote was cast and revealed successfully.',
+      title: `Your vote was cast ${
+        jurorDraft.outcome ? 'and revealed' : ''
+      } successfully.`,
       paragraph: (
         <VoteInfo
           commitmentDate={jurorDraft.commitmentDate}
--- a/src/components/Disputes/DisputeActions.js
+++ b/src/components/Disputes/DisputeActions.js
@@ -239,7 +239,7 @@ const useInfoAttributes = ({
       }
 
       // Juror has revealed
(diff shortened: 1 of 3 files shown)


src/components/Disputes/DisputeActions.js
```diff
@@ -282,7 +282,7 @@ const useInfoAttributes = ({
 
     // Juror has voted and reveal period hasn't ended
     return {
-      title: 'Your vote was cast successfuly.',
+      title: 'Your vote was cast and revealed successfully.',
```
sohkai, line 285: Not sure, but I think at this point the juror might not have revealed yet?

src/components/Onboarding/content.js
```diff
@@ -104,8 +104,8 @@ highlights.rinkeby = [
       small: null,
       large: (
         <span>
-          When you vote in favor of the majority ruling, you will be rewarded
-          with arbitration fees. Just note that these rewards will be awarded in
+          When you vote in favor of the plurality ruling, you will be rewarded
```
sohkai, line 107: I'm not opposed to it, but we've used "majority" in quite a few places to describe this mechanism.

And while I agree "plurality" is the technically more correct term, it's also less familiar to non-English speakers.

src/components/Disputes/DisputeActions.js
fabriziovigevani: Yeah we should show this last part conditionally based on `jurorDraft.outcome`. 
I can push the change, want to check also if the length of the text wouldn't make it look weird. I think is short enough but just in case.

src/components/Onboarding/content.js
john-light: > I'm not opposed to it, but we've used "majority" in quite a few places to describe this mechanism.

I already removed this term from all of the help.aragon.org documentation (e.g. [this section](https://help.aragon.org/article/41-aragon-court#plurality)) because it is not correct. It was a mistake to ever describe Aragon Court as a majority-rule system. This PR also fixes the terminology throughout the Dashboard.

> And while I agree "plurality" is the technically more correct term, it's also less familiar to non-English speakers.

The other equally-accurate term would be to say "relative majority", which I decided against because 1) most people would also have to look up what that means, and 2) it's two words and that takes up more space and feels awkward to write over and over again: 

> Aragon Court is a _relative majority-rule_ system.... You have ruled with the _relative majority_... Make sure to vote the way that you believe the _relative majority_ of jurors will vote...

vs

> Aragon Court is a plurality-rule_ system.... You have ruled with the _plurality_... Make sure to vote the way that you believe the _plurality_ of jurors will vote...

src/components/Disputes/DisputeActions.js
fabriziovigevani: Just pushed this last bit.

PR #364, opened 2020-06-05
Tasks: Use court clock

### LINKED ISSUE #319: "Status" column seems redundant
Noticed that the Status column says "due in 2 days" while the due date column is right next to it and also says 1 day and 7 hours... contradictory no?

It also seems that the Status column only has two states: "open" and not visible at all, because the task has been done or is past due.

Proposal: I think the Status column here is redundant and probably unnecessary. The due date already has its own column, and if a task is "open" we can just show it on this tab, and if it's not open then we don't show it on this tab, so there is no need for a new column just to say that the task is "open".

Please correct me if I am wrong about anything here :) 

https://court.aragon.org/tasks

Screenshot_2020-05-12 Aragon Court Dashboard(1)

fabriziovigevani:
closes #319 

- Removes redundant status from task table
- Uses court clock for updating tasks

While removing the redundant status, realized that we forgot to update the usage of the clock by tasks as well when implementing https://github.com/aragon/court-dashboard/pull/338 😅

### DIFF
--- a/src/hooks/subscription-hooks.js
+++ b/src/hooks/subscription-hooks.js
@@ -23,7 +23,10 @@ import { bigNum } from '../lib/math-utils'
 import { dayjs } from '../utils/date-utils'
 import { groupMovements } from '../utils/anj-movement-utils'
 import { transformAppealDataAttributes } from '../utils/appeal-utils'
-import { transformDisputeDataAttributes } from '../utils/dispute-utils'
+import {
+  transformRoundDataAttributes,
+  transformDisputeDataAttributes,
+} from '../utils/dispute-utils'
 import { transformJurorDataAttributes } from '../utils/juror-draft-utils'
 import { transformClaimedFeesDataAttributes } from '../utils/subscription-utils'
 import {
@@ -330,7 +333,8 @@ export function useTasksSubscription() {
     variables: subscriptionVariables,
   })
 
-  const tasks = data?.adjudicationRounds || null
+  const tasks =
+    data?.adjudicationRounds.map(transformRoundDataAttributes) || null
 
   return { tasks, fetching: !data && !error, error }
 }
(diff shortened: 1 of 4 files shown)


src/hooks/useTasks.js
```diff
@@ -52,23 +57,26 @@ function useOpenTasks(tasks, now, courtSettings) {
...
+        currentPhase !== DisputePhase.AppealRuling &&
+        currentPhase !== DisputePhase.ConfirmAppeal
       ) {
         for (let j = 0; j < incompleteTasks[i].jurors.length; j++) {
           if (isVotingTaskOpen(incompleteTasks[i].jurors[j], currentPhase)) {
             openTasks.push({
               number: incompleteTasks[i].number,
               state: incompleteTasks[i].state,
-              createdAt: parseInt(incompleteTasks[i].createdAt, 10) * 1000,
+              createdAt: incompleteTasks[i].createdAt,
```
sohkai, line 74: Just wanted to sanity check that this was correct, since it didn't look like we changed `createdAt` in this PR.

Was this changed previously and we just forgot to update it here?

src/hooks/subscription-hooks.js
```diff
@@ -330,7 +333,8 @@ export function useTasksSubscription() {
     variables: subscriptionVariables,
   })
 
-  const tasks = data?.adjudicationRounds || null
+  const tasks =
+    data?.adjudicationRounds.map(transformRoundDataAttributes) || null
```
fabriziovigevani, line 337: See that we use the same round data transformations as we do for disputes.

src/hooks/useTasks.js
fabriziovigevani: Ah yeah i moved the tasks data transformation to the `subscription-hooks https://github.com/aragon/court-dashboard/pull/364/files#r436175026

src/hooks/subscription-hooks.js
sohkai: Ah gotcha, this wasn't being transformed before at all?

fabriziovigevani: No it wasn't.

PR #366, opened 2020-06-09
Rewards module improvements

### LINKED ISSUE #235: Rewards: small improvements
See changes proposed in https://www.notion.so/aragonone/Improvements-to-rewards-module-2628acaeeaa344ed8ec3d7281b521e8e.

### LINKED ISSUE #356: Rewards module: Break down rewards by disputes and appeals even after rewards settled
Currently, we can break down rewards by dispute fees, appeal fees, etc, for rounds whose penalties have not been settled yet. After they are settled, the rewards are already transferred to the treasury.

It would be great to be able to have the break down even after the settlement. 
Technically ,this requires adding a settlement timestamp to the the graph schema for the round entity and fetch all rounds whose settlement timestamp is after the user's last withdraw from the treasury.

fabriziovigevani:
closes #356 closes #235 

This PR requires https://github.com/aragon/court-subgraph/pull/62 and  https://github.com/aragon/court-subgraph/pull/65 to be deployed first.

In this PR we include already settled rewards to be taken into account for the Rewards module, with the condition that the time at which they were settled is after the time at which jurors did their last withdrawal from the treasury.

With this new implementation jurors can know exactly where do their rewards come from even after they were settled. 

Regarding [the notion item](https://www.notion.so/9e7e5cd565b44800a79b5640a5f094a8?v=d3b5e76c06634dafa4b7083edbaca881&p=2628acaeeaa344ed8ec3d7281b521e8e), we removed the Total divider, but included a new line separator as i felt the need to have some separation between the total and the partial reward amounts. 

**Preview**

Screen Shot 2020-06-08 at 9 04 29 PM

### DIFF
--- a/src/components/Dashboard/RewardsModule.js
+++ b/src/components/Dashboard/RewardsModule.js
@@ -18,23 +18,29 @@ import { useCourtConfig } from '../../providers/CourtConfig'
 import useJurorSubscriptionFees from '../../hooks/useJurorSubscriptionFees'
 
 import { addressesEqual } from '../../lib/web3-utils'
+import {
+  getTotalFees,
+  getTotalSettledFees,
+  filterSettledRounds,
+} from '../../utils/rewards-utils'
 import { bigNum, formatTokenAmount } from '../../lib/math-utils'
 
 const useTotalFeeRewards = (arbitrableFees, appealFees, subscriptionFees) => {
   return useMemo(() => {
-    const totalArbitrableFees = Array.isArray(arbitrableFees)
-      ? arbitrableFees.reduce((acc, fee) => acc.add(fee.amount), bigNum(0))
-      : bigNum(0)
+    const [totalArbitrableFees, totalAppealFees, totalSubscriptionFees] = [
+      arbitrableFees,
+      appealFees,
+      subscriptionFees,
+    ].map(fees => getTotalFees(fees))
 
-    const totalAppealFees = Array.isArray(appealFees)
-      ? appealFees.reduce((acc, fee) => acc.add(fee.amount), bigNum(0))
-      : bigNum(0)
+    const totalSettledFees = getTotalSettledFees(arbitrableFees, appealFees)
 
-    const totalSubscriptionFees = Array.isArray(subscriptionFees)
-      ? subscriptionFees.reduce((acc, fee) => acc.add(fee.amount), bigNum(0))
-      : bigNum(0)
-
-    return { totalArbitrableFees, totalAppealFees, totalSubscriptionFees }
+    return {
+      totalAppealFees,
+      totalArbitrableFees,
+      totalSettledFees,
+      totalSubscriptionFees,
+    }
   }, [appealFees, arbitrableFees, subscriptionFees])
 }
 
@@ -57,8 +63,9 @@ const RewardsModule = React.memo(function RewardsModule({
   const { anjRewards, feeRewards } = rewards || {}
 
   const {
-    totalArbitrableFees,
     totalAppealFees,
+    totalArbitrableFees,
+    totalSettledFees,
     totalSubscriptionFees,
   } = useTotalFeeRewards(
     feeRewards?.arbitrableFees,
@@ -78,7 +85,10 @@ const RewardsModule = React.memo(function RewardsModule({
   const totalDisputesFees = totalArbitrableFees.add(totalAppealFees)
 
   // All dispute fees are sent to the treasury after being settled
-  const totalTreasuryFees = totalDisputesFees.add(treasuryBalance)
+  // Note that the total dispute fees could include already settled fees so we must discount them from the total treasury fees
+  const totalTreasuryFees = totalDisputesFees
+    .sub(totalSettledFees)
+    .add(treasuryBalance)
   const totalFeeRewards = totalTreasuryFees.add(totalSubscriptionFees)
 
   // Form submission
@@ -88,10 +98,15 @@ const RewardsModule = React.memo(function RewardsModule({
 
       if (!rewards) return
 
+      const nonSettledArbitrables = filterSettledRounds(
+        feeRewards.arbitrableFees
+      )
+      const nonSettledAppeals = filterSettledRounds(feeRewards.appealFees)
+
       onClaimRewards(
         wallet.account,
-        feeRewards.arbitrableFees,
-        feeRewards.appealFees,
+        nonSettledArbitrables,
+        nonSettledAppeals,
         totalTreasuryFees,
         subscriptionFees,
         feeToken.id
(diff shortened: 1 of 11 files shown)


src/components/Dashboard/RewardsModule.js
```diff
@@ -78,7 +85,10 @@ const RewardsModule = React.memo(function RewardsModule({
   const totalDisputesFees = totalArbitrableFees.add(totalAppealFees)
 
   // All dispute fees are sent to the treasury after being settled
-  const totalTreasuryFees = totalDisputesFees.add(treasuryBalance)
+  // Note that the total dispute fees could include already settled fees so we must discount them from the total treasury fees
+  const totalTreasuryFees = totalDisputesFees
+    .sub(totalSettledFees)
+    .add(treasuryBalance)
```
sohkai, line 91: Quick question on the treasury: it looks like its an array, so it must hold a number of tokens? Is this due to the fee token potentially changing, or is it because there could be multiple different tokens held by the treasury (e.g. ANJ and DAI)?

src/components/Dashboard/RewardsModule.js
```diff
@@ -88,10 +98,15 @@ const RewardsModule = React.memo(function RewardsModule({
...
+        feeRewards.arbitrableFees
+      )
+      const nonSettledAppeals = filterSettledRounds(feeRewards.appealFees)
+
       onClaimRewards(
         wallet.account,
-        feeRewards.arbitrableFees,
-        feeRewards.appealFees,
+        nonSettledArbitrables,
+        nonSettledAppeals,
```
sohkai, line 109: Just a bit worried from the comment before in the utils; it looks like the data types for these variables may have changed?

src/queries/juror.js
```diff
@@ -9,6 +9,21 @@ export const JurorFeesClaimed = gql`
   }
 `
 
+// Last juror fee withdrawal movement
+export const JurorLastFeeWithdrawal = gql`
```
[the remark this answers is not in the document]
fabriziovigevani: Yes exactly.

src/utils/rewards-utils.js
```diff
@@ -0,0 +1,22 @@
...
+    .map(fees => getTotalFees(fees, 'settledAmount'))
+    .reduce((acc, amount) => acc.add(amount), bigNum(0))
+}
+
+export function filterSettledRounds(fees) {
+  return fees.map(({ rounds, ...fee }) => ({
+    ...fee,
+    rounds: rounds
+      .filter(({ settled }) => !settled)
+      .map(({ roundId }) => roundId),
```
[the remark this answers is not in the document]
fabriziovigevani: True, will change.

src/hooks/subscription-hooks.js
```diff
@@ -361,3 +361,26 @@ export function useTotalRewardsSubscription() {
...
+/**
+ * Queries for the last withdrawal fee movement time made by the given juror
+ * @param {String} jurorId Address of the juror
+ * @returns {Object} Juror's last withdrawal fee movement
+ */
+export function useJurorLastWithdrawalTimeSubscription(jurorId) {
+  const [{ data }] = useSubscription({
+    query: JurorLastFeeWithdrawal,
+    variables: { owner: jurorId?.toLowerCase() },
+    pause: !jurorId,
```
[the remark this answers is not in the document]
fabriziovigevani: If you switch accounts it would resubscribe again as usual.

> I was imagining you would just not subscribe if the jurorId wasn't given.

Yeah that's the intention of the `pause` attribute!

src/hooks/subscription-hooks.js
```diff
@@ -361,3 +361,26 @@ export function useTotalRewardsSubscription() {
 
   return { data: rewards, error }
 }
+
+/**
+ * Queries for the last withdrawal fee movement time made by the given juror
+ * @param {String} jurorId Address of the juror
+ * @returns {Object} Juror's last withdrawal fee movement
```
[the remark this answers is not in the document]
fabriziovigevani: Ah good catch! Initially i was returning the whole movement object and after deciding to just return the timestamp, forgot to update it.

src/hooks/useJurorRewards.js
```diff
@@ -9,28 +9,36 @@ import {
...
+  const lastWithdrawalTime = useJurorLastWithdrawalTimeSubscription(
+    wallet.account
+  )
 
   // For arbitrable and appeal fees we will use a map where map = [disputeId, { amount, rounds }]
   // Where rounds::Set
-  // This is useful since it can happen that a juror has rewards from many rounds on the same dispute
-  // We also need the data set this way so it's easier to claim rewards later
+  // This is useful as jurors could have rewards from many rounds for the same dispute
+  // This data set also helps to claim rewards transparently.
```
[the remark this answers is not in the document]
fabriziovigevani: Ah i think i will remove it or rephrase it since now we do have to do a small conversion on rounds before passing them to `useCourtContracts`

src/components/Dashboard/RewardsModule.js
fabriziovigevani: Yes, it's the former. For example if a juror has some rewards in DAI that has not withdrawn form the treasury and the `feeToken` changes. We have a `TODO` there for handling potentially multiple tokens but probably we will need designs? Or we could just do a divider for every token.

src/components/Dashboard/RewardsModule.js
fabriziovigevani: Ah no actually the conversion is needed since we changed the data structure for the rounds in `useJurorRewards`.

Before rounds were just an array of `roundId` and now they are an object of `{roundId, settled}` so we needed to convert it back to just an array of `roundId`s. 

Or maybe we could just change how they are handled in `useCourtContracts` so we don't need to do this re conversion ?

fabriziovigevani:
Thanks for the comments!

>  The amounts total to 516 DAI, not 520 DAI.

Ah this is because i did a lot of draftings from this account and got the drafting fees sent directly to the treasury. Sorry for the confusion

src/components/Dashboard/RewardsModule.js
sohkai: Gotcha, let's keep it as a todo (perhaps we can improve the TODO A bit :smile:)

fabriziovigevani: Will do 😄

src/hooks/subscription-hooks.js
```diff
@@ -361,3 +361,26 @@ export function useTotalRewardsSubscription() {
...
+
+  if (!data) {
+    return null
+  }
+
+  if (data.feeMovements.length === 0) {
+    return -1
+  }
+
+  return data.feeMovements[0].createdAt * 1000
```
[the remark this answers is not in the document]
fabriziovigevani: Yeah true, probably makes sense to handle it at the subgraph level yeah, cc @facuspagnuolo 

For now i'll create a utility function to use on every place required to transition to ms.

src/components/Dashboard/RewardsModule.js
fabriziovigevani: Ended up removing the already settled roundIds from the fees data structure as we don't really need them. That way we don't need to do the re conversion either.

src/hooks/subscription-hooks.js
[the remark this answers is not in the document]
sohkai: Made an issue for this: https://github.com/aragon/court-subgraph/issues/66

src/components/Dashboard/RewardsModule.js
```diff
@@ -422,4 +394,35 @@ const FeeSection = ({ children }) => {
...
+    : bigNum(0)
+}
+
+function getTotalSettledFees(arbitrableFees, appealFees) {
+  return [arbitrableFees, appealFees]
+    .map(fees => getTotalFees(fees, 'settledAmount'))
+    .reduce((acc, amount) => acc.add(amount), bigNum(0))
+}
+
+const useTotalFeeRewards = (arbitrableFees, appealFees, subscriptionFees) => {
```
sohkai, line 409: We should probably default each variable to `[]`; it looks like from above that we're using `?.` and a null value could be given.

PR #368, opened 2020-06-11
Dispute detail: request auto reveal service

fabriziovigevani:
Includes the option to request the auto reveal service for the current dispute directly from the `DisputeDetail` 

This is useful in the strange cases where the auto-reveal service request fails at the moment jurors are committing their vote.
It also provides the option for jurors that decided to `opt-out` from the service and changed their mind, to be able to request it.

**Preview**

Screen Shot 2020-06-10 at 8 50 53 PM

Screen Shot 2020-06-10 at 8 51 21 PM

 **Decided to add this info indicator so the juror can rest asure the auto reveal has been requested.**
Screen Shot 2020-06-10 at 9 15 35 PM


src/hooks/useCourtContracts.js
```diff
@@ -206,6 +210,25 @@ export function useDisputeActions() {
     [disputeManagerContract, processRequests]
   )
 
+  const autoReveal = useCallback(
```
sohkai, line 213: Should we add a comment header here, since it doesn't really belong under "draft jurors"? Maybe "request auto reveal"?

src/components/Disputes/DisputeAutoReveal.js
```diff
@@ -0,0 +1,119 @@
...
+      )}
+    </form>
+  )
+}
+
+function useAutoRevealPolling(account, disputeId, roundId) {
+  const [autoRevealRequested, setAutoRevealRequested] = useState(false)
+  const timer = 3000
+
+  // We start the timeout timer in 0 in order to immediately make the api call
```
sohkai, line 74:
```suggestion
  // We start the timeout timer at 0 to immediately make the api call on render
```

But do we need to use a ref here? Couldn't we use a local variable in the closure? (E.g. we could provide as an argument to `fetchAutoReveal()`).

src/utils/crvoting-utils.js
```diff
@@ -158,3 +158,34 @@ export function isvoteLeaked(outcome) {
...
+export function removeCodeFromLocalStorage(connectedAccount, disputeId) {
+  localStorage.removeItem(`oneTimeCode:${connectedAccount}:${disputeId}`)
+}
+
+// Auto reveal service preference
+export function getAutoRevealPreference(connectedAccount) {
+  return Boolean(localStorage.getItem(`autoRevealService:${connectedAccount}`))
+}
+
+export function saveAutoRevealPreference(connectedAccount, enabled) {
```
sohkai, line 189: Should we mark this to be per-dispute instead?

src/components/Disputes/DisputeAutoReveal.js
```diff
@@ -0,0 +1,119 @@
...
+import {
+  getAutoRevealPreference,
+  getCodeFromLocalStorage,
+  getOutcomeFromCommitment,
+} from '../../utils/crvoting-utils'
+
+function DisputeAutoReveal({ commitment, disputeId, onAutoReveal, roundId }) {
+  const { account } = useWallet()
+
+  const autoRevealRequested = useAutoRevealPolling(account, disputeId, roundId)
```
sohkai, line 15: Perhaps we should add a loading indicator and return null, to avoid a flash with the API.

src/components/Disputes/DisputeAutoReveal.js
```diff
@@ -0,0 +1,119 @@
...
+      disputeId={disputeId}
+      onAutoReveal={onAutoReveal}
+      roundId={roundId}
+    />
+  )
+}
+
+function RequestAutoReveal({ commitment, disputeId, onAutoReveal, roundId }) {
+  const { account } = useWallet()
+  const autoRevealServiceEnabled = getAutoRevealPreference(account)
```
sohkai, line 33: I think the variable naming here is a bit confusing.

What do you think about renaming it to `autoRevealPreviouslyRequested`? And adding a comment that the service must have failed if we're rendering this component and it was previously requested?

src/components/Disputes/DisputeAutoReveal.js
```diff
@@ -0,0 +1,119 @@
...
+      />
+      {autoRevealServiceEnabled && (
+        <Info
+          mode="warning"
+          css={`
+            margin-top: ${2 * GU}px;
+          `}
+        >
+          The auto-reveal service request has failed. Please click on the button
+          above to resend the request, reveal your vote and enable the service.
```
sohkai, line 63:
```suggestion
          Your previous request to enable the auto-reveal service for this vote failed.
```

src/components/Disputes/DisputeAutoReveal.js
```diff
@@ -0,0 +1,119 @@
...
+        </Info>
+      )}
+    </form>
+  )
+}
+
+function useAutoRevealPolling(account, disputeId, roundId) {
```
sohkai, line 70: Rather than polling, I wonder if we should only do this once on the component render?

We only use the polling to know if the user's request to re-register was successful, right?

src/hooks/useCourtContracts.js
fabriziovigevani: Yeah definitely

src/components/Disputes/DisputeAutoReveal.js
fabriziovigevani: True, we don't need a ref in this case.

src/utils/crvoting-utils.js
fabriziovigevani: We should :D

src/components/Disputes/DisputeAutoReveal.js
fabriziovigevani: Yes exactly. When implementing this, that was the first direction i took, but given that the request is asynchronous (we use the request queue), didn't find a way to force re-running the effect.

Do you have one in mind?

sohkai: Would it be possible to use `await` on the `onAutoReveal()` in `handleSubmit()` and update state afterwards? Outside of a separate tab, you have to go through the `handleSubmit()` handler to change your auto reveal state, right?

fabriziovigevani: I don't think it would be possible because we are using the same flow as we do for normal transactions. This means that the `onAutoReveal()` would return when the auto reveal request is added to the request queue for processing and we don't have any more the control of that request.

Nevertheless, note that we stop polling if the auto reveal was requested successfully so in that case would only be doing a single query. And in the case that the juror didn't request the auto reveal we would be polling every 3 seconds which i don't think it's that bad.  

Do you see a potential issue that could arise?

sohkai: > Do you see a potential issue that could arise?

Nope, but just feels a bit awkward to poll when we _know_ when to check. But it is a nice side effect to cover the case that you updated the API in a separate tab / etc. (although this is honestly unlikely here—it's much more common for authentication mechanisms to do this type of polling).

Let's leave it like this for now, and add a few comments explaining why we're polling with `useAutoRevealPolling()`.

src/components/Disputes/DisputeAutoReveal.js
```diff
@@ -0,0 +1,139 @@
...
+  useEffect(() => {
+    if (autoRevealRequested) {
+      return
+    }
+
+    let cancelled = false
+    let timeoutId
+
+    // Assumes jurorDraft exists
+    const fetchAutoReveal = timeout => {
```
fabriziovigevani, line 102: Realized that by using a local timeout as you suggested, we didn't really needed the `controlledTimer`

src/components/Disputes/DisputeAutoReveal.js
```diff
@@ -0,0 +1,139 @@
...
+          if (!autoRevealRequested) {
+            fetchAutoReveal(timer)
+          }
+        }
+      }, timeout)
+    }
+
+    setLoading(true)
+    // We start the timeout timer at 0 to immediately make the api call on render
+    fetchAutoReveal(0)
```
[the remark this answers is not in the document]
fabriziovigevani: Ah yes nice!

src/utils/crvoting-utils.js
```diff
@@ -182,10 +182,13 @@ export function removeCodeFromLocalStorage(connectedAccount, disputeId) {
 }
 
 // Auto reveal service preference
-export function getAutoRevealPreference(connectedAccount) {
-  return Boolean(localStorage.getItem(`autoRevealService:${connectedAccount}`))
+export function getAutoRevealPreference(connectedAccount, voteId) {
+  return localStorage.getItem(`autoRevealService:${connectedAccount}:${voteId}`)
```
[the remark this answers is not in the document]
fabriziovigevani: Ah yeah, the `Boolean` conversion wasn't accurate because we are storing the Boolean type which is converted to string, so if the auto reveal is `flase` gave a wrong result. Though i will move the updated evaluation inside this function again.

PR #371, opened 2020-07-03
Account: move more instances of 'Ethereum provider' to wallet terminology

sohkai:
Updates a few areas missed in #358.


rperez89:
> Updates a few areas missed in #358.

could we merge this one?

PR #373, opened 2020-07-08
Handle disputable metadata

### LINKED ISSUE #369: Disputes: Handle disputables metadata
See https://github.com/aragon/court-subgraph/pull/67/

- [x] For non disputables: Convert `metadata` to string (now comes as `Bytes`) 
```
#can use 

decodeURIComponent(
   metadata.replace('0x', '').replace(/[0-9a-f]{2}/g, '%$&')
)
```

- [x] Handle disputables metadata
- [x] Update `docs/metadata.md`

fabriziovigevani:
closes #369 

Needs https://github.com/aragon/court-subgraph/pull/67 to be deployed first.

Uses [Aragon connect](https://connect.aragon.org/) to describe the dispute metadata for disputes created by agreements (disputables):

- `disputedActionRadspec` (long description) => Built with `describeScript`
- `disputedActionText` (short description) => Built by using the `disputableAppName` and terminal action's role (e.g. `Tokens: Mint tokens`)
- `disputedActionURL` URL of the frontend where the disputed action is taking place.

Old disputes (eg Precedence campaign disputes) are still processed in the same way as before.

We still need to handle Conviction Voting which will be done after this is merged.

⚠️ Caveats

When fetching disputables metadata, it can take up to 5 seconds for the dispute to load (A cache for storing already described disputes was implemented)

**Previews**
####  Taken from PieDAO's Org
Screen Shot 2020-08-05 at 5 16 42 PM

#### Taken from address book used for app mining
Screen Shot 2020-08-05 at 5 15 18 PM

#### Where the terminal action is not the immediate step
Screen Shot 2020-07-08 at 12 52 03 PM
Screen Shot 2020-08-05 at 5 13 07 PM

#### When the dispute does not have an executable action
Screen Shot 2020-08-05 at 5 23 34 PM


sohkai:
> When fetching disputables metadata, it can take up to 5 seconds for the dispute to load (A cache for storing already described disputes was implemented)

We should probably come up with a loading state for this! Is this due to the IPFS look up or the radspec look up?

> Where the terminal action is not the immediate step

I think in this case, we would likely still want some explanation that it's going through a more complex process (e.g. your Agreement's rules may be different if a vote is involved than not).

How does it handle the terminal action doing multiple things (E.g. separate mint and burn actions?)

fabriziovigevani:
> We should probably come up with a loading state for this!

We do have a loading state for the dispute detail. Are you suggesting we have a different one for this?

Screen Shot 2020-07-08 at 3 45 00 PM

>  Is this due to the IPFS look up or the radspec look up?

It's due to the radspec lookup. 


> I think in this case, we would likely still want some explanation that it's going through a more complex process (e.g. your Agreement's rules may be different if a vote is involved than not).

Yeah this makes sense. We can probably do something similar as the new vote | delay descriptions?

> How does it handle the terminal action doing multiple things (E.g. separate mint and burn actions?)

It just prints them separated by a new line. Haven't tried with terminal actions which correspond to different methods but i assume it should handle them the same way.
Screen Shot 2020-07-08 at 12 49 54 PM

sohkai:
> It's due to the radspec lookup.

Interesting, this is surprisingly long even with Connect! Maybe we should have just a loading state for the description, since it's just this part that takes so long.

Let's use the round of testing or precedence campaign disputes to see what this is like on a live dispute and evaluate.

> We can probably do something similar as the new vote | delay descriptions?

Yes, we may have a couple things from the client we could bring over here as we complete them there (multi transaction stepper, transaction path viewer).

src/disputables/index.js
```diff
@@ -0,0 +1,261 @@
...
+        })
+        .join('\n')
+    : ''
+}
+
+/**
+ * Builds the URL corresponding to the disputed action. (e.g the URL of a vote)
+ * @param {String} organization Address of the organization
+ * @param {String} appAddress Address of the app
+ * @param {String} entity Entity where the action is being disputed (e.g vote, delay, etc)
```
sohkai, line 240: Maybe `entityPath` is a slightly more descriptive name :)

src/disputables/index.js
```diff
@@ -0,0 +1,261 @@
...
+  // (e.g grantPermission would become Grant permission)
+  const actionText = role?.name || describeFunctionSig(method.sig)
+
+  // If `name` is null most likely means the app is Kernel or ACL
+  const appName = name
+    ? name
+        .split('-')
+        .map(capitalize)
+        .join(' ')
+    : 'System app' // TODO: Update to more meaningfull name?
```
sohkai, line 216: Oh interesting, we should have the name available for these system apps!

If they're not available, please add an issue to Connect!

In this case, I think we might want to put something like `Unknown app (<app id>)` instead.

src/disputables/index.js
```diff
@@ -0,0 +1,261 @@
...
+
+    if (DISPUTABLE_ACTIONS.has(appId)) {
+      const { abi, entity, fn, scriptPosition } = DISPUTABLE_ACTIONS.get(appId)
+
+      // Get disputed action script
+      const disputableAppContract = getContract(disputableAddress, abi)
+      const result = await disputableAppContract[fn](disputableActionId)
+      const evmScript = result[scriptPosition]
+
+      // Get long and short description corrsponding to the disputed action.
```
sohkai, line 96:
```suggestion
      // Get long and short description corresponding to the disputed action.
```

src/disputables/index.js
```diff
@@ -0,0 +1,261 @@
...
+        ),
+      }
+
+      // Cache disputed action description
+      cachedDescriptions.set(disputeId, disputedActionDescription)
+
+      return disputedActionDescription
+    }
+  } catch (err) {
+    console.error('Error describing disputable action', err)
```
sohkai, line 121: Should we also return with a description that mentions it errorred?

It might also be useful to log this error to sentry with something like [this](https://github.com/aragon/aragon/blob/master/src/sentry.js#L11) so we can see if this happens a lot to people.

src/disputables/index.js
```diff
@@ -0,0 +1,261 @@
...
+  )
+
+  // Get disputed action long description
+  const disputedActionRadspec = terminalActions.length
+    ? buildDisputedActionRadspec(terminalActions)
+    : 'No description'
+
+  // Get disputed action short text
+  // Find method corresponding to the function of the disputed action.
+  const method = await findAppMethodFromIntent(app, terminalAction)
```
sohkai, line 162: 😅 I didn't know this existed on the connect export list (it doesn't appear to be documented).

I can roughly figure out what it's doing, but I'm not sure it's actually meant to be exported as an API from Connect! Did Gabi mention this was available?

src/disputables/index.js
```diff
@@ -0,0 +1,261 @@
...
+  const org = await connect(organization, 'thegraph', {
+    chainId: env('CHAIN_ID'),
+  })
+  const apps = await org.apps()
+
+  const transactionRequests =
+    (await describeScript(evmScript, apps, org.provider)) || []
+
+  // In order to get the terminal action/s we must search through the `transactionRequests` children (if any)
+  const terminalActions = getTerminalActions(transactionRequests)
```
sohkai, line 144: I wonder if we should re-organize the rest of the function a bit:

- Bail early if `terminalActions` is empty
- Push `disputedActionText` and `terminalAction` together, since the comment about the first action only applies to `disputedActionText`

src/components/Disputes/DisputeInfo.js
```diff
@@ -121,7 +121,13 @@ const DisputeInfo = React.memo(function({
               </Row>
               <Row compactMode={compactMode}>
                 {disputedActionRadspec ? (
-                  <Field label="Description" value={disputedActionRadspec} />
+                  <Field
+                    label="Description"
+                    value={disputedActionRadspec}
+                    css={`
+                      overflow-wrap: anywhere;
```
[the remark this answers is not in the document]
fabriziovigevani: `overflow-wrap: break-word` strangely doesn't break long words. 

This is a not so common scenario when the long word corresponds to an `appId`
Screen Shot 2020-07-23 at 12 06 58 AM

We could also add the `word-break: break-word` style since `overflow-wrap: anywhere` is [not supported by some browsers](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap#Browser_compatibility)

src/lib/web3-utils.js
```diff
@@ -139,7 +139,7 @@ export function isLocalOrUnknownNetwork(chainId = env('CHAIN_ID')) {
 }
 
 export function hexToAscii(hexx) {
```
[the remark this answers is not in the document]
fabriziovigevani: Ah yeah probably better to use `ehters` utils

src/web3-contracts.js
```diff
@@ -14,26 +14,23 @@ export function useContract(address, abi, signer = true) {
...
-    return new EthersContract(address, abi, ethProvider)
-  }, [abi, address, ethProvider])
+    return getContract(address, abi)
+  }, [abi, address])
+}
+
+export function getContract(
+  address,
+  abi,
+  provider = new Providers.JsonRpcProvider(defaultEthNode)
```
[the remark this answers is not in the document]
fabriziovigevani: Yes, that's a good idea.

src/utils/dispute-utils.js
```diff
@@ -43,9 +44,38 @@ export const transformRoundDataAttributes = round => {
...
+ *  Note that this function is meant to parse only dispute description and metadata uri (in case it exists). More relevant information will be processed elsewhere.
+ * @param {Object} dispute Dispute in question
+ * @returns {Array<String>} Array where the first item is the dispute description and second item is the metadata uri if it exists
+ */
+function parseMetadata(dispute) {
+  if (dispute.disputable) {
+    return [dispute.disputable.title]
+  }
+
+  const decodedMetadata = hexToAscii(dispute.metadata)
```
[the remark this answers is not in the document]
fabriziovigevani: Ah, we changed the type of the `metadata` attribute for the `dispute` entity in the subgraph (from `string` to `bytes`)  
See https://github.com/aragon/court-subgraph/pull/67/files#diff-e37ced9649dd2723e16a424f49db9193R74 

Though is only deploy to the staging subgraph. You can see all the descriptions messed up here https://court-staging.aragon.org/disputes 😄

src/disputables/index.js
fabriziovigevani: > Should we also return with a description that mentions it errorred?

Good question. Currently we are just returning `No description`  as description but maybe makes sense to tell that it actually errored yeah.

src/disputables/index.js
fabriziovigevani: Just checked the subgraphs again and i can see that they are available on the Aragon Mainnet subgraph but not on the Rinkeby subgraph. Just notified @0xGabi  about it.

src/disputables/index.js
fabriziovigevani: haha yes when i was working on this i mentioned @0xGabi what data we needed to have, and he mentioned about this function. Maybe he exposed it so we could use it in this case.

src/disputables/index.js
fabriziovigevani: Much better

src/components/Disputes/DisputeInfo.js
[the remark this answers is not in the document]
sohkai: Sounds good, let's add it as well and adjust based on how it looks with real usage!

src/disputables/index.js
sohkai: Probably not worth showing the full error message.

I think something along the lines of "Failed to fetch description" is fine as long as we're receiving alerts (so we can later debug it rather than ask the user specifically for what they saw).

fabriziovigevani:
@sohkai Added the missing bits:

- Loading state for disputed action description exclusively
   **Notes on this:** 
   Moved the rendering of the dispute data into a separate component `DisputeInfoContent`, you can see the commit [HERE](https://github.com/aragon/court-dashboard/pull/373/commits/522724b01fbb1e359946a0590d039da4967866bb) 
  
   Decided also to move the describing of the disputed action (previously on `useDispute` hook) to the component in charge to render it `DisputeInfoContent` using a `useDisputeAction` hook . The reason for this is that we already have a `fetching` indicator for when the dispute is processed in the `useDispute` hook, and adding another loading variable just for this didn't feel right. Relevant commit [HERE](https://github.com/aragon/court-dashboard/pull/373/commits/3c5e2062256ce4e8cd8431459c782c1f233d2bd8)

  Also, added some more  not so relevant files changed for deduping our Loading components that i should have done on a separate PR (sorry for that), here's the commit https://github.com/aragon/court-dashboard/pull/373/commits/e1a8f16cd3a826ee55d5d828e258cf8fd6b9f239

- Return `Failed to fetch description` when there's an error in the process, and log with Sentry.

- Describe all the steps when it's a multi step execution path

**Preview**

Loading
Screen Shot 2020-07-29 at 6 06 19 PM

Multi step

Screen Shot 2020-07-29 at 6 06 25 PM

src/disputables/index.js
```diff
@@ -226,17 +230,16 @@ function buildDisputedActionText(app, method) {
 
 /**
  * Builds the disputed action radspec description
+ * Assumes array of transactions not empty
```
sohkai, line 249: Just curious about the motivation for this change. Can we safely assume this, and are we looking to throw an error (instead of defaulting to an empty string) if it's not the case?

src/disputables/index.js
```diff
@@ -190,6 +192,27 @@ function getTerminalActions(transactions) {
     .flat()
 }
 
+/**
+ * Gets execution path specified by `transactions`
+ * @param {Array} transactions Array of transactions in the execution path
+ * @param {Array} apps Array of all the org's apps
+ * @returns {Array} Execution path
+ */
+function contexctualizeExecutionPath(transactions, apps) {
```
sohkai, line 198:
```suggestion
function contextualizeExecutionPath(transactions, apps) {
```

😄

src/disputables/index.js
```diff
@@ -190,6 +192,27 @@ function getTerminalActions(transactions) {
...
+  return transactions.map(transaction => {
+    const app = apps.find(app => addressesEqual(app.address, transaction.to))
+    const appName = app ? describeAppName(app.name) : 'Unknown app'
```
sohkai: To double check, we are only using the list of installed applications (from Connect) for this, right? If someone, for some reason, uninstalled an app later on, it would appear as "Unknown app"?

I think this is fine, but I'll keep a mental note about it and also ask the Connect team about making sure we can detect uninstalled apps.

src/disputables/index.js
fabriziovigevani: >  Can we safely assume this

Yes we can assume this since we are now early returning when the transactions array is empty 
See [here](https://github.com/aragon/court-dashboard/pull/373/files/e1a8f16cd3a826ee55d5d828e258cf8fd6b9f239..f9d01ba38e89367da5f80b9cac81c80a74693849#diff-07f7c052c786459c4930bc8491c43713R152)

> and are we looking to throw an error

Ah yeah forgot to return an error description here as well instead of the  `No description`.

src/disputables/index.js
fabriziovigevani: > To double check, we are only using the list of installed applications (from Connect) for this, right? If someone, for some reason, uninstalled an app later on, it would appear as "Unknown app"?

Yes that's correct.

> I think this is fine, but I'll keep a mental note about it and also ask the Connect team about making sure we can detect uninstalled apps.

Cool, we can add a comment and ask connect team yes :)

src/components/Disputes/DisputeInfoContent.js
```diff
@@ -296,7 +298,10 @@ const DisputedActionNA = () => {
       >
         N/A
       </span>
-      <Help hint="">The disputed Action</Help>
+      <Help hint="">
+        This dispute is simply between the two arguments. There's no executable
+        action involved.
+      </Help>
```
fabriziovigevani, line 304: Not sure if this is enough information for the user or if it's formal enough. Suggestions are welcome

src/disputables/index.js
```diff
@@ -25,10 +25,18 @@ const DELAY_APP_ID =
   '0x133c97c74d2197a068988549d31108ce57af8f0ccf90ff9edf0ba5d349f2a450' // disputable-delay.aragonpm.eth
 const VOTING_APP_ID =
   '0x09cdc3e6887a0002b11992e954a40326a511a1750a2f5c69d17b8b660b0d337a' // disputable-voting.aragonpm.eth
-// '0x705b5084c67966bb8e4640b28bab7a1e51e03d209d84e3a04d2a4f7415f93b34' // disputable-voting.open.aragonpm.eth
+const VOTING_OPEN_APP_ID =
+  '0x705b5084c67966bb8e4640b28bab7a1e51e03d209d84e3a04d2a4f7415f93b34' // disputable-voting.open.aragonpm.eth
```
fabriziovigevani, line 29: Added support for the `open` version as well of disputable voting since that's where Facu was testing.

src/hooks/useEvidences.js
```diff
@@ -89,5 +89,16 @@ export default function useEvidences(rawEvidences) {
...
+  // First evidence submitted by defendant is treated as the dispute description
+  const processedEvidences = useMemo(() => {
+    if (dispute.disputable) {
+      return evidences.filter(
+        e => e.metadata !== dispute.disputable.actionContext
+      )
+    }
+
+    return evidences
+  }, [dispute.disputable, evidences])
```
fabriziovigevani, line 101: From the convo offline, removed the defendant's `Justification` as argument.

src/utils/dispute-utils.js
```diff
@@ -44,6 +44,20 @@ export function transformRoundDataAttributes(round) {
...
+function transformDisputableDataAttributes(dispute) {
+  if (!dispute.disputable) {
+    return null
+  }
+
+  const { disputable } = dispute
+
+  return {
+    ...disputable,
+    actionContext: toUtf8String(disputable.actionContext || '0x'),
```
sohkai, line 56: Right, we weren't able to decode the string from the subgraph? Was wondering if we should also use the `true` flag like we do in `parseMetadata()`?

src/components/Disputes/DisputeInfoContent.js
```diff
@@ -166,37 +161,151 @@ function Field({ label, loading, value, ...props }) {
...
-  const lastRound = rounds?.[lastRoundId]
-  const voteWinningOutcome = lastRound?.vote?.winningOutcome
-  const appealedRuling = lastRound?.appeal?.appealedRuling
-
-  if (!isFinalRulingEnsured) {
-    return (
-      <Link external href={disputedActionURL}>
-        {disputedActionText}
+  const ActionTextComponent = useMemo(() => {
+    // Disputes could not have an executable action
```
sohkai, line 173:
```suggestion
    // Disputes may not include an embedded executable action
```

src/components/Disputes/DisputeInfoContent.js
sohkai: Could you leave a screenshot for this case? I was looking at the deployed version to Staging, but couldn't find a dispute like this (should dispute 26 show this?).

fabriziovigevani: Hmm, it should show properly
 
Screen Shot 2020-08-07 at 3 08 51 PM

src/components/Disputes/DisputeInfoContent.js
```diff
@@ -296,7 +298,10 @@ const DisputedActionNA = () => {
       >
         N/A
       </span>
-      <Help hint="">The disputed Action</Help>
+      <Help hint="">
+        This dispute is simply between the two arguments. There's no executable
+        action involved.
```
sohkai, line 303:
```suggestion
        This dispute does not include a binding action and is simply between the given context and arguments.
```

src/utils/dispute-utils.js
fabriziovigevani: @facuspagnuolo just updated the subgraph where it's decoded from the subgraph, so just updated the PR.

See https://github.com/aragon/court-subgraph/pull/75

fabriziovigevani:
@sohkai there's some new changes for review.

Since now we cannot get the disputable voting `evmScript` via contract call, decided to implement a function for each disputable app which handles how it should fetch the evm script. This `extractor` function is added to each disputable app in  the `DISPUTABLE_ACTIONS` mapping. 

For disputable voting we are using `@aragon/connect-thegraph` library and performing the direct query as opposed to using the disputable voting connector. 

Also moved the disputable actions mapping to its own file to tidy things up a bit.
 
The main changes are: 

- Implementing a `scriptExtractor` function  by disputable app which is added to the `DISPUTABLE_ACTIONS` mapping.
- Upgraded  `@aragon/connect` to `latest` as i was experiencing the same issue we had when working on the `honey-pot` with the older version.
- In consequence of the previous point, had to upgrade ethers to v5 as well.

src/lib/web3-utils.js
```diff
@@ -1,17 +1,18 @@
 import env from '../environment'
-import { providers as Providers } from 'ethers'
-import { id as keccak256 } from 'ethers/utils'
+import { providers as Providers, utils } from 'ethers'
 import { InvalidURI, InvalidNetworkType, NoConnection } from '../errors'
 import { validHttpFormat } from './uri-utils'
 
+const { id: hash256, solidityKeccak256: soliditySha3, toUtf8String } = utils
```
sohkai, line 6: Oh interesting, what is `hash256` if it's not `keccak256`? (Function signatures should be using keccak256)

src/disputables/queries.js
```diff
@@ -0,0 +1,17 @@
+import gql from 'graphql-tag'
+import { GraphQLWrapper } from '@aragon/connect-thegraph'
```
sohkai, line 2: I'm out of the loop on this one—is this better than the previous graphql transport we were using against the court subgraph?

src/disputables/scriptExtractors.js
```diff
@@ -0,0 +1,65 @@
...
+    disputableActionId,
+    'getVote',
+    10
+  )
+}
+
+export async function votingExtractor(disputableAddress, disputableActionId) {
+  // Disputable voting now saves the hash of the evmScript so we need to get it from the subgraph.
+  const networkType = sanitizeNetworkType(getNetworkType())
+  const subgraphUrl = `https://api.thegraph.com/subgraphs/name/aragon/aragon-dvoting-${networkType}${
```
sohkai, line 37: Ah, could we move this to be a dynamic utility function based on the app id? We'll have a different subgraph for the precedence campaign apps.

src/disputables/scriptExtractors.js
```diff
@@ -0,0 +1,65 @@
...
+    SUBGRAPH_NAME ? `-${SUBGRAPH_NAME}` : ''
+  }`
+
+  const { data } = await performQuery(subgraphUrl, disputableVotingQuery, {
```
sohkai, line 41: I wonder if it would be a better API to hide `performQuery()` and only expose `performDisputableVotingQuery(id, voteId)`. Otherwise it might be a bit hard to know which params to use?

sohkai (approved):
I think we should also be able to bump the bundle-checker down now, since we should be deduping ethers versions!

I otherwise have a couple of comments on things I know we'll want to generalize further down the line. Otherwise, looking good to me!

src/lib/web3-utils.js
fabriziovigevani: Ah you are right https://docs.ethers.io/v5/api/utils/hashing/#utils-id

Bad naming 😄

src/disputables/scriptExtractors.js
fabriziovigevani: Yes, good one!

src/disputables/queries.js
fabriziovigevani: We could actually just do what the library does (creating the Client and performing the query) ourselves.  
It's using the same library internally.

src/disputables/index.js
```diff
@@ -229,27 +234,38 @@ function buildDisputedActionRadspec(transactions) {
...
+  arbitrableAddress
 ) {
-  const orgUrl = buildOrgUrl(organization)
-  return [orgUrl, appAddress, entityPath, actionId].join('/')
+  const networkType = sanitizeNetworkType(getNetworkType())
+  const arbitrable = getKnownArbitrable(networkType, arbitrableAddress)
+
+  let url
+  if (arbitrable) {
+    url = [arbitrable.link, arbitrable.entityPath, actionId].join('/')
```
sohkai, line 252: Not a big deal, but maybe this is something we would hide inside the API since each arbitrable could have its own URL scheme?

Perhaps something like `isArbitrableKnown(arbitrableAddress, networkType) && buildArbitrableUrl(arbitrableAddress, actionId, networkType)` would be better? And then the mapping would be from `address => function`.

src/disputables/queries.js
```diff
@@ -1,9 +1,21 @@
 import gql from 'graphql-tag'
-import { GraphQLWrapper } from '@aragon/connect-thegraph'
+import { Client } from 'urql/'
```
sohkai, line 2:
```suggestion
import { Client } from 'urql'
```

Not sure, but did we need the `/`?

fabriziovigevani: Oops, no we don't

PR #375, opened 2020-07-28
AccountModule: Structure & Internals refresh

Evalir:
Builds on #358. Keeps refreshing the court's Account Module, so it's up to date with the Client's implementation on the relevant internals.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,5 @@
+{
+  "semi": false,
+  "singleQuote": true,
+  "trailingComma": "es5"
+}
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,5 +1,0 @@
-{
-  "semi": false,
-  "singleQuote": true,
-  "trailingComma": "es5"
-}
--- a/src/components/Account/AccountModule.js
+++ b/src/components/Account/AccountModule.js
@@ -56,8 +56,7 @@ function AccountModule() {
   const previousScreenIndex = useRef(-1)
 
   const { screenIndex, direction } = useMemo(() => {
-    const screenId = (() =>
-      status === 'disconnected' ? 'providers' : status)()
+    const screenId = status === 'disconnected' ? 'providers' : status
 
     const screenIndex = SCREENS.findIndex(screen => screen.id === screenId)
     const direction = previousScreenIndex.current > screenIndex ? -1 : 1
--- a/src/components/Account/AccountModule.js
+++ b/src/components/Account/AccountModule.js
@@ -45,7 +45,7 @@ function AccountModule() {
     return () => clearTimeout(timer)
   }, [connector, status])
 
-  const handleCancelConnection = useCallback(() => {
+  const handleResetConnection = useCallback(() => {
     wallet.reset()
   }, [wallet])
 
@@ -131,7 +131,7 @@ function AccountModule() {
             return (
               <ScreenConnecting
                 providerId={activating}
-                onCancel={handleCancelConnection}
+                onCancel={handleResetConnection}
               />
             )
           }
@@ -139,7 +139,12 @@ function AccountModule() {
             return <ScreenConnected wallet={wallet} />
           }
           if (screenId === 'error') {
-            return <ScreenError error={activationError} onBack={wallet.reset} />
+            return (
+              <ScreenError
+                error={activationError}
+                onBack={handleResetConnection}
+              />
+            )
           }
           return <ScreenProviders onActivate={handleActivate} />
         }}
--- a/src/components/Account/AccountModule.js
+++ b/src/components/Account/AccountModule.js
@@ -1,8 +1,7 @@
 import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react'
 import { useWallet } from 'use-wallet'
-import { Button, GU, IconConnect, springs, useViewport } from '@aragon/ui'
-import { Transition, animated } from 'react-spring/renderprops'
-import { shortenAddress, getUseWalletProviders } from '../../lib/web3-utils'
+import { Button, GU, IconConnect, useViewport } from '@aragon/ui'
+import { shortenAddress } from '../../lib/web3-utils'
 import AccountButton from './AccountButton'
 import HeaderPopover from '../Header/HeaderPopover'
 import ScreenConnected from './ScreenConnected'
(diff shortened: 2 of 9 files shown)


rperez89:
> Builds on #358. Keeps refreshing the court's Account Module, so it's up to date with the Client's implementation on the relevant internals.

Do you have some screenshots of the new look and feel of the wallet?

Evalir:
> Do you have some screenshots of the new look and feel of the wallet?

It's purely an internal upgrade, so it should feel exactly the same :)

src/components/Account/AccountModule.js
```diff
@@ -167,78 +124,49 @@ function AccountModule() {
...
+        onOpen={open}
         opener={buttonRef.current}
+        screenId={screenId}
+        screenData={{
+          account,
+          // This is needed because use-wallet throws an error when the
+          // activation fails before React updates the state of `activating`.
+          // A future version of use-wallet might return an
+          // `activationError` object instead, making this unnecessary.
+          activating: screen.id === 'error' ? null : activatingDelayed,
```
sohkai, line 141: Do we still need this now with use-wallet 0.8?

src/components/ClockModule.js
```diff
@@ -147,137 +150,145 @@ function ClockModule() {
       />
 
       <HeaderPopover
-        animateHeight={false}
         heading="Court clock"
-        height={(isSynced ? 27.5 : 26) * GU}
-        width={45 * GU}
```
sohkai, line 153: Wasn't sure if we wanted to keep these original dimensions, but the clock popover still looks good with the `30 x 51` GU dimensions.

src/components/ClockModule.js
```diff
@@ -147,137 +150,145 @@ function ClockModule() {
...
+        direction={DEFAULT_DIRECTION}
         onClose={handlePopoverClose}
         opener={buttonRef.current}
+        screenData={{
+          courtConfig,
+        }}
+        screenId={DEFAULT_SCREEN_ID}
+        screenKey={({ courtConfig }) => courtConfig?.id}
```
sohkai, line 161: We might be able to find a better key for this, but I'm also wondering if this really needs to use `HeaderPopover` since it doesn't transition between screens.

Maybe we should keep these two popovers (`ClockModulePopover` and `AccountModulePopover`) separate?

src/components/Account/AccountModule.js
Evalir: Nope, this won't be needed, I'm pretty sure. :) Let's update to 0.8!

src/components/Account/AccountModule.js
```diff
@@ -1,145 +1,81 @@
...
     }
 
-    if (activating) {
-      setActivatingDelayed(activating)
-      return
+    if (status === 'connecting') {
+      setActivatingDelayed(connector)
+      timer = setTimeout(() => {
+        setActivatingDelayed(null)
+      }, 400)
```
Evalir, line 42: This is something noteworthy: with the removal of the [`activating` escape hatch on error](https://github.com/aragon/court-dashboard/blob/master/src/components/Account/AccountModule.js#L187-L191), the previous implementation of the simulated delays was causing duplicated animations on connection and on disconnection. To avoid this, I had to move the timeout only on the case where we want to actually set it to null for the screen to change, which is the `connecting` case.

src/components/Account/AccountModule.js
Evalir: Done in https://github.com/aragon/court-dashboard/pull/375/commits/8bd72dc57eb044f5b7f3f20e7749fc95ab76749e

src/components/ClockModule.js
Evalir: I re-introduced the old Popover again; should we update to the new `30 x 51` GU dimensions anyway?

sohkai: No, let's keep it as it was!

src/components/Account/AccountModule.js
```diff
@@ -1,145 +1,81 @@
...
 
   const { screenIndex, direction } = useMemo(() => {
-    const screenId = (() => {
-      if (activationError) return 'error'
-      if (activatingDelayed) return 'connecting'
-      if (account) return 'connected'
-      return 'providers'
-    })()
+    const screenId = (() =>
+      status === 'disconnected' ? 'providers' : status)()
```
fabriziovigevani, line 60: Should we remove the wrapped function now?

.prettierrc
```diff
@@ -0,0 +1,5 @@
+{
+  "semi": false,
+  "singleQuote": true,
+  "trailingComma": "es5"
+}
```
fabriziovigevani, line 5: Did the prettier rules on the `eslint` config not work for you?

src/components/Account/AccountModule.js
```diff
@@ -1,145 +1,81 @@
...
-      setActivatingDelayed(null)
-    }, 500)
+    return () => clearTimeout(timer)
+  }, [connector, status])
 
-    return () => {
-      clearTimeout(timer)
-    }
-  }, [activating, activationError])
+  const handleCancelConnection = useCallback(() => {
```
fabriziovigevani, line 48: Wonder if makes sense to rename it to `handleResetConnection` so we use it both in `ScreenConnecting` `onCancel` prop and `ScreenError` `onBack` prop

src/components/Account/AccountModule.js
```diff
@@ -1,145 +1,81 @@
...
+  { id: 'connected', title: 'Active account' },
+  { id: 'error', title: 'Connection error' },
 ]
 
 function AccountModule() {
-  const buttonRef = useRef()
-  const wallet = useWallet()
   const [opened, setOpened] = useState(false)
-  const [animate, setAnimate] = useState(false)
   const [activatingDelayed, setActivatingDelayed] = useState(false)
```
fabriziovigevani, line 21:
```suggestion
  const [activatingDelayed, setActivatingDelayed] = useState(null)
```

src/components/Account/AccountModule.js
Evalir: Let's do it!

.prettierrc
Evalir: Yeah, had to make some changes to my editor in between. They're working now. :)

src/components/Account/AccountModule.js
Evalir: Ah definitely!

src/components/Account/AccountModule.js
Evalir: Done in https://github.com/aragon/court-dashboard/pull/375/commits/3c757b38a0b01dd3fdfebffd763ad3170faff9f2

PR #384, opened 2020-09-01
Disputable subgraph urls: Provide full URL by network

fabriziovigevani:


src/disputables/mappings.js
```diff
@@ -35,7 +35,18 @@ export const DISPUTABLE_ACTIONS = new Map([
...
-export const DISPUTABLE_SUBGRAPH_NAMES = new Map([
-  ...VOTING_APP_IDS.map(appId => [appId, 'aragon-dvoting']),
+// corresponding subgraph urls by network
+export const DISPUTABLE_SUBGRAPH_URLS = new Map([
+  ...VOTING_APP_IDS.map(appId => [
+    appId,
+    {
+      // TODO: Update to Aragon's own node
+      mainnet:
+        'https://api.thegraph.com/subgraphs/name/aragon/aragon-dvoting-mainnet-staging',
```
fabriziovigevani, line 45: Will update when we get green light.

sohkai: Let's update it to our own hosted version now—the network dashboard is already relying on it.

sohkai (approved):
Looks good! In the chat I was also thinking about doing this for the main court-subgraph URLs (since our hosted version has now synced).

Let's merge this first though, and do that separately!

PR #386, opened 2020-09-01
Configure court subgraph urls by network

fabriziovigevani:
- Moves subgraph urls entirely separated by network as opposed to dynamically constructed.
- Disputable subgraphs: Uses new Aragon's subgraph node.
- Overrides Aragon connect subgraph when connecting to an org.
- Flags new disputes for precedence campaign.


src/disputables/mappings.js
```diff
@@ -40,13 +40,12 @@ export const DISPUTABLE_SUBGRAPH_URLS = new Map([
   ...VOTING_APP_IDS.map(appId => [
     appId,
     {
-      // TODO: Update to Aragon's own node
       main:
-        'https://api.thegraph.com/subgraphs/name/aragon/aragon-dvoting-mainnet-staging',
+        'https://graph.backend.aragon.org/subgraphs/name/aragon/aragon-dvoting-mainnet-staging', // TODO: Update to main subgraph when available
       rinkeby:
```
sohkai, line 45: Ah, let's keep `api.thegraph.com` for rinkeby/ropsten for now!

src/networks.js
```diff
@@ -21,24 +15,39 @@ export const networkConfigs = {
     network_reserve: '0xec0dd1579551964703246becfbf199c27cb84485',
     nodes: {
       defaultEth: 'https://mainnet.eth.aragon.network/',
+      subgraph: {
+        http:
+          'https://graph.backend.aragon.org/subgraphs/name/aragon/aragon-court',
+        ws: 'wss://graph.backend.aragon.org/subgraphs/name/aragon/aragon-court',
```
sohkai, line 21: Just wondering, I think we've removed the websocket endpoints because they're not working reliably with subgraphs at the moment (#381). Are we able to remove these as well?

src/networks.js
```diff
@@ -55,18 +64,22 @@ export const networkAgentAddress = getNetworkConfig().network_agent
...
-export const defaultEthNode =
-  getDefaultEthNode() || getNetworkConfig().nodes.defaultEth
+function getRinkebySubgraphUrls(subgraphName) {
+  const subgraphBase = `api.thegraph.com`
+  const subgraphPath = '/subgraphs/name/aragon/aragon-court'
 
-export const defaultIpfsGateway = getIpfsGateway()
+  const [httpEndpoint, wsEndpoint] = ['https', 'wss'].map(
+    protocol =>
+      `${protocol}://${subgraphBase}${subgraphPath}-${subgraphName ||
```
sohkai, line 80: Not a big deal, but should we just inline `subgraphBase` and `subgraphPath`?

src/disputables/utils.js
```diff
@@ -1,6 +1,11 @@
-import { getNetworkType } from '../lib/web3-utils'
```
sohkai, line 1: Wondering if we should rename this file differently, since it's just related to getting subgraph urls. Perhaps `connect-endpoints`?

src/endpoints.js
```diff
@@ -23,31 +30,16 @@ export function courtServerEndpoint() {
...
-    networkType === 'main'
-      ? GRAPH_API_PATH
-      : `${GRAPH_API_PATH}-${SUBGRAPH_NAME || networkType}`
+export const defaultIpfsGateway = getIpfsGateway()
 
-  return [`${API_BASE_HTTP}${API_PATH}`, `${API_BASE_WS}${API_PATH}`]
-}
+export const defaultSubgraphHttpEndpoint = getSubgraphHttpEndpoint()
+
+export const defaultSubgraphWsEndpoint = getSubgraphWsEndpoint()
```
sohkai, line 45: Just sanity checking if we still need this :)

src/endpoints.js
```diff
@@ -1,10 +1,17 @@
...
+  getDefaultEthNode,
+  getIpfsGateway,
+  getSubgraphHttpEndpoint,
+  getSubgraphWsEndpoint,
+} from './local-settings'
 
 const CHAIN_ID = environment('CHAIN_ID')
 const COURT_SERVER_NAME = environment('COURT_SERVER_NAME')
-const SUBGRAPH_NAME = environment('SUBGRAPH_NAME')
+// const SUBGRAPH_NAME = environment('SUBGRAPH_NAME')
```
sohkai, line 14: Should we remove this? The logic is now in `networks.js`?

sohkai (approved):
🏎💨

Couple of nitpicks :).

src/networks.js
rperez89: are the normal subscriptions (without the polling query) working with your infra?

fabriziovigevani: Oh yeah true, i think we can remove them yes.

> are the normal subscriptions (without the polling query) working with your infra?

Maybe they are, but we currently we are using their node for mainnet and for testnets we are using the subgraph's original one.

src/endpoints.js
fabriziovigevani: yep definitely

src/endpoints.js
fabriziovigevani: Removed in https://github.com/aragon/court-dashboard/pull/391

src/networks.js
fabriziovigevani: Done in https://github.com/aragon/court-dashboard/pull/391

PR #387, opened 2020-09-01
Routing: update links for HashRouter

fabriziovigevani:
From https://reactrouter.com/web/api/HashRouter

It appears that for anchors we need to add the `#` in front of the path.


sohkai (approved):
Ouch, thanks for finding these @fabriziovigevani!

PR #391, opened 2020-09-01
Remove subscriptions

fabriziovigevani:
Subscriptions are not working properly at the moment so we went for a polling instead.

See https://github.com/aragon/court-dashboard/pull/381


src/networks.js
```diff
@@ -33,21 +30,15 @@ export const networkConfigs = {
     court: '0x3b26bc496aebaed5b3E0E81cDE6B582CDe71396e',
     nodes: {
       defaultEth: 'https://ropsten.eth.aragon.network/',
-      subgraph: {
-        http:
-          'https://api.thegraph.com/subgraphs/name/aragon/aragon-court-ropsten',
-        ws: 'wss://api.thegraph.com/subgraphs/name/aragon/aragon-court-ropsten',
-      },
+      subgraph:
+        'https://api.thegraph.com/subgraphs/name/aragon/aragon-court-ropsten',
```
rperez89, line 34: are we using the graph endpoint here because the aragon ones are still syncing or we have some other issue?

fabriziovigevani: Yes that's the reason.

sohkai (approved):
Let's rebase and merge!

src/networks.js
sohkai: Yes, to explain this a bit more:

1. Websockets are expensive compared to http requests, and polling is not so bad at the moment when block times are 13-15s; even if they do work, we may not want to maintain the infrastructure for them to be used reliable at peak load
2. We are experimenting a bit and haven't synced Rinkeby subgraphs yet.

rperez89: awesome thxs for the answers :)

PR #393, opened 2020-09-02
Dispute: polish URI resolution

sohkai:
Just a few quick things I noticed :).

### DIFF
--- a/src/components/Disputes/DisputeInfoContent.js
+++ b/src/components/Disputes/DisputeInfoContent.js
@@ -8,11 +8,11 @@ import IdentityBadge from '../IdentityBadge'
 import Loading from '../Loading'
 import { useWallet } from '../../providers/Wallet'
 
+import { describeDisputedAction } from '../../disputables'
 import { IPFS_ENDPOINT } from '../../endpoints'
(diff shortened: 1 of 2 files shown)


src/components/Disputes/DisputeInfoContent.js
```diff
@@ -109,10 +109,7 @@ function Field({ label, loading, value, ...props }) {
         if (typeof value === 'string') {
           const ipfsPath = getIpfsCidFromUri(value)
           if (ipfsPath) {
-            const ipfsUrl = resolvePathname(
-              ipfsPath,
-              `${IPFS_ENDPOINT}/${ipfsPath}`
-            )
+            const ipfsUrl = resolvePathname(ipfsPath, `${IPFS_ENDPOINT}/`)
```
sohkai, line 112: Note that this is functionality the same as above (and the above is a bit confusing—why are we appending the same path onto itself?).

rperez89: ahh, yes good catch, seems that the library somehow internally was resolving it right anyways? since the second argument is the current pathname and the first one is the one that you would like to resolve

sohkai: Yes, since resolving `foo` into `bar/baz` results in `bar/foo`.

PR #394, opened 2020-09-02
Evidences: Prevent accordion from auto-collapsing.

fabriziovigevani:
Since we are now using a polling mechanism, new objects come every 13 seconds. 

This PR prevents `DisputeEvidences` component from re-rendering if there are no new evidences

### DIFF
--- a/src/components/Disputes/DisputeEvidences.js
+++ b/src/components/Disputes/DisputeEvidences.js
@@ -168,6 +168,7 @@ function getSubmitterLabel(submitter, defendant, plaintiff) {
 }
 
 export default function Evidences({ dispute, evidences }) {
+  // This hook ensures us that evidenceProcessed won't be updated unless there are new evidences.
   const [evidenceProcessed, fetchingEvidences] = useEvidences(
     dispute,
     evidences
--- a/src/components/Disputes/DisputeEvidences.js
+++ b/src/components/Disputes/DisputeEvidences.js
@@ -11,19 +11,16 @@ import { dateFormat } from '../../utils/date-utils'
 import folderIcon from '../../assets/folderIcon.svg'
 
 const DisputeEvidences = React.memo(function DisputeEvidences({
-  dispute,
+  defendant,
   evidences,
+  loading,
+  plaintiff,
 }) {
-  const [evidenceProcessed, fetchingEvidences] = useEvidences(
-    dispute,
-    evidences
-  )
-
   return (
     <React.Fragment>
-      <SyncIndicator visible={fetchingEvidences} label="Loading evidences…" />
-      {evidenceProcessed &&
-        evidenceProcessed.map((evidence, index) => {
+      <SyncIndicator visible={loading} label="Loading evidences…" />
+      {evidences &&
+        evidences.map((evidence, index) => {
           const { createdAt, submitter, metadata, error } = evidence
           return (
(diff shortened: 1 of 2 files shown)


src/components/Disputes/DisputeEvidences.js
```diff
@@ -154,16 +155,30 @@ const EvidenceContent = React.memo(function EvidenceContent({
...
-  if (addressesEqual(submitter, dispute.plaintiff)) {
+  if (addressesEqual(submitter, plaintiff)) {
     return 'Plaintiff'
   }
 
   return ''
 }
 
-export default DisputeEvidences
+export default ({ dispute, evidences }) => {
```
sohkai, line 170: Do we need to separate the `useEvidences()` out?

(Otherwise, let's use the function form and add a name to this component).

sohkai (commented):
It looks good... but I'm not sure where the caching happens to avoid re-rendering (and the auto-collapse). I was expecting a `useMemo()` with a unique string identifier built from the evidences / etc. but the change is quite different haha.

src/components/Disputes/DisputeEvidences.js
fabriziovigevani: Yes we do since `dispute` and `evidences` are always a new object (due to polling) ? 

What we do here is to ensure we are always providing the same data to `DisputeEvidences` if it has not changed.

fabriziovigevani:
> but I'm not sure where the caching happens to avoid re-rendering

The caching happens on the `useEvidences` hook. We had already the logic for dedupping evidences (issue we had on the court lunch) which already ensures us evidences are not updated if no new evidences are added. 

The logic inside the useEffect kind of behaves as a `useMemo` since we are not updating the evidences unless there's a new one.

We also removed the `useMemo` used to filter evidences which removes the defendant evidence (since it's now part of the dispute description), because the `disputable` entity was a dependency of it. Now that we changed to a polling it would recompute the value on every polling. Instead we filter out that evidence on the `useEffect`.

src/components/Disputes/DisputeEvidences.js
```diff
@@ -154,16 +155,30 @@ const EvidenceContent = React.memo(function EvidenceContent({
...
+  const [evidenceProcessed, fetchingEvidences] = useEvidences(
```
fabriziovigevani, line 171: The hook ensures us that `evidenceProcessed` won't be updated unless there are new evidences.

sohkai (approved):
Thanks for the explanation! Maybe worth leaving a comment on `DisputeEvidences#l171` :).

> We also removed the useMemo used to filter evidences which removes the defendant evidence (since it's now part of the dispute description), because the disputable entity was a dependency of it. Now that we changed to a polling it would recompute the value on every polling. Instead we filter out that evidence on the useEffect.

About this, what would happen if the defendant had more than one evidence attached? We should move the first one they submitted to be the description, but then show all the rest as evidence.

fabriziovigevani:
>  Maybe worth leaving a comment on DisputeEvidences#l171

Sure, i'll keep in mind to add more comments in the future :) 

> About this, what would happen if the defendant had more than one evidence attached? We should move the first one they submitted to be the description, but then show all the rest as evidence.

Ah yeah, the rest of the evidences would indeed be displayed as normal evidences in the scenario you described.

PR #395, opened 2020-09-03
Update token price api 

fabriziovigevani:
For tokens other than `ANJ` it uses [`0x`](https://0x.org/docs/api#examples-2) api

For ANJ price we use Uniswap's subgraphs as it's done [here](https://github.com/aragon/conviction-funding-pilot/commit/2641df4c07f56d45db16dd326186c9318d2d960c)
The reason being `0x` does not provide any price data for `ANJ`


src/hooks/useTokenAmountToUsd.js
```diff
@@ -0,0 +1,92 @@
...
+    return () => {
+      cancelled = true
+    }
+  }, [amount, decimals, symbol])
+
+  return amountInUsd
+}
+
+function convertAmount(amount, price, decimals) {
+  const rate = price * PRECISION
```
Evalir, line 87: Ah interesting; if `PRECISION` is a `BigNumber`, how are we using normal operators? AFAIK we're not using native `BigInts` right?

src/hooks/useTokenAmountToUsd.js
```diff
@@ -0,0 +1,92 @@
...
+
+  return formatUnits(amount.mul(rate.toString()).div(PRECISION), {
```
Evalir, line 89: Just a note, but in the future we might want to migrate to using `TokenAmount.format` :)

Evalir (approved):
This looks good! Really nice that the 0x API returns bignumbers. :)

src/hooks/useUniswapAnjPrice.js
```diff
@@ -0,0 +1,63 @@
...
+  const [anjPrice, setAnjPrice] = useState(0)
+  const client = new Client({ url: UNISWAP_URL })
+
+  useEffect(() => {
+    let cancelled = false
+    let retryTimer
+    async function fetchPrice() {
+      try {
+        const anjResults = await client.query(ANJ_PRICE_QUERY).toPromise()
+        const ethResults = await client.query(ETH_PRICE_QUERY).toPromise()
```
Evalir, line 38: Ah, also, if we want to make this more performant, we can just use `Promise.all`

src/hooks/useUniswapAnjPrice.js
```diff
@@ -0,0 +1,63 @@
...
+  query {
+    pair(id: "${DAI_ETH_PAIR}") {
+      token0Price
+    }
+  }
+`
+
+export function useUniswapAnjPrice() {
+  const [anjPrice, setAnjPrice] = useState(0)
+  const client = new Client({ url: UNISWAP_URL })
```
sohkai, line 30: I'm just wondering, would this be expensive to continue re-creating if multiple components use the same graphQL client?

We could use a weakmap cache based on URL to avoid re-creating it.

src/hooks/useTokenAmountToUsd.js
```diff
@@ -0,0 +1,92 @@
+import { useEffect, useState } from 'react'
+import { captureException } from '@sentry/browser'
+
+import { useCourtConfig } from '../providers/CourtConfig'
+import { useUniswapAnjPrice } from './useUniswapAnjPrice'
+import { bigNum, formatUnits } from '../lib/math-utils'
+import { getNetworkType } from '../lib/web3-utils'
+
+const API_BASE = 'https://api.0x.org'
+const PRECISION = bigNum(10).pow(18)
```
sohkai, line 10: Just checking, since USDC only has 6 decimals, is the API precision always 18 units, or is it dependent on the token to sell?

src/hooks/useTokenAmountToUsd.js
fabriziovigevani: Created an issue for it :) https://github.com/aragon/court-dashboard/issues/396

src/hooks/useTokenAmountToUsd.js
fabriziovigevani: Ah when using a BigNumber on a native operation it's converted to its primitive value via `valueOf`.

src/hooks/useTokenAmountToUsd.js
fabriziovigevani: Ah yeah, they actually come up in `sellToken` denomination. Good catch

src/hooks/useUniswapAnjPrice.js
fabriziovigevani: Yeah true, we should probably cache it. 

> We could use a weakmap cache based on URL to avoid re-creating it. 

What advantage you see over just using a constant with the client initialized? Also doesn't weakmaps accept objects as key?

sohkai: Ah yes, a constant is probably just the same, since it's always the same endpoint!

For the weakmap I was thinking of caching via `url => graphQL client`.

sohkai (approved):
🙌 awesome! We'll be able to switch the URL out to our own in the future (it could also be something we parameterize), but we can do that later :).

PR #405, opened 2020-09-11
Dispute timeline: Prevent accordion from auto-collapsing

### LINKED ISSUE #404: Disputes: round accordion collapses on re-renders
If a dispute has gone past the first round, we currently show a collapsed accordion for the previous rounds:

Screen Shot 2020-09-11 at 1 42 18 PM

Opening this accordion works, but it appears to re-collapse on ever new graphQL poll or component re-render.

fabriziovigevani:
fixes #404 

We had to convert `getDisputeTimeline` to a hook `useDisputeTimeline`.

### DIFF
--- a/src/components/Disputes/DisputeTimeline.js
+++ b/src/components/Disputes/DisputeTimeline.js
@@ -7,9 +7,6 @@ import DisputeRoundPill from './DisputeRoundPill'
 import Step from '../Step'
 import Stepper from '../Stepper'
 
-import { useCourtClock } from '../../providers/CourtClock'
-import { useCourtConfig } from '../../providers/CourtConfig'
-
 import {
   IconFlag,
   IconFolder,
@@ -26,22 +23,15 @@ import {
   getPhaseStringForStatus,
 } from '../../types/dispute-status-types'
 import { dateFormat } from '../../utils/date-utils'
-import { getDisputeTimeLine } from '../../utils/dispute-utils'
+import useDisputeTimeline from '../../hooks/useDisputeTimeline'
 
-const DisputeTimeline = React.memo(function DisputeTimeline({ dispute }) {
+const Timeline = React.memo(function Timeline({ timeline }) {
   const theme = useTheme()
-  const courtConfig = useCourtConfig()
-  const { currentTermId } = useCourtClock()
-  const disputeTimeLine = getDisputeTimeLine(
-    dispute,
-    courtConfig,
-    currentTermId
-  )
 
   return (
     <div>
       <Stepper lineColor={theme.accent.alpha(0.3)} lineTop={12}>
-        {disputeTimeLine.map((item, index) => {
+        {timeline.map((item, index) => {
           if (!Array.isArray(item)) {
             return <ItemStep key={index} item={item} index={index} />
           }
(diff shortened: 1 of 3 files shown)


src/hooks/useDisputeTimeline.js
```diff
@@ -0,0 +1,98 @@
...
+          phase: DisputePhase.ClaimRewards,
+          active: currentPhase.phase === DisputePhase.ClaimRewards,
+          roundId: currentPhase.roundId,
+        })
+      }
+
+      return timeLine
+    },
+
+    // We are leaving out courtConfig and dispute.rounds as dependencies to prevent  the timeline  from being recomputed on every poll.
```
sohkai, line 93:
```suggestion
    // We are leaving out courtConfig and dispute.rounds as dependencies to prevent the timeline from being recomputed on every poll.
```

src/components/Disputes/DisputeTimeline.js
```diff
@@ -302,4 +292,8 @@ const StyledAccordion = styled.div`
   }
 `
 
-export default DisputeTimeline
+export default function DisputeTimeline({ dispute }) {
+  const disputeTimeLine = useDisputeTimeline(dispute)
```
sohkai, line 296: Not sure if the `memo()` is really that beneficial (since the timeline object would change on each hook re-run anyway?), otherwise we could consider inlining this hook with `Timeline`?

sohkai (approved):
LGTM!

sohkai:
@fabriziovigevani Could we also port this over to the protocol dashboard? 😄

src/components/Disputes/DisputeTimeline.js
fabriziovigevani: Are you referring to the memoized component or the memoized timeline value?

The problem with inlining the hook is that the `dispute` object is a new one on every poll which will cause a re-render on `Timeline` and causing the auto-collapsing issue.

fabriziovigevani:
> @fabriziovigevani Could we also port this over to the protocol dashboard? 😄

Yes! Want to first clear out your comment first and can do a new PR with the update (if required).

src/components/Disputes/DisputeTimeline.js
sohkai: Hmm unless I'm misunderstanding something, I'm not sure how this fixes the issue.

If `disputeTimeline` is a new value on every poll, then `Timeline`'s `timeline` prop will be different on every poll, and even with the `memo()` applied, we'd still be re-rendering `Timeline`?

fabriziovigevani: Ah no, the new value on every poll is `dispute`, `useDisputeTimeline` memoizes the `disputeTimeline` value with [these values](https://github.com/aragon/court-dashboard/blob/f040a43ff1d91f92d86699eddaf907a9085cce3a/src/hooks/useDisputeTimeline.js#L96) as dependencies.

sohkai: Got it, that finally makes sense for me 😄. Thanks for the explanation 🙇‍♂️!

sohkai:
@fabriziovigevani Since this was relevant here, what do you think about merging this here and then re-applying it to the protocol's dashboard? 😄

src/components/Disputes/DisputeTimeline.js
fabriziovigevani: No problem! :)

PR #406, opened 2020-10-05
Remove “Buy ANJ” references

bpierre:


sohkai (approved):
LGTM, hopefully we'll be able to revert these soon with a different link and asset :wink:

